PR #12, opened 2024-07-18
v close match for signin_up, needs correct fonts though

joetaylor-harness:
Very very close now to pixel point reference for sign in/up views. @tnhu fonts are the main things missing, this is probably something you can help with. I'm not actually sure that we are even using Inter as the base font tbh, and we do need Cal Sans for the sign in/up titles.

https://www.loom.com/share/eb8c05d7ca834bdca00e7ac82f1f0f7b


cjlee1:
Nice job on adding  the components from Shadcn and matching the mockups 
@joetaylor-harness

PR #18, opened 2024-07-19
Started componentizing appshell layout components, added decent pp-match NavBar

joetaylor-harness:
https://www.loom.com/share/72cf50ada0454593af79c7abe3f7e665

New navbar component is a fairly decent PP match, which will get much closer to 100% when finished (namely when PP's icon set is shared).


bradrydzewski:
awesome, as a next step we should (as a team) decide how we want to break into components.  for example:

```jsx
<Navbar.Root>
  ...
  <Navbar.Group>
    <NavBar.Item asChild>
      <Link to="...">
        ${ i18n.repositories }
      </Link>
    </NavBar.Item>
    <NavBar.Item>Pipelines</NavBar.Item>
    <NavBar.Item>Executions</NavBar.Item>
  </Navbar.Group>
  <Navbar.Group>
    ...
  </Navbar.Group>
</Navbar.Root>
```

PR #36, opened 2024-07-25
Screen create pipeline

joetaylor-harness:
Almost there, now a v close match with Pixel Point. Needs PP gradients applying to ai button and the template cards.

image


joetaylor-harness:
First attempt at better matching PP template card background effect 
image

PR #87, opened 2024-08-14
Mapped out basic structure and layout

joetaylor-harness:


abhinavrastogi-harness (approved):
we should check if we can invert the outlets for pipeline pages

PR #95, opened 2024-08-15
feat: [pipe-20916]: start adding components to playground for pr conv…

cjlee01:
add components for pr conversation  and pr listing page to playground
image

image


joetaylor-harness (commented):
Looking good although it doesn't make sense to have the top level components such as entity-list moved to the playground components folder, this component should still be imported from Canary into Playground

vardanbansal-harness:
> Looking good although it doesn't make sense to have the top level components such as entity-list moved to the playground components folder, this component should still be imported from Canary into Playground

Agreed, this component is generic enough to go to canary and then imported into playground components/pages if needed.

PR #96, opened 2024-08-16
test responsive page in rootlayout and pipline layout

jessieyu777:
test layout with tailwind tag css


packages/playground/src/layouts/RepoLayout.tsx
```diff
@@ -7,10 +7,10 @@ const RepoLayout: React.FC = () => {
 
   return (
     <div>
-      <header style={{ background: '#333', padding: '20px', height: '56px' }}>
+      <header className="p-5 h-16 bg-black">
```
abhinavrastogi-harness, line 10: let's not use absolute colors like `black`. use semantic colors instead, like `bg-background`

packages/playground/src/layouts/RepoLayout.tsx
```diff
@@ -7,10 +7,10 @@ const RepoLayout: React.FC = () => {
...
         <h1>Repository {repoId}</h1>
       </header>
-      <header style={{ background: '#222', padding: '20px' }}>
+      <header className="p-5 h-16" style={{ lineHeight: 1}}>
```
abhinavrastogi-harness, line 13: do we need to define the `lineHeight`? if so, let's not use inline styles.. use TW classes instead

packages/playground/src/layouts/RootLayout.tsx
```diff
@@ -9,9 +9,9 @@ const RootLayout: React.FC = () => {
   const showNavbar = !hideNavbarPaths.includes(location.pathname)
 
   return (
-    <div className="bg-background flex">
+    <div className="bg-background" style={{ display: 'flex' }}>
```
abhinavrastogi-harness, line 12: add tw class `flex` instead of inline style

jessieyu777: solved to the flex tag in tailwind

packages/playground/src/layouts/RepoLayout.tsx
jessieyu777: deleted line height in advance

packages/playground/src/layouts/RepoLayout.tsx
jessieyu777: Solved from bg-black to bg-background

PR #102, opened 2024-08-20
App shell components

joetaylor-harness:


packages/canary/src/icons/chaos-engineering-icon.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
abhinavrastogi-harness, line 1: let's remove width and height from all svgs

packages/canary/src/icons/chaos-engineering-icon.svg
```diff
@@ -0,0 +1,10 @@
...
+</g>
+<defs>
+<clipPath id="clip0_23_73">
+<rect width="12" height="12" fill="white"/>
```
abhinavrastogi-harness, line 7: should we remove fill colors as well?

packages/canary/src/icons/harness-logo.svg
```diff
@@ -0,0 +1,14 @@
...
...ZiVV2n1cACX+y0TLPhsukfLzgn6ZHsbpDnyWb/aqNlnl2U9EP7GZjAOizOLFpTK18yfQklg2XfzQmP40yHzjG9ChWB8pziHM+NZUvmx7FyrGbin5RnXgC+BrwgulRrH38GZhpo+UvNlx+UpN4GtyvAL81PYpFbke8yjRqE380PYq1L7up6EePpMeQcW8HqoGY6XGiSZ6B7IXUjLQrYB+yKy4/OqPydXBnAPOAnabHiaBfg37NRsu/7IrLz1a+WkHLwPNQrgVNmB4nArKILEPcK5g6Yp3pYawDs+EKgob0aah7C8JHTY8SYjsRFtGWvYW6qk2mh7EOzoYrKBpTR+LqtQhnA/1MjxMyKVRuot/2B5g4tsX0MFbXbLiCpH5dP8pi5wDfAo42PU4oKM/huNdQPfJJ06NY3WfDFUSN6z+EylUgU4AK0+ME1C6Uhyh3bs59GGIFiQ1XUK3YeAiZtrMRLgXeb3qcgHkF0dtpcx+irmqX6WGswtlwBd2y5AdwuAQ4Cxhkehyf24mwHHVut0fCB5sNVxgs0XKGpT4FXAyMx24+dqTA86B3k3FX2FVW8NlwhUn9psMob6nG1RkIJ2OPugd4A+RBNPt9akeuNz2MVRw2XGG0YsNwstmzEL6KcgLRDNgGVB4mJkvtpWjCx4YrzOrXjSDm1CByDnASUG56pNKTJOijOM6DtA5/njrJmp7IKj4brihoSB8BejroF4FTgENMj1RkCrwGugKVn5GNv2CDFW42XFGyYuMhuG2n4FKH8J/AKIL9GtgO/AlYhsQep3r4P0wPZHkjyC9aq6fqNUZZ89HgnA5MAv4VGGJ4qu5qBV5D5EnUfRRp/QPVY7aYHsrylg1X1K1KD2A7xyPup0DGgx4HDMNflzzaDvwD5X9xeJK27LPUVTWbHsoyx4bLarcqPYBtbUcTcz6MK6cgnAQcSe7AVu9CpuxASIL8FdXnQJ5BZA3Vwzciop7NYfmWDZfVOVXh4fWDKY+NAT6IK8cj+n5gNLkV2WHkDnTt6SeVWXKbfTuAd0DWg76K6BpcZw0x9zVat2ygblxrMf44VrjYcFndt/LVCtoOHUS2rRI3NgJHRwEjQeLgHgEyGOQQ0D4I/VEccnHaRe5KrlsQNqKkUUkirEVlHbgpBjpvM6Fyh9E/n2VZlmWVyv8HWLAsP5MtI6MAAAAASUVORK5CYII="/>
```
abhinavrastogi-harness, line 12: this is not a real svg.. it seems to be base64 encoded into an svg. 
also, we should generally keep the harness logo independent of the background circle. 
we can pick the image from https://github.com/harness/uicore/blob/main/packages/icons/src/harness.svg

packages/canary/src/icons/chaos-engineering-icon.svg
joetaylor-harness: Don't forget that we today found out that we cannot use nucleo icons in most cases because Pixel Point alters the base icons. Until a better approach can be found we need to export the relevant icons from Figma, which is why you see width and height set. This should not be a merge blocker right now IMO.

packages/canary/src/icons/chaos-engineering-icon.svg
joetaylor-harness: See comment above. This should not be a merge blocker.

packages/canary/src/icons/harness-logo.svg
joetaylor-harness: This should not be a merge blocker.

joetaylor-harness:
@abhinavrastogi-harness this should have been merged so the MV team could update their local versions and had all Tuesday to work with it. Nothing here was a merge blocker, so we've pushed out another day. Let's find a sensible trade off between thoroughness and velocity please.

abhinavrastogi-harness:
> @abhinavrastogi-harness this should have been merged so the MV team could update their local versions and had all Tuesday to work with it. Nothing here was a merge blocker, so we've pushed out another day. Let's find a sensible trade off between thoroughness and velocity please.

Yep, none of these are blockers, otherwise they would have been marked as such. 

Team has been busy with doing other pages and making progress. I don't think anything has been "pushed out".

I'll merge this now.

PR #103, opened 2024-08-20
Proposal: Figma tokens (colors and dimensions) including different themes are added to tailwind directly

muhammad-noorariffin-harness:
# Summary
- This PR includes the CSS files (for Harness Open Source, Harness Theme, and UI Core theme) and tailwind files (color, dimensions) which is exported from Figma

# TL;DR
- Implementation is only exporting semantic and component tokens from Figma into CSS and tailwind files.
- These same semantic and component tokens are used by the designers in the Figma UI
- Adding these tokens directly to Tailwind ensures that the developers can easily reference the tokens from Figma and use them in tailwind. What they see in Figma is what they can use in code.

# Background
## Starting from Figma...
- Our Figma tokens are organized as such: A core collection and its theme collections.
- The theme collections contains base tokens for that specific theme. At this point since there are 3 themes, there are 3 collections. The tokens in this theme have two modes: Light, Dark. The variables within this collection is not used by the designers directly in Figma. Instead, these are only used as reference by the core collection.
- The core collection contains both semantic and component tokens for the 3 themes. The tokens here reference the base tokens from the respective theme collections meaning that the tokens in the core collection has access to to Light/Dark values in the theme collections. The designers will be using these tokens in their Figma UIs.
- In summary, think of the core collection as an interface that allows the designers access to semantic and component tokens that can be used irregardless of theme. 

> Core collection in Figma with colors for different themes
Screenshot 2024-08-21 at 1 12 43 PM

> Theme collection in Figma with light/dark values 
Screenshot 2024-08-21 at 1 13 38 PM

## Through the token pipeline...
- Given the explanation above, we only need to export the Figma tokens from the core collection (remember the interface?). Meaning the exported tokens that will be built to CSS and Tailwind inlcudes semantic and component tokens but not base tokens.
- With this method, we can ensure that for the different themes in the generated CSS files, they all have the same set of CSS variables. Same goes for Tailwind. This provides confidence for the developers that they can use the semantic and component level tokens within their code irregardless of the theme

## To Tailwind file in this codebase
- Previously we discussed that we will be following shadcn method of theming. Meaning we have a CSS file that contains the mapping between the semantic and component level tokens to the CSS variables. 
- However, given that we are only exporting semantic and component level tokens, this CSS file that includes the shadcn mapping is not generated. 
- With that same reason, we can actually add the semantic and component level tokens (color and dimensions) to tailwind directly, which is then picked up by the Tailwind intellisense. With this method, the developer has that confidence that what they see in Figma is what they can use in Tailwind. Ensuring design consistency
- On a side note, we added a prefix to the tailwind tokens from Figma, so that the likelihood of the tokens from Figma clashing with tokens used directly in Tailwind is mitigated

# Tailwind intellisense
> What devs see in Figma is what they can use in Tailwind
Screenshot 2024-08-21 at 1 05 31 AM

### DIFF
--- a/apps/storybook/.storybook/preview.js
+++ b/apps/storybook/.storybook/preview.js
@@ -25,10 +25,16 @@ const preview = {
   decorators: [
     withThemeByClassName({
       themes: {
-        light: '',
+        "Harness Open Source: Light": 'harness-open-source-light',
+        "Harness Open Source: Dark": 'harness-open-source-dark',
+        "Harness Theme: Light": 'harness-theme-light',
+        "Harness Theme: Dark": 'harness-theme-dark',
+        "UI core: Light": 'ui-core-light',
+        "UI core: Dark": 'ui-core-dark',
+        light: 'light',
         dark: 'dark'
       },
-      defaultTheme: 'dark'
+      defaultTheme: "Harness Open Source: Light"
     })
   ]
 }
--- a/packages/canary/postcss.config.js
+++ b/packages/canary/postcss.config.js
@@ -1,6 +1,6 @@
 export default {
   plugins: {
     tailwindcss: {},
-    autoprefixer: {},
-  },
+    autoprefixer: {}
+  }
 }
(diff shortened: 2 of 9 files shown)


bot2-harness:

PR #108, opened 2024-08-21
Get standard list views formatted

joetaylor-harness:


joetaylor-harness:
Better matching to PP bg colors for page and navbar

image

abhinavrastogi-harness (approved):

PR #112, opened 2024-08-21
add responsive in rootlayout

jessieyu777:


jessieyu777 (commented):
sidebar will hide under 768px and content will responsive in the page

packages/playground/src/layouts/RootLayout.tsx
```diff
@@ -85,9 +85,9 @@ const RootLayout: React.FC = () => {
   ]
 
   return (
-    <div className="bg-background flex">
+    <div className="bg-background grid md:grid-cols-[220px_minmax(900px,_1fr)]">
```
abhinavrastogi-harness, line 88: maybe we should also add the class `min-w-screen` to ensure a min-width of 100vh, incase the content doesn't have enough width

packages/playground/src/layouts/PipelineLayout.tsx
```diff
@@ -7,8 +7,8 @@ const PipelineLayout: React.FC = () => {
   const { pipelineId } = useParams<{ pipelineId: string }>()
 
   return (
-    <div>
-      <div style={{ display: 'flex', gap: '10px' }}>
+    <div className="h-[calc(100vh-100px)]">
+      <div className="flex p-2 gap-2.5 h-[44px] items-center">
```
abhinavrastogi-harness, line 11: why do we need a fixed height and `items-center` here? adding the correct vertical padding will fix both height and alignment.

packages/playground/src/pages/pipeline-edit-page.tsx
```diff
@@ -275,7 +275,7 @@ export default function PipelineEditPage() {
   )
 
   return (
-    <div style={{ height: '650px' }}>
+    <div className="h-[calc(100vh-144px)]">
```
abhinavrastogi-harness, line 278: can we remove this wrapper div and add the correct height on `Container.Root` directly by passing `className` to it?

joetaylor-harness:
I'm running this locally and see the yaml/visual editor layout is improved, nice work. I don't see the navbar hide responsively.

packages/playground/src/layouts/PipelineLayout.tsx
jessieyu777: fix all of them to work properly height with padding is 44px so I removed the 44px

packages/playground/src/pages/pipeline-edit-page.tsx
jessieyu777: Solved in the new commit

jessieyu777:
> I'm running this locally and see the yaml/visual editor layout is improved, nice work. I don't see the navbar hide responsively.

Solve in the current commit, please check it out

jessieyu777 (commented):
fix the layout and discuss them after meeting today

abhinavrastogi-harness (approved):

PR #117, opened 2024-08-22
Add "compact" variant to toggle group component

srdjan-harness:


pnpm-lock.yaml
abhinavrastogi-harness, line 1: why do we have this major change in the lockfile?

abhinavrastogi-harness (approved):

PR #118, opened 2024-08-22
Handle more lists, custom error page, better home page

joetaylor-harness:


joetaylor-harness:
image

joetaylor-harness:
Using same list component but in a repo-pipelines page: 
image

joetaylor-harness:
Added execution list standalone and repo-contained
image

joetaylor-harness:
Remove the 2nd level of tabs, add edit button 

image

packages/playground/src/components/repo-execution-list.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+
+const mockExecutions = [
```
abhinavrastogi-harness, line 4: ideally we should not haev any mock data in the `components` folder. the data should be passed in as props from the `pages` wrappers.

abhinavrastogi-harness (approved):

packages/playground/src/components/repo-execution-list.tsx
joetaylor-harness: @abhinavrastogi-harness fully agree on no mock data not being in the raw component, but I am slightly confused by this component because it is not referenced in a page anywhere. I think you might have created it yesterday, and I copied it + renamed but never used

https://github.com/harness/canary/commit/88ee3c03bbd384dbe2f20f3b1890df7850e42984#diff-ac27a4a88d11f8b5b53d6129a3b7062d8eea43fee3ab679c940a26def9d742d6

PR #127, opened 2024-08-23
Demo tweaks

joetaylor-harness:


joetaylor-harness:
Added the animated homepage
image

joetaylor-harness:
Added FullscreenLayout to handle WIP component for /signin and /signup
image

joetaylor-harness:
Tweaked studio view with temporary css override to move the controls to the right, flip horizontally and invert. Better matched FooterBar to PixelPoint.
image

joetaylor-harness:
Better matched Repos > Pull Requests list to PP. Colors are not a match yet, but we have a basic method to create colorized badges now.
image

packages/canary/src/styles.css
```diff
@@ -195,3 +195,11 @@
 .entity-list-action-chevron {
   @apply transition-transform transform rotate-0 text-tertiary-background group-hover:text-primary;
 }
+
+/* Temporary studio controls override */
+.react-flow__panel {
```
abhinavrastogi-harness, line 200: these should not be in canary package, as canary should not be aware of pipeline studio.
we can add them in playground styles.css

abhinavrastogi-harness (approved):

packages/canary/src/styles.css
joetaylorland: This was a short term fix for the demo but you are absolutely correct.

PR #134, opened 2024-08-27
Add selection feature to yaml-editor

srdjan-harness:
Highlighting code using paths. Optionally we can pass revealInCenter.

### DIFF
--- a/packages/yaml-editor-playground/src/App.css
+++ b/packages/yaml-editor-playground/src/App.css
@@ -34,3 +34,7 @@ body,
   --indent-color: transparent !important;
   --indent-color-active: transparent !important;
 }
+
+.highlightYaml {
+  background-color: #244c27;
+}
--- a/packages/yaml-editor/src/hooks/useDecoration.tsx
+++ b/packages/yaml-editor/src/hooks/useDecoration.tsx
@@ -1,0 +1,62 @@
+import React, { RefObject, useEffect, useMemo, useState } from 'react'
+import * as monaco from 'monaco-editor'
+import { getOutlineModel, processOutlineModel } from '../utils/outline-model-to-path'
+
+type UseDecoration = (arg: {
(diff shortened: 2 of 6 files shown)


packages/yaml-editor/src/hooks/useDecoration.tsx
```diff
@@ -0,0 +1,62 @@
...
+          if (revealInCenter) {
+            editorRef?.current?.revealLineInCenter(range.startLineNumber)
+          }
+        }
+      },
+    []
+  )
+
+  useEffect(() => {
+    const timeout = setTimeout(() => computeRange(editorRef, selection?.path, selection?.revealInCenter), 100)
```
abhinavrastogi-harness, line 42: what's the purpose of the 100ms hardcoded timeout?
can we make it configurable via props if a delay is required?
or maybe replace this with a `requestIdleCallback()` if the intent is to just queue the action?

abhinavrastogi-harness (approved):

packages/yaml-editor-playground/src/App.css
```diff
@@ -34,3 +34,7 @@ body,
   --indent-color: transparent !important;
   --indent-color-active: transparent !important;
 }
+
+.highlightYaml {
```
vardanbansal-harness, line 38: Do we have more css other than this file? We should move to tailwind here so that we can use other tailwind utility classes.

packages/yaml-editor/src/hooks/useDecoration.tsx
srdjan-harness: Will be fixed in the following PR.

packages/yaml-editor-playground/src/App.css
srdjan-harness: This is inside yaml-editor-playground package. yaml-editor does not have any css. In the playground we can use tailwind class.

PR #136, opened 2024-08-27
Fix build issues for `playground` package

vardanbansal-harness:
Fix type issues encountered while building `playground` package.

### DIFF
--- a/packages/playground/src/components/no-list-data.tsx
+++ b/packages/playground/src/components/no-list-data.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Icon, Text, Button, cn } from '@harnessio/canary'
+import { Icon, Text, Button, cn, IconProps } from '@harnessio/canary'
 import PlaygroundListSettings from '../components/playground/list-settings'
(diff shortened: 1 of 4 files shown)


cjlee01 (approved):

packages/playground/src/components/no-list-data.tsx
```diff
@@ -37,7 +37,7 @@ const NoListData: React.FC<NoListDataProps> = ({
       className={cn('w-full h-full flex flex-col place-content-center place-items-center', {
         'py-20 pb-24': insideTabView
       })}>
-      {iconName && <Icon name={iconName} size={iconSize} />}
+      {iconName && <Icon name={iconName as IconProps['name']} size={iconSize} />}
```
abhinavrastogi-harness, line 40: instead of casting, set the type on line 8 as `Partial<IconProps['name'], '...'>`

vardanbansal-harness: Fixed in https://github.com/harness/canary/pull/137

PR #146, opened 2024-08-30
Input accordion advanced

jessieyu777:
- Add recursive component currently and fix complex padding style of children components and border 
- Screenshot 2024-08-30 at 4 32 06 PM
- Change arrow from left to right from accordion extension

### DIFF
--- a/packages/playground/src/components/execution/key-value-table.tsx
+++ b/packages/playground/src/components/execution/key-value-table.tsx
@@ -25,9 +25,9 @@ interface KeyValueTableProps {
   tableSpec: KeyValuePair[]
 }
 
-//manage style for repeated use
-const accordionContentStyle = 'w-full pl-0 pr-0 border-0 pb-0'
-const specTitleStyle = 'text-ring flex-grow text-left -ml-1'
+//manage style for using repeatly
+const accordionContentStyle = 'w-full pl-1 pr-0 border-0 pb-0'
+const specTitleStyle = 'text-studio-2 flex-grow text-left -ml-1'
 
 export const KeyValueTable: React.FC<KeyValueTableProps> = ({
   className,
@@ -43,13 +43,13 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({
       if (typeof item.value === 'string') {
         return (
           <ul className="border-b" key={index}>
-            <li className="py-2.5 inline-block w-1/2 " style={{ paddingLeft: `${level * 1 + 2}rem` }}>
-              <Text size={2} weight="normal" className="text-ring">
+            <li className="py-2.5 inline-block w-1/2" style={{ paddingLeft: `${level * 1 + 1.5}rem` }}>
+              <Text size={2} weight="normal" className="text-studio-7">
                 {item.name}
               </Text>
             </li>
             <li className="pr-2.5 py-2.5 inline-block w-1/2">
-              <Text size={2} weight="normal" className="text-ring">
+              <Text size={2} weight="normal" className="text-studio-7">
                 {item.value}
               </Text>
             </li>
@@ -62,7 +62,7 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({
               <AccordionTrigger
                 className="w-full pt-2 pb-2 pr-4 flex"
                 leftChevron
-                style={{ paddingLeft: `${level * 1 + 2}rem` }}>
+                style={{ paddingLeft: `${level * 1 + 1}rem` }}>
                 <Text size={2} weight="normal" className={specTitleStyle}>
                   {item.name}
                 </Text>
@@ -84,12 +84,12 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({
         return (
           <TableRow key={index} className="border-b">
             <TableCell className="pt-2.5 pl-4 w-1/2">
-              <Text size={2} weight="normal" className="text-ring">
+              <Text size={2} weight="normal" className="text-studio-7">
                 {item.name}
               </Text>
             </TableCell>
             <TableCell className="pt-2.5 w-1/2">
-              <Text size={2} weight="normal" className="text-ring">
+              <Text size={2} weight="normal" className="text-studio-7">
                 {item.value}
               </Text>
             </TableCell>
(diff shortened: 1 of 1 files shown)


jessieyu777 (commented):
- test different nested accordion in table components

- TableCell Can include Accordion but the Accordion cannot add more table inside, so I used ul, li instead. 
- padding style with level in the children components 
- border-b only happen in simple column<tablerow> and <ul> for listItem rendering

jessieyu777 (commented):
delete double border-0

joetaylor-harness:
Looking at structure, the /pages/execution-details-page does not pass any mock data to the component <ExecutionDetails />, it seems like the mock data is being handled at the component level. This seems similar to what the components/pull-request stuff was doing before I refactored to have data passed at the pages level. 

The ExecutionDetails component itself looks like it is actually a page, since it using Layout props etc. Unless I am missing something, it feels like components/execution folder is rather muddy structure compared to the other sections of playground?

packages/playground/src/components/execution/key-value-table.tsx
```diff
@@ -25,13 +25,98 @@ interface KeyValueTableProps {
   tableSpec: KeyValuePair[]
 }
 
+//manage style for repeated use
+const accordionContentStyle = 'w-full pl-0 pr-0 border-0 pb-0'
```
joetaylor-harness, line 29: There is a lot of inline styling for the Accordion component – would it be better to create the necessary variant(s) in the canary Accordion component to handle any necessary styling deltas for this particular view?

packages/playground/src/components/execution/key-value-table.tsx
```diff
@@ -25,13 +25,98 @@ interface KeyValueTableProps {
...
 }) => {
-  // ToDO:const getAccordian = () => <div></div> // do this function recursively
+  const renderListItems = (items: KeyValuePair[], level: number = 1) => {
+    //detect if the listItems is objects or array, tailwind css will not generate
+    const listItems = Array.isArray(items) ? items : [items]
+
+    return listItems.map((item, index: number) => {
+      if (typeof item.value === 'string') {
+        return (
+          <ul className="border-b" key={index}>
```
joetaylor-harness, line 45: ul and li usage here feels like an antipattern. why doesn't AccordionContent handle <Table>? Could the Accordion component be altered to allow for that?

packages/playground/src/components/execution/key-value-table.tsx
```diff
@@ -25,13 +25,98 @@ interface KeyValueTableProps {
...
+        )
+      } else if (Array.isArray(item.value) || typeof item.value === 'object') {
+        return (
+          <Accordion type="single" key={index} className="border-0" defaultValue={item.name} collapsible>
+            <AccordionItem value={item.name} className="border-0">
+              <AccordionTrigger
+                className="w-full pt-2 pb-2 pr-4 flex"
+                leftChevron
+                style={{ paddingLeft: `${level * 1 + 2}rem` }}>
+                <Text size={2} weight="normal" className={specTitleStyle}>
```
joetaylor-harness, line 66: Overriding the Text component with className here, but alignment is already a Text prop. Do we really need to diverge with all the styling on this element so much?

packages/playground/src/components/execution/key-value-table.tsx
```diff
@@ -25,13 +25,98 @@ interface KeyValueTableProps {
...
+      }
+      return null
+    })
+  }
+
+  const renderTableRows = (tableSpec: KeyValuePair[]) => {
+    return tableSpec.map((item, index: number) => {
+      if (typeof item.value === 'string') {
+        return (
+          <TableRow key={index} className="border-b">
```
joetaylor-harness, line 85: Would adding a variant to the Table component to handle the styling delta be better than lots of inline alterations?

joetaylor-harness:
I'm left with the thought did we need to get so heavily into implementation with this relatively minor thing, which has perhaps taken several days of work? 

Or should we have, from a design engineering angle, mocked the ui to show correctly styled elements quickly, and moved on to the next piece of work? 

Getting the view as close to the Pixel Point figma is the current priority, and this Executions / Log view as a whole feels like there is still some way to get the kind of match we are now seeing in the other parts of the Playground app.

We need to get the most impact from the time spent at this stage.

jessieyu777:
> I'm wondering from a design engineering angle, could we have quickly mocked the table to show correctly styled elements, and moved on to getting the whole page stylistically a match with the figma?
> 
> Getting the view as close to the Pixel Point is the current priority, and this Executions / Log view as a whole feels like there is still some way to get the kind of match we are now seeing in the other parts of the Playground app.
> 
> We need to get the most impact from the time spent at this stage.

Hi @joetaylor-harness Thanks for your all information,. Since I discussed this page with @abhinavrastogi-harness on Friday, we don't know if the data structure will be like this way, so we decided to pause the progress and would like to close the branch on that day. I will double confirmed with the manager to see if the page is prioritized than other pages first before you assigned the task to me. I guess it's better to organize better for the team? Thanks for your effort to review this branch so far :)

packages/playground/src/components/execution/key-value-table.tsx
jessieyu777: Just test for a bit for accordion text in the accordion trigger but it's not intuitive handle for accordion text even I over-written the text-align: left on its className. For the further maintaining, I used<Text> first. I've written on this branch for fixing some style and table cell, please wait for merging this branch if you would like to fix anything related on this page. Thanks!
Screenshot 2024-09-03 at 5 37 33 PM

Screenshot 2024-09-03 at 5 35 17 PM

packages/playground/src/components/execution/key-value-table.tsx
jessieyu777: has fixed on flex attributes instead

jessieyu777:
> Looking at structure, the /pages/execution-details-page does not pass any mock data to the component , it seems like the mock data is being handled at the component level. This seems similar to what the components/pull-request stuff was doing before I refactored to have data passed at the pages level.
> 
> The ExecutionDetails component itself looks like it is actually a page, since it using Layout props etc. Unless I am missing something, it feels like components/execution folder is rather muddy structure compared to the other sections of playground?
> 
> My expectation is that pages folder would contain an execution-X-page.tsx file for each of the tabbed views in Pipeline / Executions, and each of those page view would compose the necessary structure using components from components / executions folder, with data passed from page to component.

I've noted on that for further changing since it is related to team member's work but I have  to close this branch first for pixel point. FYI, @vardanbansal-harness. Maybe we can discuss which stage we can follow up this data props in execution-details-page!

packages/playground/src/components/execution/key-value-table.tsx
jessieyu777, line 1: add the text color correctly in the color system based on unified-pipeline

jessieyu777 (commented):
test unnecessary style on accordionContent

abhinavrastogi-harness (approved):

PR #153, opened 2024-09-04
feat: [pipe-20916]: add codeOwnersTable to changes section

cjlee01:
add codeOwnersTable to changes section
 will fix styling in next pr 
image

### DIFF
--- a/packages/playground/src/data/mockCodeOwner.ts
+++ b/packages/playground/src/data/mockCodeOwner.ts
@@ -4,7 +4,7 @@ export const mockCodeOwnerData = {
   evaluation_entries: [
     {
       line_number: 2,
-      pattern: '**',
+      pattern: 'WORKSPACE',
       owner_evaluations: [
         {
           owner: {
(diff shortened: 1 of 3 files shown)


packages/playground/src/data/mockOverviewRealData.ts
```diff
@@ -17,8 +23,8 @@ export const mockOverviewRealData = [
         pullreq_id: 59,
         order: 2,
         sub_order: 0,
-        type: 'branch-delete',
-        kind: 'system',
+        type: 'branch-delete' as EnumPullReqActivityType,
```
vardanbansal-harness, line 26: Can we add type info to `mockOverviewRealData` instead so that this type cast is not required?

packages/playground/src/data/mockCodeOwner.ts
```diff
@@ -16,7 +16,33 @@ export const mockCodeOwnerData = {
             created: 1707765218469,
             updated: 1707765218469
           },
-          review_decision: '' as EnumPullReqReviewDecision,
+          review_decision: 'approved' as EnumPullReqReviewDecision,
```
vardanbansal-harness, line 19: Same as below

vardanbansal-harness (approved):

PR #154, opened 2024-09-04
Execution output log page

jessieyu777:
- added output data for execution output page
- removed unnecessary wrapper <div>
- fixed the parent padding that caused the padding of children moving right
- aligned the same children in the same level of the padding(called padding h*ll)
- added keyValueTable in the execution output page
Screenshot 2024-09-04 at 2 14 54 PM
Screenshot 2024-09-04 at 3 46 39 PM
Screenshot 2024-09-04 at 4 07 45 PM

### DIFF
--- a/packages/playground/src/components/execution/key-value-table.tsx
+++ b/packages/playground/src/components/execution/key-value-table.tsx
@@ -42,13 +42,13 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({
     return listItems.map((item, index: number) => {
       if (typeof item.value === 'string') {
         return (
-          <ul className="border-b flex flex-row align-middle" key={index} style={{ paddingLeft: `${level / 2}rem` }}>
-            <li className="pr-2.5 py-2.5 w-1/2 text-studio-7" style={{ paddingLeft: `${level * 1 + 1}rem` }}>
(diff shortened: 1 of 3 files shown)


jessieyu777 (commented):
- Test successfully - optimize some style in tailwind and removed the calculation of padding

packages/playground/src/components/execution/key-value-table.tsx
```diff
@@ -42,13 +42,13 @@ export const KeyValueTable: React.FC<KeyValueTableProps> = ({
     return listItems.map((item, index: number) => {
       if (typeof item.value === 'string') {
         return (
-          <ul className="border-b flex flex-row align-middle" key={index} style={{ paddingLeft: `${level / 2}rem` }}>
-            <li className="pr-2.5 py-2.5 w-1/2 text-studio-7" style={{ paddingLeft: `${level * 1 + 1}rem` }}>
+          <ul className="border-b flex flex-row align-middle" key={index}>
```
abhinavrastogi-harness, line 45: `index` is not a good key for lists. see https://dev.to/shiv1998/why-not-to-use-index-as-key-in-react-lists-practical-example-3e66

abhinavrastogi-harness (approved):

PR #156, opened 2024-09-06
feat: add gitness signin page

abhinavrastogi-harness:
- uses existing signin page from playground
- adds signin callback and loading props
- uses code-service-client to call login mutation
- token gets set in cookie on success
- redirects to homepage on success


srdjan-harness (approved):

abhinavrastogi-harness:
known checks are failing. will go ahead with merge

PR #158, opened 2024-09-09
Handle empty data

joetaylor-harness:
Add graceful handling to <StackedList description strings when data items are not present.


joetaylor-harness:
Pipelines list now better-handles empty/missing fields in the mockPipelines data. Optional fields will no longer render if empty/undefined. The meter component is refactored to no longer require a full set of 10 bars, instead only needing the bars corresponding to non-empty states. The component then pads out at the left any empty bars:

image

joetaylor-harness:
Executions list (standalone and repository) now handles empty/missing data fields. E.g. 1st item has now success field, 3rd items has no description:

image

joetaylor-harness:
Pull Requests list now handles missing or empty fields in data. E.g. 1st item has now merged state set, no author, no version, no review required set:

image

joetaylor-harness:
Added simple node icon option to NodeGroup component:
image

packages/canary/src/components/branch-chooser.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+function Left({ children }: { children: React.ReactNode }) {
+  return <div className="flex flex-grow gap-6 items-center">{children}</div>
+}
+
+function Right({ children }: { children: React.ReactNode }) {
+  return <div className="flex gap-6 items-center">{children}</div>
+}
+
+function Dropdown({ title, items }: DropdownProps) {
```
abhinavrastogi-harness, line 25: dropdown should take onchange/onselect callback props

packages/canary/src/components/meter.tsx
```diff
@@ -16,30 +16,22 @@ interface MeterRootProps {
...
   return (
     <div className={cn('flex h-[19px] gap-[4px] items-stretch', className)}>
-      {data?.map((col, col_idx) => {
-        return (
-          <div
-            key={col_idx}
-            className={cn('flex w-[5px] h-full rounded-[1px]', stateToBgColor[col.state], stateToGlow[col.state])}
-          />
-        )
+      {bars.map((col, col_idx) => {
```
abhinavrastogi-harness, line 32: index is not a good key: https://dev.to/shiv1998/why-not-to-use-index-as-key-in-react-lists-practical-example-3e66

packages/canary/src/components/treeview.tsx
```diff
@@ -204,14 +205,15 @@ const TreeIndicator = forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivEle
 TreeIndicator.displayName = 'TreeIndicator'
 
 /* eslint-disable @typescript-eslint/no-empty-object-type */
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
abhinavrastogi-harness, line 208: can we just remove the unused vars instead of disabling their lint rules?

abhinavrastogi-harness (approved):

PR #164, opened 2024-09-10
Fixing AuthN issue

vardanbansal-harness:
- Sets token in local storage to make api calls upon successful login.
- Logout user if api results in 401.
- Fix vite config


apps/gitness/src/App.tsx
```diff
@@ -23,6 +23,14 @@ export default function App() {
           return newRequest
         }
         return request
+      },
+      responseInterceptor: (response: Response) => {
+        switch (response.status) {
```
abhinavrastogi-harness, line 28: we might need to do `response.clone()` to allow consuming it later in the app (incase we DONT redirect)

abhinavrastogi-harness (approved):

PR #184, opened 2024-09-14
feat: [pipe-21741]: add simple commenting and reply system

cjlee01:
add simple commenting and reply system
**NOTE: will have to implement a similiar commenting and reply system as gitness this is just to progress forward

https://github.com/user-attachments/assets/ba079eda-37c0-4e8f-8bc4-68299f1f31d8

### DIFF
--- a/packages/playground/src/components/image-carousel.tsx
+++ b/packages/playground/src/components/image-carousel.tsx
@@ -20,7 +20,7 @@ interface ImageCarouselProps {
 // TODO: rewrite this to actually work
 const ImageCarousel = (props: ImageCarouselProps) => {
   const { isOpen, setIsOpen, setZoomLevel, zoomLevel, imgEvent } = props
-  const [imgTitle] = useState(imgEvent[0])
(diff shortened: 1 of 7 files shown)


packages/playground/src/components/image-carousel.tsx
```diff
@@ -20,16 +20,9 @@ interface ImageCarouselProps {
 // TODO: rewrite this to actually work
 const ImageCarousel = (props: ImageCarouselProps) => {
   const { isOpen, setIsOpen, setZoomLevel, zoomLevel, imgEvent } = props
-  const [
-    imgTitle
-    // setImageTitle
-  ] = useState(imgEvent[0])
+  const [imgTitle] = useState(imgEvent[0])
```
vardanbansal-harness, line 23: Does it need a null check here `imgEvent?.[0]`?

vardanbansal-harness (approved):

packages/playground/src/components/image-carousel.tsx
cjlee01: added null check

PR #185, opened 2024-09-16
Playground: add new variant to Button comp to replicate PP ai-button

joetaylor-harness:
- Accommodates future custom button border gradients


joetaylor-harness:
Matched PP gradient outline for the 'ai button'. This is accessible in the ShadCn Button component, using:

`variant="gradient-border"` with prop `gradientType="ai-button"` (this is actually the default gradientType currently, so this prop is technically not required)

image

joetaylor-harness:
I converted the tailwind config file to use ES6 module syntax, because of a 'require' linting error.

vardanbansal-harness (approved):

PR #190, opened 2024-09-17
chore: css bundling fixes

abhinavrastogi-harness:


apps/gitness/src/main.tsx
```diff
@@ -1,4 +1,7 @@
 import ReactDOM from 'react-dom/client'
 import App from './App'
 
+import '@harnessio/canary/styles'
```
joetaylor-harness, line 4: What's the reason to take these imports out of the gitness / styles.css? I think possibly there is an ordering thing going on, that possibly these two imports needed to be in the styles.css AND before the 3 tailwind imports.

abhinavrastogi-harness: let me try that

bot2-harness:

PR #192, opened 2024-09-17
Playground: new view - Repositories > Code

joetaylor-harness:
Building this:

gitness-repository-code


joetaylor-harness:
To accommodate the nested left sub pane menu called for by this view, we need to finally address the layout. Hopefully this will not break things (!). First thing is to fix the header(s) for repo layout:

https://github.com/user-attachments/assets/9508296d-72e2-4c2a-81fe-f34d6013bcec

PR #197, opened 2024-09-18
Playground: fix incorrectly styled first or last node connector icons

joetaylor-harness:
`<NodeGroup.Connector />` does not handle styling of first/last icons correctly.


joetaylor-harness:
When using the `simpleNode` prop with `<NodeGroup />`, the first and last items have too much connector line. 

Screenshot 2024-09-18 at 07 34 37

This fix provides special case props `first` and `last`, which fix the connector line. Apply via `map` and `index`.

abhinavrastogi-harness (approved):

PR #203, opened 2024-09-19
feat: [pipe-21741]: add pr page provider, utils, hooks and types

cjlee01:
add pr page provider, utils, hooks and types

### DIFF
--- a/packages/playground/src/utils/utils.ts
+++ b/packages/playground/src/utils/utils.ts
@@ -70,3 +70,44 @@ export const timeDistance = (date1 = 0, date2 = 0, onlyHighestDenomination = fal
     minutes ? minutes + 'm' : hours || days ? '0m' : ''
   } ${seconds}s`
 }
+
+/**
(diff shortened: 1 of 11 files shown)


packages/playground/src/utils/utils.ts
```diff
@@ -70,3 +70,44 @@ export const timeDistance = (date1 = 0, date2 = 0, onlyHighestDenomination = fal
     minutes ? minutes + 'm' : hours || days ? '0m' : ''
   } ${seconds}s`
 }
+
+/**
```
vardanbansal-harness, line 74: I already added this util https://github.com/harness/canary/blob/main/apps/gitness/src/pages/pipeline-edit/utils/time-utils.ts#L47

vardanbansal-harness: Looks like we need the same util in `package/playground` as well.

vardanbansal-harness (approved):

PR #206, opened 2024-09-19
Playground: new layout sandbox to avoid conflicts/blocks

joetaylor-harness:
To de-risk getting the layout right, factoring in the need to allow body overflow scroll and nested menu columns (such as Repositories > Code), I've temporarily added a layout sandbox. This will stop new layout changes from breaking existing views. Accessed via a navbar accordion group:

image

### DIFF
--- a/packages/playground/src/App.tsx
+++ b/packages/playground/src/App.tsx
@@ -27,6 +27,10 @@ import RepoExecutionListPage from './pages/repo-execution-list-page'
 import RepoWebhooksListPage from './pages/repo-webhooks-page'
 import { CreatePipelinePage } from './pages/create-pipeline-page'
 import { SandboxRoot } from './layouts/SandboxRoot'
+import { SandboxRepo } from './layouts/SandboxRepo'
+import { SandboxRepoListPage } from './pages/sandbox-repo-list-page'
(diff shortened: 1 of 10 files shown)


joetaylor-harness:
All layout POC stuff is separated from the rest of app via use of `sandbox` prefix on demo files. Merging is fine, with the only caveat that the navbar will show the extra accordion group for layout (not a bad thing IMO).

This video shows how to demo the new layout for 2 views: `Repos list` and `Repos > Drone > Summary`. The more exciting one is the summary, because it nicely shows the following:

- The top headers are fixed
- The body scrolls, for accessibility reasons
- The content bottom padding is now working
- A left nested sub panel can be added
- The content can be set to float (maxing at 1200px) or expand to the full width, with or without the nested panel

https://github.com/user-attachments/assets/afd18cd8-a783-48ef-a35b-20a8538247d9

joetaylor-harness:
The main reason I've separated this new layout from the existing approaches is because it is a very different way to do things, and will cause conflicts. Currently, layout is handled in a scrappy, ad hoc fashion with numerous quasi layout components. This new approach fully componentizes layout and allows routes/outlets to inject components such as the `TopBarWidget` and breadcrumbs, which are currently mostly handled in actual pages.

vardanbansal-harness (approved):

PR #209, opened 2024-09-19
chore: Fix build issues in "packages/playground"

vardanbansal-harness:


cjlee01 (approved):

packages/playground/src/pages/landing-page.tsx
jessieyu777, line 1: It seems like we can use noop from lodash for remaining potential props with api call in the future?

packages/playground/src/pages/create-project-page.tsx
jessieyu777, line 1: Thanks for letting me know. I will delete the temperate comment in the future

packages/playground/src/pages/create-pipeline-page.tsx
jessieyu777, line 1: Do we have two icon systems, one is logoName and another one is IconName? I just implemented Icon before but logoName not yet so curious about it.

packages/playground/src/pages/landing-page.tsx
vardanbansal-harness: Yeah you can use it instead of passing `() => {}`

PR #222, opened 2024-09-23
Repo branchlist with api

jessieyu777:
- Add NoData status in branchlist
- Add copied status (not included copy function since not sure what we have to copied for)
- Fetch API in branchlist 
- Add initial Avatar for the use without data
-
Screenshot 2024-09-23 at 2 56 32 PM
Screenshot 2024-09-23 at 2 56 43 PM
Screenshot 2024-09-23 at 2 57 07 PM

PR #237, opened 2024-09-25
Branchlist with apis follow up

jessieyu777:
- Fixed issues and opened up the conversation about tooltips and copied button with UI design(hold for this idea and waiting for the final design), currently decided to remove this idea
- Revised details based on PR #222 
- migrate copied Icon as a component in the playground
- hidden the column without api call or lack of response


packages/canary/src/components/tooltip.tsx
jessieyu777, line 1: export arrow for the use in the canary in the future

vardanbansal-harness (approved):

PR #242, opened 2024-09-26
Playground: new component FileExplorer for Repos > Code

joetaylor-harness:

PR #243, opened 2024-09-26
fix: repo files page

shaurya-harness:
fixes : 
1. Chevron placement to left
2. Chevron only for folders not for files
3. File name highlighted and not the commit message


joetaylor-harness:
This seems like duplicate work with what I did earlier today: https://github.com/harness/canary/pull/242.

PR #246, opened 2024-09-26
Integrating Pagination component on Listing Pages

sans-harness:
- Added pagination across listing pages including PRs, Pipelines, Branches, Commits, web hooks etc.
- Introduced a custom hook "usePagination" to extract out logic and can be re-used across listing pages 
- Component conditionally rendered only when data exists. 
- Currently, we hardcode the total number of pages to 10 and display numerical pagination on all pages. 
- Conditionally changing  and rendering the pagination component in cases where it is not feasible to calculate "total pages" will be taken care of in subsequent PRs. Same for ellipsis. 

Screenshot 2024-09-26 at 12 11 35 PM
Screenshot 2024-09-26 at 12 11 49 PM


vardanbansal-harness (approved):
Nice work!

joetaylorland:
The isActive prop is only for pagination links (the page numbers). The previous and next components now have a disabled prop to control their display.

apps/gitness/src/pages/repo/repo-branch-list.tsx
```diff
@@ -15,16 +14,25 @@ import {
...
 import { useListBranchesQuery, TypesBranch } from '@harnessio/code-service-client'
 import { timeAgo } from '../pipeline-edit/utils/time-utils'
 
 const filterOptions = [{ name: 'Filter option 1' }, { name: 'Filter option 2' }, { name: 'Filter option 3' }]
 const sortOptions = [{ name: 'Sort option 1' }, { name: 'Sort option 2' }, { name: 'Sort option 3' }]
 
 export function ReposBranchesListPage() {
+  // lack of data: total branches
+  // hardcoded
+  const totalPages = 10
```
cjlee01, line 28: can we just add this to a constant file and just export it

sans-harness: Thats a good point. It'll reduce the changes once we get the API working. 

 Will change in the next PR later today.

cjlee01: thanks @sans-harness

PR #259, opened 2024-10-02
fix: add netlify redirect rule to support SPAs

abhinavrastogi-harness:
Adds a `_redirects` config file for netlify to work correctly for single page apps. We need to serve `index.html` for all requests.. not just `/`.

ref: https://docs.netlify.com/routing/redirects/rewrites-proxies/#history-pushstate-and-single-page-apps

`vite` will automatically copy this file from `public` to `dist`, which is the root directory for the netlify deployment.


joetaylor-harness (approved):
LGTM

PR #266, opened 2024-10-03
Playground: very temporary fix for Gitness searchbox padding issue

joetaylor-harness:
This is a very short term fix for the Gitness-specific issue where the `<Searchbox />` component cannot handle tailwind padding, and therefore the left icon overlays the text.

This is a symptom of how the products/modules build tailwind, which must be addressed. Once addressed, this fix will no longer be required.


vardanbansal-harness (approved):
Adding till we have a proper fix.

PR #267, opened 2024-10-03
feat: Integrating Server sent events for Log streaming and Execution progress

vardanbansal-harness:
Integrating Server sent events for Log streaming and Execution progress

### DIFF
--- a/apps/gitness/src/framework/hooks/useLogs.ts
+++ b/apps/gitness/src/framework/hooks/useLogs.ts
@@ -1,0 +1,59 @@
+import { useEffect, useRef, useState } from 'react'
+import { LivelogLine } from '@harnessio/code-service-client'
+import { ExecutionState } from '../../types'
+
+interface UseLogsReturnType {
+  logs: LivelogLine[]
+}
+
+enum LogsSSEEvent {
+  MESSAGE = 'message'
+}
+
+interface UseLogsInterface {
+  repoPath: string
+  pipelineId: string
+  stageNum: number
+  stepNum: number
+  stepStatus?: ExecutionState
+  executionNum: string
+}
+
+export const useLogs = ({
+  repoPath,
+  pipelineId,
+  stageNum,
+  stepNum,
+  stepStatus,
+  executionNum
+}: UseLogsInterface): UseLogsReturnType => {
+  const [logs, setLogs] = useState<LivelogLine[]>([])
+  const eventSourceRef = useRef<EventSource | null>(null)
+
+  useEffect(() => {
+    if (stepStatus === ExecutionState.RUNNING) {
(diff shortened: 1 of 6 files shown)


apps/gitness/src/framework/hooks/useLogs.ts
```diff
@@ -0,0 +1,59 @@
...
+interface UseLogsInterface {
+  repoPath: string
+  pipelineId: string
+  stageNum: number
+  stepNum: number
+  stepStatus?: ExecutionState
+  executionNum: string
+}
+
+export const useLogs = ({
```
cjlee01, line 22: should i similarily implement this into the usespacesse ? @vardanbansal-harness

apps/gitness/src/framework/hooks/useLogs.ts
```diff
@@ -0,0 +1,59 @@
...
+
+interface UseLogsReturnType {
+  logs: LivelogLine[]
+}
+
+enum LogsSSEEvent {
+  MESSAGE = 'message'
+}
+
+interface UseLogsInterface {
```
cjlee01, line 13: add this to a interface file?

apps/gitness/src/framework/hooks/useLogs.ts
```diff
@@ -0,0 +1,59 @@
...
+        )}/${String(stepNum)}/stream`
+      )
+      eventSourceRef.current.onmessage = (event: MessageEvent<string>) => {
+        try {
+          if (event.type === LogsSSEEvent.MESSAGE) {
+            const newLog = JSON.parse(event.data)
+            setLogs((existingLogs: LivelogLine[]) => [...existingLogs, newLog])
+          }
+        } catch (_e) {
+          //ignore error
```
cjlee01, line 50: should we incorporate error with toast?

vardanbansal-harness: Will do that once we have toast working, I am yet to pick that up. It needs some more work.

apps/gitness/src/framework/hooks/useLogs.ts
vardanbansal-harness: It's upto you and not a bad idea actually. `useSpaceSSE` currently handles more than one type of events so we can make a separate `usePREvents` that only handles events for pull requests.

apps/gitness/src/framework/hooks/useLogs.ts
vardanbansal-harness: I am thinking we should only add reusable/common interfaces to the interface file, like we have `.types` file today in some directories. This interface is too specific to this hook. What do you think?

cjlee01 (approved):

PR #276, opened 2024-10-05
Add common filter hook with sort filter

praneshg239:
### Screen recording

https://github.com/user-attachments/assets/f8c1f097-1e24-49e6-961e-e0e97cfb1472

---

#### Added proper types for return values

image

---

#### When no filter is applied

image

#### When a filter is applied

> [!NOTE]
>  Label will be bold if that filter is applied and on clicking the current selected filter item, the filter will be removed

image

### DIFF
--- a/apps/gitness/src/pages/pull-request-list-page.tsx
+++ b/apps/gitness/src/pages/pull-request-list-page.tsx
@@ -34,12 +34,12 @@ function PullRequestListPage() {
   const repoRef = useGetRepoRef()
   const { currentPage, previousPage, nextPage, handleClick } = usePagination(1, totalPages)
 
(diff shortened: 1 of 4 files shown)


c-joetaylor-harness:
Nice work. We should run over some of the UX a bit later, but this is a great start. For example, the active filter state is very subtle with the bold font only, we may want to represent more obviously.

vardanbansal-harness (approved):

praneshg239:
> Nice work. We should run over some of the UX a bit later, but this is a great start. For example, the active filter state is very subtle with the bold font only, we may want to represent more obviously.

Thank you for your suggestion, @c-joetaylor-harness . I am currently working on implementing a small circle in front of the label when it gets selected, and your input is highly appreciated. This update is in progress.

PR #279, opened 2024-10-07
fix the sign-in page redirect repeatedly

jessieyu777:


jessieyu777 (commented):
test sign in successfully

apps/gitness/src/framework/context/AppContext.tsx
```diff
@@ -36,7 +36,10 @@ export const AppProvider: React.FC<{ children: ReactNode }> = ({ children }) =>
         switch (response.status) {
           case 401:
             localStorage.removeItem('token')
-            window.location.href = '/signin'
+            //prevent infinite loop for redirect signin page
+            if (window.location.pathname !== '/signin') {
```
abhinavrastogi-harness, line 40: while this works, this is not the correct solution. we should find the root cause and fix that.

vardanbansal-harness: Added a fix here : https://github.com/harness/canary/pull/286, closing this PR.

PR #285, opened 2024-10-08
feat: [pipe-22104]: add current user to context

cjlee01:
add current user to context


apps/gitness/src/framework/context/AppContext.tsx
```diff
@@ -1,15 +1,21 @@
...
 import {
   CodeServiceAPIClient,
   TypesMembershipSpace,
   membershipSpaces,
-  TypesSpace
+  TypesSpace,
+  TypesUser,
+  getUser
 } from '@harnessio/code-service-client'
+import { useAtom } from 'jotai'
```
vardanbansal-harness, line 10: Lets add a SecOps ticket for this.

vardanbansal-harness:
We haven't implemented cache in this new gitness app yet and I feel we should discuss on a broader level first before we introduce caching as part of this PR. 

For now, if user info is required, let's just add it to the App context **without** the caching part.

apps/gitness/src/framework/context/AppContext.tsx
cjlee01: sure will add a secops ticket

apps/gitness/src/framework/context/currentUser.ts
```diff
@@ -0,0 +1,4 @@
+import { TypesUser } from '@harnessio/code-service-client'
```
vardanbansal-harness, line 1: Lets comment/remove this as well

apps/gitness/src/framework/context/AppContext.tsx
vardanbansal-harness: What does `useAtom` do?

cjlee01: Useatom is just a more flexible state management that can update particular states of a object and was used in diffs and PRs and then tan adopted it to other places @vardanbansal-harness

vardanbansal-harness: So it's like a reducer right? @srdjan-harness has already added one here https://github.com/harness/canary/blob/main/apps/gitness/src/hooks/useThunkReducer.tsx

Can we just try to use it? Two different reducers in a single app is not a good idea.

vardanbansal-harness (approved):

PR #286, opened 2024-10-08
fix: Fix login 🔄 logout loop

vardanbansal-harness:
- Added `useToken` hook to get, set and delete token in local storage.
- Added `/logout` route to log user out and fix login 🔄 logout loop.
- Code cleanup

https://github.com/user-attachments/assets/77ea4424-c542-4c9c-99a7-d44b9622c57c

### DIFF
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -11,7 +11,7 @@ import {
 import { TooltipProvider } from '@harnessio/canary'
 import { queryClient } from './framework/queryClient'
 import PipelineListPage from './pages/pipeline-list'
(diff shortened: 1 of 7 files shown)


abhinavrastogi-harness (approved):
is the backend not setting cookies for the token? ideally UI should not need to manage the token at all.

vardanbansal-harness:
> is the backend not setting cookies for the token? ideally UI should not need to manage the token at all.

It does have a `set-cookie` response header but need to look how it's actually using it.
image

vardanbansal-harness:
Merging this a temporary fix, will check how to leverage `set-cookie` response header instead.

PR #289, opened 2024-10-09
feat: Fix side nav items for gitness, changes for user info badge

vardanbansal-harness:
For `apps/gitness`:

- Adds a new `AppLayout` with only option for `Repositories`.
- Hide `Search` (top-left corner).
- Fixed User info badge on Side nav.
- Clicking on `Repositories` takes to the Repository listing page for the selected project.

**Sample**:

image

https://github.com/user-attachments/assets/b62dca37-a3a9-45d4-a87c-e62ed0f02781

**Playground**:

No changes, stays as is with all the elements visible.

image


joetaylor-harness:
Hey @Vardan, a couple of comments:
- In the Playground screenshot, Repositories is missing
- `isPlayground` feels a bit 'hardcoded' to me, why not instead control which sections of the navbar are shown by explicitly using `showPinned`, `showSandbox`, `showMore` and have these default to true. Then when any other product (e.g. gitness) composes the navbar, it can choose what to show
- The static data for navbar sections actually should be provided by the composing app, so if gitness only wants repositories, then it should only pass repositories in the data

vardanbansal-harness:
> Hey @vardan, a couple of comments:
> 
> * In the Playground screenshot, Repositories is missing
> * `isPlayground` feels a bit 'hardcoded' to me, why not instead control which sections of the navbar are shown by explicitly using `showPinned`, `showSandbox`, `showMore` and have these default to true. Then when any other product (e.g. gitness) composes the navbar, it can choose what to show
> * The static data for navbar sections actually should be provided by the composing app, so if gitness only wants repositories, then it should only pass repositories in the data

Thanks for the feedback @joetaylor-harness. After discussing with @abhinavrastogi-harness as well, I feel this is one of those instances where we might not need to directly import `RootLayout` from Playground, rather it should act as a reference or a template to add a new `AppLayout` in `apps/gitness` itself, with only the options we want and support as of today.

PR #299, opened 2024-10-10
feat: [pipe-21741]: add diff review api

cjlee01:
Add diff review api on changes tab
- can now approve and request changes on diff tab
- also added a primary button for split button
image

### DIFF
--- a/apps/gitness/src/pages/pull-request/types/types.ts
+++ b/apps/gitness/src/pages/pull-request/types/types.ts
@@ -1,4 +1,10 @@
-import { TypesCheck, TypesPullReqActivity } from '@harnessio/code-service-client'
+import {
+  ReviewerListPullReqOkResponse,
+  TypesCheck,
+  TypesPullReq,
+  TypesPullReqActivity,
+  TypesUser
+} from '@harnessio/code-service-client'
 
 export enum orderSortDate {
   ASC = 'asc',
@@ -17,7 +23,8 @@ export enum PullReqReviewDecision {
   approved = 'approved',
   changeReq = 'changereq',
   pending = 'pending',
-  outdated = 'outdated'
+  outdated = 'outdated',
+  approve = 'approve'
 }
 export interface TypeCheckData {
   bypassable: boolean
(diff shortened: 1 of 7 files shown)


packages/playground/src/components/pull-request/interfaces.ts
```diff
@@ -5,7 +5,7 @@ export enum PRCommentFilterType {
   RESOLVED_COMMENTS = 'resolvedComments',
   UNRESOLVED_COMMENTS = 'unresolvedComments'
 }
-export type EnumPullReqReviewDecision = 'approved' | 'changereq' | 'pending' | 'reviewed'
+export type EnumPullReqReviewDecision = 'approved' | 'changereq' | 'pending' | 'reviewed' | 'outdated'
```
vardanbansal-harness, line 8: Why not use existing type here? https://github.com/harness/react-api-client/blob/pkg-update-oct-10/packages/code-service/src/services/schemas/EnumPullReqReviewDecision.ts#L5

If any other value is needed, just extend or add to it.

apps/gitness/src/pages/pull-request/types/types.ts
```diff
@@ -152,3 +159,28 @@ export interface CommentItem<T = unknown> {
...
+  active?: string
+  currentUser: TypesUser
+  pullRequestMetadata?: TypesPullReq | undefined
+  reviewers?: ReviewerListPullReqOkResponse
+  submitReview?: (decision: PullReqReviewDecision) => void
+  refetchReviewers?: () => void
+  loading?: boolean
+}
+
+export type EnumPullReqReviewDecision = 'approved' | 'changereq' | 'pending' | 'reviewed' | 'outdated'
```
vardanbansal-harness, line 186: Same as below

vardanbansal-harness: Why not use existing type here? https://github.com/harness/react-api-client/blob/pkg-update-oct-10/packages/code-service/src/services/schemas/EnumPullReqReviewDecision.ts#L5

If any other value is needed, just extend or add to it (as a local declaration)

vardanbansal-harness (approved):

PR #304, opened 2024-10-11
Delete tokens

sans-harness:
- Implemented delete functionality for tokens and SSH keys with confirmation dialog.
- Replaced temporary token fetching and deletion with `useListTokensQuery` hook  and `useDeleteTokenMutation`.
- Added UI improvements including delete icons and conditional rendering based on data presence.
- Integrated dialog for delete confirmation. 
- Fixed various type issues in both `gitness` and `playground`. Integrated dialogs in `Playground` as well to make sure it remains exhaustive. 
- With this our token and key creation/deletion flow should be complete. 

E2E video: First half of video is `gitness` and second half on tab chnage is `playground`. 

https://github.com/user-attachments/assets/42f56c5b-6c79-41d5-8e59-673ed63d2406


apps/gitness/src/pages/profile-settings/profile-settings-keys-container.tsx
```diff
@@ -12,26 +12,39 @@ import {
...
+  ListTokensErrorResponse
 } from '@harnessio/code-service-client'
 import { TokenCreateDialog } from './token-create/token-create-dialog'
 import { TokenFormType } from './token-create/token-create-form'
 import { SshKeyCreateDialog } from './ssh-key-create/ssh-key-create-dialog'
 import { TokenSuccessDialog } from './token-create/token-success-dialog'
-import { TokensList, KeysList } from '@harnessio/playground'
+import { TokensList, KeysList, DeleteTokenAlertDialog } from '@harnessio/playground'
+
+export interface AlertDeleteParams {
```
cjlee01, line 30: put this in type.ts file?

apps/gitness/src/pages/profile-settings/profile-settings-keys-container.tsx
```diff
@@ -12,26 +12,39 @@ import {
...
   const [tokens, setTokens] = useState<TokensList[]>([])
   const [openCreateTokenDialog, setCreateTokenDialog] = useState(false)
   const [openSuccessTokenDialog, setSuccessTokenDialog] = useState(false)
   const [saveSshKeyDialog, setSshKeyDialog] = useState(false)
+  const [isAlertDeleteDialogOpen, setIsAlertDeleteDialogOpen] = useState(false)
+  const [alertParams, setAlertParams] = useState<AlertDeleteParams | null>(null)
+
   const [apiError, setApiError] = useState<{
-    type: 'keyFetch' | 'tokenFetch' | 'keyCreate' | 'tokenCreate'
+    type: 'keyFetch' | 'tokenFetch' | 'keyCreate' | 'tokenCreate' | 'tokenDelete' | 'keyDelete'
```
cjlee01, line 47: use a enum?

packages/playground/src/components/profile-settings-tokens-list.tsx
```diff
@@ -45,8 +46,12 @@ export const ProfileTokensList: React.FC<PageProps> = ({ tokens }) => {
               <TableCell>{token.expires_at ? new Date(token.expires_at).toLocaleString() : 'No Expiration'}</TableCell>
               <TableCell>{timeAgo(new Date(token.issued_at!).getTime())}</TableCell>
               <TableCell className="content-center">
-                <div className="flex gap-1.5 items-center justify-end">
-                  <Icon name="vertical-ellipsis" size={14} className="text-tertiary-background" />
+                <div
+                  className="flex gap-1.5 items-center justify-end cursor-pointer"
+                  onClick={() => {
+                    openAlertDeleteDialog({ identifier: token.identifier!, type: 'token' })
```
cjlee01, line 52: say if api fails, will we have identifier ??

apps/gitness/src/pages/profile-settings/profile-settings-keys-container.tsx
sans-harness: Good point

packages/playground/src/components/profile-settings-tokens-list.tsx
sans-harness: The identifier isn't being fetched dynamically; it's already available when the user attempts to delete. If the identifier is missing, it indicates that the `useTokenListQuery`, which fetches the tokens along with the `identifier`, has failed, and would consequently throw an error there. This identifier has to be `non-null` over here.

apps/gitness/src/pages/profile-settings/profile-settings-keys-container.tsx
```diff
@@ -51,13 +60,34 @@ export const SettingsProfileKeysPage = () => {
...
+  }
+
   const queryParams: ListPublicKeyQueryQueryParams = {
     page: 1,
     limit: 30,
     sort: 'created',
     order: 'asc'
   }
 
+  useListTokensQuery(
```
abhinavrastogi-harness, line 76: why are we maintaining state ourselves? can we simply use the state reference returned by the hook?

```
const { data, error } = useListTokensQuery()
```

abhinavrastogi-harness: nevermind.. realised we need this for optimistic updates

abhinavrastogi-harness (approved):

PR #306, opened 2024-10-12
Project settings followup issues

jessieyu777:
- fix` save button` is disable after saving data successfully
- using react query instead of the state
- setting props for the updating status & the deleting status
- found an issue from api call if the project not exist
https://github.com/user-attachments/assets/062a1d82-ba8a-4bca-aa1d-a00115fbbbb6


jessieyu777 (commented):
handle edge case 
we need Project not found page from UI 
Screenshot 2024-10-11 at 11 41 36 PM

apps/gitness/src/pages/project-settings/components/FormDialogProjectDelete.tsx
```diff
@@ -27,9 +27,15 @@ interface FormProjDeleteProps {
   handleDeleteProject: () => void
   isDeleteSuccess: boolean
   isDeleting: boolean
+  deleteError: string | null
 }
 //delete project form with dialog
-export const FormDialogProjectDelete = ({ handleDeleteProject, isDeleteSuccess, isDeleting }: FormProjDeleteProps) => {
+export const FormDialogProjectDelete = ({
```
cjlee01, line 33: naming should follow normal conventions?

apps/gitness/src/pages/project-settings/project-settings-sandbox-page.tsx
```diff
@@ -57,28 +62,36 @@ export const ProjectSettingsSandboxPage = ({
...
-
     onFormSubmit(formData)
   }
 
+  useEffect(() => {
+    if (isUpateSuccess) {
+      setSubmitted(true)
+      setIsCancelDisabled(true)
+
+      const timer = setTimeout(() => {
```
cjlee01, line 81: why did we need a timer on this?

apps/gitness/src/pages/project-settings/project-settings-general-page.tsx
```diff
@@ -93,7 +98,9 @@ export const ProjectSettingsGeneralPage = () => {
     deleteSpaceMutation.mutate(
       { space_ref: space?.path },
       {
-        onSettled: () => setIsDeleting(false) // Ensure isDeleting is reset after the mutation completes
+        onSettled: () => {
+          setDeleteError(null)
+        } // Ensure isDeleting is reset after the mutation completes
```
sans-harness, line 103: Shouldn't we reset `onSuccess` ?

apps/gitness/src/pages/project-settings/components/FormDialogProjectDelete.tsx
jessieyu777: fixed

apps/gitness/src/pages/project-settings/project-settings-general-page.tsx
jessieyu777: It is not necessary to reset onSuccess inside onSettled. Both onSuccess and onSettled are lifecycle methods for handling the state of a mutation, but they serve different purposes.

Key Differences Between onSuccess and onSettled:
onSuccess: This is called only when the mutation succeeds. It is where you typically handle any logic that should occur after a successful API call, such as updating local state or redirecting the user.

onSettled: This is called regardless of whether the mutation succeeds or fails. It’s useful for doing any cleanup tasks that should occur after the mutation completes, such as resetting UI elements, loading states, or clearing error messages.

sans-harness: Wouldn't it be better to set the error to null only when the mutation succeeds, rather than resetting it regardless of the outcome, as is done with onSettled?

jessieyu777 (commented):
Reply comments & review it again

apps/gitness/src/pages/project-settings/project-settings-general-page.tsx
jessieyu777: Yeah I can set up this line in the `deleteMutation onSuccess`

jessieyu777: Fixed

jessieyu777:
Screenshot 2024-10-14 at 2 46 04 PM

cjlee01 (approved):

PR #319, opened 2024-10-15
Fix navbar

sans-harness:
- Fixes `navbar` user. It now uses and displays the current user from `App Context`. 
- Adds `types` to `layouts`. 
- Removes`extractInitials` function and uses `getInitials` from `utils`

Gitness: (bottom left)
gitness

Playground: (bottom left)
playground


vardanbansal-harness (approved):
Great work!

PR #321, opened 2024-10-16
feat: Remove usage of "token" from Local Storage

vardanbansal-harness:
With this change, there would be no need of saving `token` in Local Storage. Instead api calls would have `Cookie` set on them for authorization.

- Removes all references of usage of `token` from LS.
- Fixes login, logout and signup pages.

### DIFF
--- a/apps/gitness/src/framework/context/AppContext.tsx
+++ b/apps/gitness/src/framework/context/AppContext.tsx
@@ -14,6 +14,7 @@ interface AppContextType {
   addSpaces: (newSpaces: TypesSpace[]) => void
   isUserAuthorized: boolean
   setIsUserAuthorized: React.Dispatch<React.SetStateAction<boolean>>
+  resetApp: () => void
   currentUser?: TypesUser
 }
 
@@ -55,6 +56,12 @@ export const AppProvider: React.FC<{ children: ReactNode }> = ({ children }) => 
     }
   }, [isAuthorized])
 
+  const resetApp = (): void => {
+    setSpaces([])
+    setCurrentUser({})
+    setIsAuthorized(false)
+  }
+
   const addSpaces = (newSpaces: TypesSpace[]) => {
     setSpaces(prevSpaces => [...prevSpaces, ...newSpaces])
   }
@@ -67,7 +74,8 @@ export const AppProvider: React.FC<{ children: ReactNode }> = ({ children }) => 
         addSpaces,
         currentUser,
         isUserAuthorized: isAuthorized,
-        setIsUserAuthorized: setIsAuthorized
+        setIsUserAuthorized: setIsAuthorized,
+        resetApp
       }}>
       {children}
     </AppContext.Provider>
--- a/apps/gitness/src/framework/context/AppContext.tsx
+++ b/apps/gitness/src/framework/context/AppContext.tsx
@@ -7,12 +7,13 @@ import {
   getUser,
   MembershipSpacesOkResponse
 } from '@harnessio/code-service-client'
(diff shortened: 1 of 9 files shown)


sans-harness:
Last case in the video defaults to `Steven M.` as opposed to `user 5` in the bottom left?

vardanbansal-harness:
> Last case in the video defaults to `Steven M.` as opposed to `user 5` in the bottom left? 

I think those are the default values? This would be when user info is not yet available in `AppContext`

vardanbansal-harness:
> Last case in the video defaults to `Steven M.` as opposed to `user 5` in the bottom left? 

Let me check

apps/gitness/src/framework/context/AppContext.tsx
```diff
@@ -47,26 +41,47 @@ export const AppProvider: React.FC<{ children: ReactNode }> = ({ children }) =>
...
       membershipSpaces({
         queryParams: { page: 1, limit: 10, sort: 'identifier', order: 'asc' }
       }).then((response: MembershipSpacesOkResponse) => {
         if (response.length > 0) {
           const spaceList = response.filter(item => item?.space).map(item => item.space as TypesSpace)
           setSpaces(spaceList)
+        } else {
+          window.location.href = '/create-project'
         }
       })
```
cjlee01, line 54: should we debounce the call like would it repeat calls?

apps/gitness/src/pages/logout.tsx
```diff
@@ -1,16 +1,25 @@
...
+    if (isSuccess) {
+      resetApp()
+      navigate('/signin') // Redirect to sign-in page
+    }
+  }, [isSuccess])
 
   const handleLogout = () => {
-    removeToken() // Remove the token when the user logs out
-    navigate('/signin') // Redirect to sign-in page
+    logout({})
```
cjlee01, line 20: should we add error handling on this ?

apps/gitness/src/framework/context/AppContext.tsx
```diff
@@ -21,21 +23,13 @@ const BASE_URL_PREFIX = `${window.apiUrl || ''}/api/v1`
 const AppContext = createContext<AppContextType | undefined>(undefined)
 
 export const AppProvider: React.FC<{ children: ReactNode }> = ({ children }) => {
```
cjlee01, line 25: how do we handle casses where the token expires?

abhinavrastogi-harness: using an expired token would result in the BE returning a 401 and unsetting the cookie, which we should handle by redirecting user to the signin page.

vardanbansal-harness:
> Last case in the video defaults to `Steven M.` as opposed to `user 5` in the bottom left?

Was a genuine issue where `user` info was not getting set, fixed it now.

https://github.com/user-attachments/assets/e668d26b-b502-42d0-b716-23b23961d12e

apps/gitness/src/framework/context/AppContext.tsx
vardanbansal-harness: Yes, a `401` logs user out.

apps/gitness/src/pages/logout.tsx
vardanbansal-harness: Yes we would, when we eventually has `toaster` integrated.

apps/gitness/src/framework/context/AppContext.tsx
vardanbansal-harness: I removed it

sans-harness (approved):
Nice!

PR #323, opened 2024-10-16
Project membership page

jessieyu777:
Add the members page in the playground with 4 different status: 

- data-loaded
- no-data
- loading
- no-search-matches

https://github.com/user-attachments/assets/923788de-d280-42c0-bdc3-4688db3b4984

Screenshot 2024-10-16 at 12 27 18 PM
Screenshot 2024-10-16 at 12 27 33 PM
Screenshot 2024-10-16 at 12 27 48 PM
Screenshot 2024-10-16 at 12 28 08 PM
Screenshot 2024-10-16 at 12 28 18 PM

### DIFF
--- a/packages/playground/src/components/members-list.tsx
+++ b/packages/playground/src/components/members-list.tsx
@@ -29,7 +29,7 @@ interface PageProps {
 
 export const MembersList = ({ members }: PageProps) => {
   return (
-    <Table variant="asStackedList">
+    <Table variant="asStackedList" className="border-0">
       <TableHeader>
         <TableRow>
           <TableHead className="text-primary">Name</TableHead>
@@ -68,7 +68,7 @@ export const MembersList = ({ members }: PageProps) => {
                       <Badge
                         variant="outline"
                         size="xs"
-                        className="rounded-full font-normal text-xs p-2 h-5 text-tertiary-background text-center m-auto bg-gray-900">
+                        className="rounded-full font-normal text-xs p-2 h-5 text-tertiary-background text-center m-auto bg-tertiary-background/10">
                         {member.role}
                       </Badge>
                     </Text>
--- a/packages/playground/src/components/members-list.tsx
+++ b/packages/playground/src/components/members-list.tsx
@@ -1,0 +1,113 @@
+import React from 'react'
+import {
+  Icon,
+  Table,
+  TableBody,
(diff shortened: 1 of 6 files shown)


jessieyu777 (commented):
Self-review

packages/playground/src/components/members-list.tsx
```diff
@@ -0,0 +1,105 @@
...
+          members.map((member, index) => {
+            return (
+              <TableRow key={index}>
+                {/* NAME */}
+                <TableCell className="content-center my-6">
+                  <div className="flex items-center gap-4">
+                    <Avatar className="w-8 h-8">
+                      {member.avatarUrl && <AvatarImage src={member.avatarUrl} />}
+                      <AvatarFallback className="text-xs p-1 text-center">
+                        {getInitials(member.display_name, 2)}
```
sans-harness, line 55: NIT: maybe define a variable for truncation len?

packages/playground/src/pages/sandbox-settings-project-members-page.tsx
```diff
@@ -1,17 +1,111 @@
...
+      case 'data-loaded':
+        return ', 30 members'
+      case 'no-data':
+        return ', 0 members'
+      case 'no-search-matches':
+        return ', 30 members'
+      default:
+        return ''
+    }
+  }
```
sans-harness, line 56: Not sure what this is doing? Are we using it somewhere?

jessieyu777: it's about showing accurate numbers of members in the title of team. In the different status, we show different numbers. For example, it will show the current numbers of members if data-loaded, if there's no search data, 
Screenshot 2024-10-16 at 1 03 12 PM
I can update the new functions only in the component since no-data not related to this count anymore

packages/playground/src/components/members-list.tsx
jessieyu777: fixed: using size instead

packages/playground/src/pages/sandbox-settings-project-members-page.tsx
jessieyu777: fixed

sans-harness (approved):
Nice!

PR #324, opened 2024-10-16
feat: [pipe-20916]: add bypass

cjlee01:
add bypass on pr panel
image

### DIFF
--- a/packages/playground/src/components/pull-request/utils.tsx
+++ b/packages/playground/src/components/pull-request/utils.tsx
@@ -205,8 +205,7 @@ export const getPrState = (is_draft?: boolean, merged?: number | null, state?: s
 }
 
 export const extractInfoFromRuleViolationArr = (ruleViolationArr: TypesRuleViolations[]) => {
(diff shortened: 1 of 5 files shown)


packages/playground/src/components/pull-request/utils.tsx
```diff
@@ -201,3 +203,20 @@ export const getPrState = (is_draft?: boolean, merged?: number | null, state?: s
     return { icon: 'pr-open', text: 'Open', theme: 'success' }
   }
 }
+
+export const extractInfoFromRuleViolationArr = (ruleViolationArr: TypesRuleViolations[]) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const tempArray: any[] = ruleViolationArr?.flatMap(
```
vardanbansal-harness, line 209: Do we not know the type here? Better to use `unknown` instead.

vardanbansal-harness (approved):

packages/playground/src/components/pull-request/utils.tsx
cjlee01: done

PR #332, opened 2024-10-21
feat: [pipe-22241]: add compare pr page- pt 2  

cjlee01:
refactored code and made it a bit cleaner

### DIFF
--- a/packages/playground/src/components/pull-request/pull-request-compare-button.tsx
+++ b/packages/playground/src/components/pull-request/pull-request-compare-button.tsx
@@ -1,0 +1,93 @@
+import React from 'react'
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuGroup,
+  DropdownMenuItem,
+  DropdownMenuTrigger
+} from '@harnessio/canary'
+import { Icon, Text, Button, ButtonGroup } from '@harnessio/canary'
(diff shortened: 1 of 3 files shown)


packages/playground/src/components/pull-request/pull-request-compare-button.tsx
```diff
@@ -0,0 +1,93 @@
...
+  DropdownMenuTrigger
+} from '@harnessio/canary'
+import { Icon, Text, Button, ButtonGroup } from '@harnessio/canary'
+import { FormFields } from '../../layouts/SandboxPullRequestCompareLayout'
+
+interface PullRequestCompareButtonProps {
+  isSubmitted: boolean
+  isValid: boolean
+  isLoading: boolean
+  formRef: React.RefObject<HTMLFormElement>
```
jessieyu777, line 16: Do we need to use formRef in each form we had in gitness currently? I'm developing a form for it so curious about it

jessieyu777 (commented):
left one quesiton

packages/playground/src/components/pull-request/pull-request-compare-button.tsx
cjlee01: not necessarily, i just have a formRef because i need something in my child component that is used in the parent @jessieyu777

jessieyu777 (approved):
LGTM

PR #336, opened 2024-10-21
feat: [pipe-22240]: pr checks

cjlee01:
For the moment, we won’t have a checks tab and we will navigate traffic from pr page in the prpanel check section to the execution log details page

also other minor fixes 
- fix bypass state in pr panel
- hide chevron in change section when there is nothing to show
- fix css in panel

### DIFF
--- a/apps/gitness/src/pages/pull-request/pull-request-conversation-page.tsx
+++ b/apps/gitness/src/pages/pull-request/pull-request-conversation-page.tsx
@@ -52,6 +52,7 @@ export default function PullRequestConversationPage() {
   } = usePullRequestData()
   const { currentUser: currentUserData } = useAppContext()
   const [checkboxBypass, setCheckboxBypass] = useState(false)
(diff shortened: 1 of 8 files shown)


apps/gitness/src/pages/pull-request/pull-request-conversation-page.tsx
```diff
@@ -52,6 +52,7 @@ export default function PullRequestConversationPage() {
   } = usePullRequestData()
   const { currentUser: currentUserData } = useAppContext()
   const [checkboxBypass, setCheckboxBypass] = useState(false)
+  const { spaceId, repoId } = useParams<PathParams>()
```
sans-harness, line 55: maybe use the custom hooks we have in the framework for this?

cjlee01: didn't use  useGetRepoRef since its both repoid and spaceId

and it seemed shorter to just export both from useParams

PR #339, opened 2024-10-22
Playground: add dropdown menu to navbar footer

joetaylor-harness:
Changes to navbar footer:

- Added dropdown menu (link to destinations are fake)
- Restored avatar stlying to match figmas
- Fixed padding
- Added absolute highlight on hover

image

### DIFF
--- a/packages/canary/src/components/navbar-user.tsx
+++ b/packages/canary/src/components/navbar-user.tsx
@@ -1,25 +1,55 @@
 import React from 'react'
 import { getInitials } from '@/utils/StringUtils'
-import { Avatar, AvatarFallback, AvatarImage } from '..'
+import {
+  Avatar,
+  AvatarFallback,
+  AvatarImage,
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger
+} from '..'
+
+interface items {
+  element: React.ReactNode
+  key: number
+}
 
 const Root: React.FC<{
   username?: string
   isAdmin?: boolean
   url?: string
-}> = ({ username, isAdmin, url }) => {
+  menuItems?: items[]
+}> = ({ username, isAdmin, url, menuItems }) => {
   return (
-    <div className="grid grid-rows-2 grid-cols-[auto_1fr] gap-x-3 items-center justify-start cursor-pointer">
-      <div className="col-start-1 row-span-2">
-        <Avatar>
-          {url && <AvatarImage src={url} alt="user" />}
-          {username && <AvatarFallback>{getInitials(username)}</AvatarFallback>}
-        </Avatar>
-      </div>
-      <p className="col-start-2 row-start-1 text-xs text-primary font-medium">{username}</p>
-      <p className="col-start-2 row-start-2 text-xs font-normal text-tertiary-background">
-        {isAdmin ? 'Admin' : 'Account Member'}
-      </p>
-    </div>
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <div className="group relative grid grid-rows-2 grid-cols-[auto_1fr] gap-x-3 items-center justify-start cursor-pointer">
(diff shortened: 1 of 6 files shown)


joetaylor-harness:
This is a temporary fix for https://harness.atlassian.net/browse/PIPE-22462. 

We are working on a new navbar implementing ShadCn `sidebar`, which will achieve something similar.

packages/canary/src/components/navbar-user.tsx
```diff
@@ -1,25 +1,55 @@
...
+          <div className="absolute -inset-2 rounded-md group-hover:bg-primary/5 ease-in-out duration-100" />
+          <div className="col-start-1 row-span-2">
+            <Avatar className="rounded-md overflow-hidden">
+              {url && <AvatarImage src={url} alt="user" />}
+              {username && <AvatarFallback>{getInitials(username)}</AvatarFallback>}
+            </Avatar>
+          </div>
+          <p className="col-start-2 row-start-1 text-xs text-primary font-medium">{username}</p>
+          <p className="col-start-2 row-start-2 text-xs font-normal text-tertiary-background">
+            {isAdmin ? 'Admin' : 'Account Member'}
```
abhinavrastogi-harness, line 37: there can be other roles potentially. let's not assume just two. it'd be better to pass-in the role as a prop and display it as-is.

packages/playground/src/layouts/RootLayout.tsx
```diff
@@ -181,13 +181,21 @@ export const RootLayout: React.FC<RootLayoutProps> = ({ currentUser }) => {
...
             <Navbar.Footer>
-              <NavLink to="/sandbox/settings/profile/general" className="p-2 hover:bg-tertiary">
-                <NavbarUser.Root
-                  username={currentUser ? currentUser.display_name : 'Steven M.'}
-                  isAdmin={currentUser ? currentUser.admin : false}
-                  url={currentUser ? undefined : '../images/user-avatar.svg'}
-                />{' '}
-              </NavLink>
+              <NavbarUser.Root
+                username={currentUser ? currentUser.display_name : 'Steven M.'}
```
abhinavrastogi-harness, line 185: is this RootLayout used in Gitness? if so, we should not have hardcoded fallbacks like `Steven M`. Let's fallback to user email if username is not available.

packages/playground/src/layouts/SandboxRoot.tsx
```diff
@@ -175,13 +175,21 @@ export const SandboxRoot: React.FC<SandboxRootProps> = ({ currentUser }) => {
...
+              isAdmin={currentUser ? currentUser.admin : false}
+              url={currentUser ? undefined : '../images/user-avatar.svg'}
+              menuItems={[
+                {
+                  key: 0,
+                  element: <Link to="/sandbox/settings/profile/general">Settings</Link>
+                },
+                {
+                  key: 1,
+                  element: <Link to="logout">Log out</Link>
```
abhinavrastogi-harness, line 189: should this be `/logout`? otherwise it won't work on deeply nested pages

packages/canary/src/components/navbar-user.tsx
sans-harness: Currently the API gives us the user where admin is a boolean. 

```
export interface TypesUser {
    admin?: boolean;
    blocked?: boolean;
    created?: number;
    display_name?: string;
    email?: string;
    uid?: string;
    updated?: number;
}
```

Hence the current implementation checks for `admin` and if not assigns `Account Member`. I think we should keep this for now until we get more roles added because this aligns well with the backend for now.

abhinavrastogi-harness: Agreed, that might be true for the gitness backend, but if we decide to re-use this component for Harness, this won't apply. Ideally I'd like Playground components to be independent of Gitness API specifics.
But in either case, this is not a blocker.

abhinavrastogi-harness (approved):

PR #341, opened 2024-10-22
split run pipeline dialog and form

srdjan-harness:
- split run pipeline dialog and form
- add isInitialized state - to prevent buttons flickering

### DIFF
--- a/apps/gitness/src/pages/run-pipeline-dialog/run-pipeline-dialog.tsx
+++ b/apps/gitness/src/pages/run-pipeline-dialog/run-pipeline-dialog.tsx
@@ -1,167 +1,21 @@
-import { useEffect, useState } from 'react'
-import { parse } from 'yaml'
-import { omit } from 'lodash-es'
-import { inputComponentFactory, InputType } from '@harnessio/playground'
-import { useNavigate, useParams } from 'react-router-dom'
-import {
-  Spacer,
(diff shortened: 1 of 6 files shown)


apps/gitness/src/pages/run-pipeline-dialog/run-pipeline-dialog.tsx
```diff
@@ -1,167 +1,21 @@
...
-        </Dialog>
-      )}
-    </RootForm>
+    <Dialog
+      open={open}
+      onOpenChange={open => {
+        if (!open) onClose()
+      }}>
+      <DialogContent className="max-w-[500px] bg-primary-background border-border">
+        <RunPipelineForm {...rest} requestClose={() => onClose()} />
```
cjlee01, line 17: quick question: why do we use {..rest} here

srdjan-harness: We are destructuring props `...{ onClose, open, ...rest }: RunPipelineDialogProps...`
so `rest` contains all props except `onClose, open`. If you look at the `RunPipelineDialogProps` type you will see that it extends `RunPipelineFormProps`. This means that  `rest` contains all props defined in the `RunPipelineFormProps` type.

PR #344, opened 2024-10-23
Playground: further update to navbar

joetaylor-harness:
- Figmas show the item vertical spacing has changed (gone tighter), matched
- Used Harness text svg to match font to Figma (this is coming via <Icon /> for simplicity, but arguably this should not be the case
- Made search box sticky


joetaylor-harness:
image

packages/canary/src/components/icon.tsx
```diff
@@ -92,6 +92,7 @@ import SshKey from '../icons/ssh-key.svg'
 import GreenCircle from '../icons/green-circle.svg'
 import GreenTick from '../icons/green-tick-nobg.svg'
 import CancelGrey from '../icons/cancel-grey.svg'
+import HarnessLogoText from '../icons/harness-logo-text.svg'
```
abhinavrastogi-harness, line 95: can we directly import and render this svg? this is technically not an "icon", and keeping it separate will probably be better for the long-term. not a blocker though.

abhinavrastogi-harness (approved):

cjlee01 (approved):

packages/canary/src/components/icon.tsx
c-joetaylor-harness: agreed, however to run in Playground we'd need to alter the vite config to allow it, I believe.

PR #345, opened 2024-10-24
feat: [pipe-22241]: add compare btn on branch list

cjlee01:
add compare btn on branch list to create new pr based on branch selected
image

### DIFF
--- a/packages/playground/src/components/branches-list.tsx
+++ b/packages/playground/src/components/branches-list.tsx
@@ -49,9 +49,9 @@ interface BranchProps {
 
 interface PageProps {
   branches: BranchProps[]
-  spaceId?: string
-  repoId?: string
-  defaultBranch?: string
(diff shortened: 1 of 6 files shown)


packages/playground/src/components/branches-list.tsx
```diff
@@ -42,9 +49,55 @@ interface BranchProps {
 
 interface PageProps {
   branches: BranchProps[]
+  spaceId?: string
+  repoId?: string
+  defaultBranch?: string
```
sans-harness, line 54: do these need to be optional?

sans-harness (approved):

packages/playground/src/components/branches-list.tsx
cjlee01: they didn't fixed this

PR #350, opened 2024-10-25
Unsaved changes prompt

srdjan-harness:
Unsaved changes prompt implementation (ExitConfirmContext and useExitPrompt). 

Other:
- first input focus in the Run Pipeline form
- fallback for requestIdleCallback - for Safari browser


apps/gitness/src/components/ExitConfirmDialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { ExitConfirmOptions } from '../framework/context/ExitConfirmContext'
+
+export type ExitConfirmDialogProps = ExitConfirmOptions & { open: boolean }
+
+export function ExitConfirmDialog({
+  open,
+  title,
+  subtitle,
+  cancelText,
+  confirmText,
```
abhinavrastogi-harness, line 19: these should have some default values

apps/gitness/src/framework/context/ExitConfirmContext.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { createContext, useCallback, useMemo, useState } from 'react'
+import { ExitConfirmDialog } from '../../components/ExitConfirmDialog'
+
+export interface ExitConfirmOptions {
+  title?: string
+  subtitle?: string
+  confirmText?: string
+  cancelText?: string
+  onConfirm?: () => void
+  onCancel?: () => void
```
abhinavrastogi-harness, line 10: I think these callbacks should be required. the display strings can have default values to make them optional

apps/gitness/src/framework/hooks/useExitPrompt.tsx
```diff
@@ -0,0 +1,69 @@
+import { useCallback, useEffect } from 'react'
```
abhinavrastogi-harness, line 1: this can be a `.ts` file

apps/gitness/src/framework/hooks/useExitPrompt.tsx
```diff
@@ -0,0 +1,69 @@
...
+  confirmText?: string
+  cancelText?: string
+}
+
+export const useExitPrompt = ({
+  isDirty = false,
+  title = 'You have unsaved changes',
+  subtitle = 'Are you sure you want to leave this page without saving?',
+  confirmText = 'Leave',
+  cancelText = 'Stay'
```
abhinavrastogi-harness, line 17: these defaults should be on the component I think.. not on the hook

apps/gitness/src/pages/run-pipeline-dialog/run-pipeline-form.tsx
```diff
@@ -112,6 +114,16 @@ export default function RunPipelineForm({
         // TODO: error toast here ?
       })
   }
+
+  const formRef = useRef<HTMLDivElement | null>(null)
+  // NOTE: custom focus implementation
```
abhinavrastogi-harness, line 119: why do we need a custom focus implementation?
can we just set `autofocus` attribute on the first form input instead?

abhinavrastogi-harness (approved):

apps/gitness/src/components/ExitConfirmDialog.tsx
srdjan-harness: Updated. Defaults are set in component.

apps/gitness/src/framework/context/ExitConfirmContext.tsx
srdjan-harness: Updated. Callbacks are mandatory.

apps/gitness/src/pages/run-pipeline-dialog/run-pipeline-form.tsx
srdjan-harness: Fixed. We are using setFocus that is exposed via useForm in the form library. Added autoFocusPath to RootForm.

PR #356, opened 2024-10-28
Fix reset password flow

jessieyu777:
- Fixed the hap between `icon-success` and the `text` in the button
- Added the flow of resetting the new password
- Deleted unnecessary props in reset password dialog
Screenshot 2024-10-28 at 1 38 22 PM
Screenshot 2024-10-28 at 1 40 17 PM
Screenshot 2024-10-28 at 2 01 00 PM
Screenshot 2024-10-28 at 2 01 06 PM


jessieyu777 (commented):
fix styles in the button and add the flow of the reset password dialog

packages/playground/src/components/project-settings/form-member-edit-dialog.tsx
```diff
@@ -125,8 +125,13 @@ export const FormEditMemberDialog: React.FC<FormEditDialogProps> = ({
...
                     </>
                   ) : (
-                    <Button variant="ghost" type="button" size="sm" theme="success" className="pointer-events-none">
-                      Saved&nbsp;&nbsp;
+                    <Button
+                      variant="ghost"
+                      type="button"
+                      size="sm"
+                      theme="success"
+                      className="flex gap-2 pointer-events-none h-9">
```
sans-harness, line 133: Do we need this custom class for the button? It might not remain consistent across.

packages/playground/src/components/user-management/form-user-reset-password.tsx
```diff
@@ -10,47 +10,53 @@ import {
...
-export const FormResetPasswordDialog: React.FC<FormResetPasswordsDialogProps> = ({
-  user,
-  onReset,
-  onClose,
-  isResetting,
-  resetSuccess
-}) => {
+export const FormResetPasswordDialog: React.FC<FormResetPasswordsDialogProps> = ({ user, onClose }) => {
+  const [isConfirm, setIsConfirm] = React.useState(false)
+
```
sans-harness, line 22: NIT:
```import {useState} from "react"```

File: packages/playground/src/components/user-management/password-copy-action.tsx
Comment (sans-harness) [inline packages/playground/src/components/user-management/password-copy-action.tsx:1]: I think there is already copy button component right? Can we use that directly?

Comment (jessieyu777) [reply packages/playground/src/components/user-management/password-copy-action.tsx:1]: No it's different about copy button since the layout is different. Copy button only shows the copy button

File: packages/playground/src/components/project-settings/form-member-edit-dialog.tsx (hunk 1 above)
Comment (jessieyu777) [reply packages/playground/src/components/project-settings/form-member-edit-dialog.tsx:133]: Yes, please see what I deleted and the gap can show the same style instead of using &nbsp

File: packages/playground/src/components/user-management/password-copy-action.tsx [hunk 3]
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect, useState } from 'react'
+import { Text, Icon, cn } from '@harnessio/canary'
+import copy from 'clipboard-copy'
+import { ShaBadge } from '../..'
+
+export const PasswordCopyAction = ({ password }: { password: string }) => {
```
cjlee01, line 6: is this styled differently than commitactions button?

packages/playground/src/components/user-management/password-copy-action.tsx
sans-harness: Maybe add a picture/explanation about the difference in layout?

packages/playground/src/components/project-settings/form-member-edit-dialog.tsx
sans-harness: probably nitpicking but imho we should just use the non-breaking space in html just so it remains consistent especially if it does the same thing.

packages/playground/src/components/user-management/password-copy-action.tsx
jessieyu777: We don't have this design on Figma right now. This is the design in the orignal gitness. I've added more props for copying password here so might not affect other pages using the commitActions. Please check it out!

Screenshot 2024-10-29 at 11 43 53 AM

@cjlee01

jessieyu777:
Screenshot 2024-10-31 at 9 33 36 AM

packages/playground/src/components/user-management/utils.tsx
```diff
@@ -0,0 +1,11 @@
+export function generateAlphaNumericHash(length: number) {
```
cjlee01, line 1: maybe add this function in the root level utils file rather than here

packages/playground/src/components/commit-copy-actions.tsx
```diff
@@ -12,6 +12,7 @@ export const CommitCopyActions = ({ sha }: { sha: string }) => {
     let timeoutId: number
     if (copied) {
       copy(sha)
+
```
cjlee01, line 15: remove space?

cjlee01 (approved):

PR #363, opened 2024-10-30
feat: [PIPE-22045]: search files in repo/code explorer

shaurya-harness:
adds search implementation to file explorer
adds padding top to codeEditor and diffEditor

Screenshot 2024-10-30 at 4 44 17 PM

### DIFF
--- a/apps/gitness/src/pages/repo/repo-files.tsx
+++ b/apps/gitness/src/pages/repo/repo-files.tsx
@@ -1,17 +1,120 @@
-import { useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { useParams, useNavigate, Outlet } from 'react-router-dom'
 import { BranchSelector, SandboxLayout } from '@harnessio/playground'
-import { Button, ButtonGroup, Icon } from '@harnessio/canary'
+import {
+  Button,
+  ButtonGroup,
+  Command,
+  CommandEmpty,
+  CommandGroup,
+  CommandInput,
+  CommandItem,
+  CommandList,
+  Icon,
+  Popover,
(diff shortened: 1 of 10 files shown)


apps/gitness/src/pages/repo/repo-files.tsx
```diff
@@ -21,24 +120,41 @@ export const RepoFiles: React.FC = () => {
...
   const { data: branches } = useListBranchesQuery({
     repo_ref: repoRef,
-    queryParams: { include_commit: false, sort: 'date', order: 'asc', limit: 20, page: 1, query: '' }
+    queryParams: {
+      include_commit: false,
+      sort: 'date',
+      order: 'asc',
+      limit: 20,
+      page: 1,
+      query: ''
```
cjlee01, line 136: when you add query to this, probably add a debounce so it doesnt rerender multiple times on input change

cjlee01:
what do you show when there's an empty search result? @shaurya-harness

shaurya-harness:
> what do you show when there's an empty search result? @shaurya-harness
@cjlee01 
Screenshot 2024-11-04 at 7 14 06 PM

apps/gitness/src/pages/repo/repo-files.tsx
shaurya-harness: we are not changing this query anywhere as of now

bradrydzewski:
Note that this should be a Textbox (with a search icon) as opposed to a Dropdown.

vardanbansal-harness (approved):

shaurya-harness:
> Note that this should be a Textbox (with a search icon) as opposed to a Dropdown.

yes Brad, will be changing this in a follow up PR

PR #366, opened 2024-10-30
feat: delay showing loader for small duration

abhinavrastogi-harness:
feat: only show skeleton loader if the API is taking longer than a certain time

- add a customisable (default 300ms) delay before showing the skeleton list loader
- fade-in the loader for a smoother experience

### DIFF
--- a/packages/playground/src/components/loaders/skeleton-list.tsx
+++ b/packages/playground/src/components/loaders/skeleton-list.tsx
@@ -8,25 +8,28 @@ const getRandomPercentageWidth = (min: number, max: number) => `${Math.floor(Mat
 const getRandomPixelWidth = (min: number, max: number) => `${Math.floor(Math.random() * (max - min + 1)) + min}px`
 
 interface SkeletonListProps {
-  delay?: number
+  className?: string
 }
 
-export const SkeletonList = ({ delay = 300 }: SkeletonListProps) => {
+export const SkeletonList = ({ className }: SkeletonListProps) => {
   const listItems = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 
-  let startDelayTimer: number
+  // do a first render with visible=false
   const [visible, setVisible] = useState(false)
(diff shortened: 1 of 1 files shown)


packages/playground/src/components/loaders/skeleton-list.tsx
```diff
@@ -1,17 +1,32 @@
...
+      setVisible(true)
+    }, delay)
+    return () => {
+      clearTimeout(startDelayTimer)
+    }
+  }, [])
+
   return (
-    <div className="relative w-full h-full">
+    <div className={cn('relative w-full h-full transition-opacity', { ['opacity-0']: !visible })}>
```
joetaylor-harness, line 29: Does it need square brackets ['opacity-0'] ?

joetaylor-harness, line 29: Also perhaps needs an easing function and duration for the transition, e.g. 'duration-100 ease-in-out'. Actually perhaps you don't need the timeout stuff at all since you can set the transition-delay in tailwind.

joetaylor-harness: eg. 

`<div
      className={cn(
        'relative w-full h-full opacity-0 transition-opacity duration-150',
        `delay-${delay}`,
        'opacity-100'
      )}
    >`
    
 Only thing to check via typing is that delay value is an preset tailwind delay-X value

joetaylor-harness: Actually no, you will need a state approach otherwise the opacity will always be 100% using pure tailwind. Unless you use an animation with start opacity-0 and end opacity-100 and use forwards.

joetaylor-harness (approved):

packages/playground/src/components/loaders/skeleton-list.tsx
abhinavrastogi-harness: yes, still need state, but removed the timer

PR #368, opened 2024-10-31
Webhooks create page

sans-harness:
- Adds the `create-webhooks` page to playground. 

https://github.com/user-attachments/assets/dee8ce3f-3806-48ce-b81a-ad74072b7d98

### DIFF
--- a/packages/playground/src/components/webhooks/create-webhook-form-data.ts
+++ b/packages/playground/src/components/webhooks/create-webhook-form-data.ts
@@ -1,0 +1,23 @@
+import { WebhookEvent } from './types'
+
+export const branchEvents: WebhookEvent[] = [
+  { id: 1, event: 'Branch created' },
+  { id: 2, event: 'Branch updated' },
+  { id: 3, event: 'Branch deleted' }
+]
+
+export const tagEvents: WebhookEvent[] = [
+  { id: 4, event: 'Tag created' },
+  { id: 5, event: 'Tag updated' },
+  { id: 6, event: 'Tag deleted' }
+]
+
+export const prEvents: WebhookEvent[] = [
+  { id: 7, event: 'PR created' },
+  { id: 8, event: 'PR updated' },
+  { id: 9, event: 'PR opened' },
+  { id: 10, event: 'PR branch updated' },
+  { id: 11, event: 'PR closed' },
+  { id: 12, event: 'PR comment created' },
+  { id: 13, event: 'PR merged' }
+]
--- a/packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
+++ b/packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
@@ -1,28 +1,9 @@
-import React, { useState } from 'react'
-import {
-  Input,
-  Textarea,
-  Text,
-  DropdownMenu,
-  DropdownMenuItem,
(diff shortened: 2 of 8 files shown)


packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
```diff
@@ -0,0 +1,166 @@
...
+    <Input id="payloadUrl" {...register!('payloadUrl')} placeholder="https://example.com/harness" />
+    {errors!.payloadUrl && (
+      <FormFieldSet.Message theme={MessageTheme.ERROR}>{errors!.payloadUrl.message?.toString()}</FormFieldSet.Message>
+    )}
+  </FormFieldSet.ControlGroup>
+)
+
+export const WebhookSecretField: React.FC<FieldProps> = ({ register, errors }) => (
+  <FormFieldSet.ControlGroup>
+    <FormFieldSet.Label htmlFor="secret" required>
```
abhinavrastogi-harness, line 69: should secret be required? github allows webhooks without a secret. does gitness not allow the same?

packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
```diff
@@ -0,0 +1,166 @@
...
+            />
+          </div>
+        }
+        right
+      />
+    </StackedList.Item>
+  </StackedList.Root>
+)
+
+export const WebhookNameField: React.FC<FieldProps & { disabled: boolean }> = ({ register, errors, disabled }) => (
```
abhinavrastogi-harness, line 31: why do we need so much boilerplate for all the form fields? are we not using the schema framework which can render basic type inputs out-of-the-box?
cc @srdjan-harness

packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
```diff
@@ -0,0 +1,166 @@
...
+export const WebhookSSLVerificationField: React.FC<FieldProps> = ({ watch, setValue }) => {
+  const sslVerificationValue = watch!('sslVerification')
+  const handleAccessChange = (value: string) => {
+    setValue!('sslVerification', value as '1' | '2')
+  }
+
+  return (
+    <FormFieldSet.Root className="mb-0">
+      <FormFieldSet.ControlGroup>
+        <FormFieldSet.Label htmlFor="sslVerification" required>
```
abhinavrastogi-harness, line 88: same here. why is this required?

packages/playground/src/components/webhooks/create-webhook-form-data.ts
```diff
@@ -0,0 +1,23 @@
...
+  { id: 3, event: 'Branch deleted' }
+]
+
+export const tagEvents: WebhookEvent[] = [
+  { id: 4, event: 'Tag created' },
+  { id: 5, event: 'Tag updated' },
+  { id: 6, event: 'Tag deleted' }
+]
+
+export const prEvents: WebhookEvent[] = [
```
abhinavrastogi-harness, line 15: can we reuse the enums from the types file instead of duplicating all the strings?

sans-harness: Already fixed in followup PR

PR #369, opened 2024-11-01
feat: Add pagination on few more pages, cleanup conditional rendering logic

vardanbansal-harness:
Integrates pagination on following pages:

1. Tokens and Keys > SSH keys
2. Pull Request Details > Commits

Also, cleans up Pagination component to avoid conditional rendering based on presence/absence of pagination response headers.

### DIFF
--- a/apps/gitness/src/pages/profile-settings/profile-settings-keys-page.tsx
+++ b/apps/gitness/src/pages/profile-settings/profile-settings-keys-page.tsx
@@ -7,58 +7,32 @@ import {
   ProfileKeysList,
   KeysList,
   ProfileTokensList,
-  TokensList
+  TokensList,
+  PaginationComponent
 } from '@harnessio/playground'
-import { ListPublicKeyErrorResponse, useListPublicKeyQuery } from '@harnessio/code-service-client'
-import { AlertDeleteParams, ApiErrorType } from './types'
+import { AlertDeleteParams } from './types'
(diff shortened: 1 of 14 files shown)


apps/gitness/src/pages/profile-settings/profile-settings-keys-page.tsx
```diff
@@ -17,15 +20,19 @@ interface SandboxSettingsAccountKeysPageProps {
...
   tokens,
   openTokenDialog,
   openSshKeyDialog,
   openAlertDeleteDialog,
-  error
+  error,
+  headers
 }) => {
+  const [page, setPage] = useQueryState('page', parseAsInteger.withDefault(1))
+  const totalPages = parseInt(headers?.get(PageResponseHeader.xTotalPages) || '')
```
sans-harness, line 35: ideally we should have no state management on a rendering page. We should do this in the `container` and just pass props/callbacks.

vardanbansal-harness: `container` can have multiple api calls in it and each of those api calls will have associated response headers. Keeping this in the relevant rendering page would just keep it cleaner.

sans-harness: Sounds good

sans-harness (approved):

PR #370, opened 2024-11-01
Integrate create webhooks flow to gitness

sans-harness:
- Integrates the `create-webhook` flow with APIs and routing to `Gitness`. 
- Edit/Delete `webhook` will be taken up later. 
- Demoed in standup on 11/1.

### DIFF
--- a/apps/gitness/src/pages/webhooks/create-webhook-container.tsx
+++ b/apps/gitness/src/pages/webhooks/create-webhook-container.tsx
@@ -1,23 +1,27 @@
-import { useState } from 'react'
-import { RepoWebhooksCreatePage, CreateWebhookFormFields, SSLVerificationEnum } from '@harnessio/playground'
+import {
+  RepoWebhooksCreatePage,
+  CreateWebhookFormFields,
+  SSLVerificationEnum,
+  BranchEvents,
+  PREvents,
+  TagEvents
+} from '@harnessio/playground'
 import { useNavigate } from 'react-router-dom'
-import { useCreateWebhookMutation, CreateWebhookErrorResponse } from '@harnessio/code-service-client'
+import { useCreateWebhookMutation } from '@harnessio/code-service-client'
 import { useGetRepoRef } from '../../framework/hooks/useGetRepoPath'
 export const CreateWebhookContainer = () => {
   const repo_ref = useGetRepoRef()
   const navigate = useNavigate()
-  const [apiError, setApiError] = useState<string | null>(null)
 
-  const { mutate: createWebHook, isLoading: creatingWebHook } = useCreateWebhookMutation(
+  const {
+    mutate: createWebHook,
+    isLoading: creatingWebHook,
+    error: createWebHookError
+  } = useCreateWebhookMutation(
     { repo_ref: repo_ref },
     {
       onSuccess: () => {
-        setApiError(null)
         navigate(`../webhooks`)
-      },
-      onError: (error: CreateWebhookErrorResponse) => {
-        const message = error.message || 'An unknown error occurred.'
-        setApiError(message)
       }
     }
   )
@@ -29,7 +33,11 @@ export const CreateWebhookContainer = () => {
       url: data.url,
       enabled: data.enabled,
       insecure: data.insecure === SSLVerificationEnum.DISABLED,
-      triggers: [...(data.branchEvents ?? []), ...(data.tagEvents ?? []), ...(data.prEvents ?? [])]
+      triggers: ([] as (BranchEvents | TagEvents | PREvents)[]).concat(
+        data.branchEvents ?? [],
+        data.tagEvents ?? [],
+        data.prEvents ?? []
+      )
     }
 
     createWebHook({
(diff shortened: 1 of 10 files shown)


apps/gitness/src/pages/webhooks/create-webhook-container.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  )
+
+  const onSubmit = (data: CreateWebhookFormFields) => {
+    const webhookRequest = {
+      identifier: data.identifier,
+      description: data.description,
+      url: data.url,
+      enabled: data.enabled,
+      insecure: data.insecure === '2',
```
vardanbansal-harness, line 31: That's a weird looking check 😅

packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
```diff
@@ -28,67 +28,67 @@ export const WebhookToggleField: React.FC<FieldProps> = ({ register, watch, setV
...
   </FormFieldSet.ControlGroup>
 )
 
-export const WebhookSSLVerificationField: React.FC<FieldProps> = ({ watch, setValue }) => {
-  const sslVerificationValue = watch!('sslVerification')
+export const WebhookSSLVerificationField: React.FC<WebhookFormFieldProps> = ({ watch, setValue }) => {
+  const sslVerificationValue = watch!('insecure')
   const handleAccessChange = (value: string) => {
-    setValue!('sslVerification', value as '1' | '2')
+    setValue!('insecure', value as '1' | '2')
```
vardanbansal-harness, line 82: nit: Anything that we can do improve readability of this? `value as '1' | '2'`

apps/gitness/src/pages/webhooks/create-webhook-container.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const CreateWebhookContainer = () => {
+  const repo_ref = useGetRepoRef()
+  const navigate = useNavigate()
+  const [apiError, setApiError] = useState<string | null>(null)
+
+  const { mutate: createWebHook, isLoading: creatingWebHook } = useCreateWebhookMutation(
+    { repo_ref: repo_ref },
+    {
+      onSuccess: () => {
+        setApiError(null)
```
abhinavrastogi-harness, line 15: don't we get access to `error` directly in `const { mutate, isLoading, error } = useCreate()`?
why do we need to maintain state locally for error?

apps/gitness/src/pages/webhooks/create-webhook-container.tsx
```diff
@@ -0,0 +1,55 @@
...
+  )
+
+  const onSubmit = (data: CreateWebhookFormFields) => {
+    const webhookRequest = {
+      identifier: data.identifier,
+      description: data.description,
+      url: data.url,
+      enabled: data.enabled,
+      insecure: data.insecure === '2',
+      triggers: [...(data.branchEvents ?? []), ...(data.tagEvents ?? []), ...(data.prEvents ?? [])]
```
abhinavrastogi-harness, line 32: 🧓 too modern for me.. would this also work to improve readability? 
`Array.concat(data.branchEvents, data.tagEvents, data.prEvents)`?

packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
```diff
@@ -28,67 +28,67 @@ export const WebhookToggleField: React.FC<FieldProps> = ({ register, watch, setV
   </StackedList.Root>
 )
 
-export const WebhookNameField: React.FC<FieldProps & { disabled: boolean }> = ({ register, errors, disabled }) => (
+export const WebhookNameField: React.FC<WebhookFormFieldProps & { disabled: boolean }> = ({
```
abhinavrastogi-harness, line 31: any thoughts on if the form framework can render most this form without requiring custom components?

sans-harness: Can be done, but it would be one huge file with 500 lines of code (harder to manage errors, any speicifc changes etc), so i broke it up and added them as separate components called "fields" . All of these form components exist in the same file.  

imho it's easier to debug and maintain when each field exists in a silo of its own.

apps/gitness/src/pages/webhooks/create-webhook-container.tsx
sans-harness: fixed!

packages/playground/src/components/webhooks/create-webhooks-form-fields.tsx
sans-harness: removed. Added in schema.

apps/gitness/src/pages/webhooks/create-webhook-container.tsx
sans-harness: fixed.

apps/gitness/src/pages/webhooks/create-webhook-container.tsx
sans-harness: fixed.

cjlee01 (approved):

PR #387, opened 2024-11-06
Integrates API's for user management

sans-harness:
- Integrates all APIs for user management page. 
- `Create user` will be implemented in a followup PR. 

e2e vid:

https://github.com/user-attachments/assets/9626c9c3-797f-4ffe-8929-c13f73c863f3

### DIFF
--- a/apps/gitness/src/user-management/user-management-container.tsx
+++ b/apps/gitness/src/user-management/user-management-container.tsx
@@ -1,0 +1,53 @@
+import { useState } from 'react'
+import { SandboxSettingsUserManagementPage } from '@harnessio/playground'
+import {
+  useAdminListUsersQuery,
+  AdminListUsersOkResponse,
+  useAdminUpdateUserMutation
+} from '@harnessio/code-service-client'
+// import { useAppContext } from '../framework/context/AppContext'
+
+export const UserManagementPageContainer = () => {
+  const [userData, setUserData] = useState<AdminListUsersOkResponse | null>(null)
+  // const { currentUser } = useAppContext()
+  // console.log(currentUser)
+
+  useAdminListUsersQuery(
+    {
+      queryParams: {
+        page: 1,
+        limit: 30
+      }
+    },
+    {
+      onSuccess: ({ body: data }) => {
+        setUserData(data)
+      },
+      onError: error => {
+        console.error(error)
+      }
+    }
+  )
+
+  const { mutate: updateUser } = useAdminUpdateUserMutation(
+    {},
+    {
+      onSuccess: ({ body: data }) => {
+        // setUserData(data)
+        console.log(data)
+      },
+      onError: error => {
+        console.error(error)
+      }
+    }
+  )
+
+  const handleUpdateUser = (userId: string, data: any) => {
+    updateUser({
+      user_uid: userId,
+      body: data
+    })
+  }
+
+  return <SandboxSettingsUserManagementPage userData={userData} handleUpdateUser={handleUpdateUser} />
+}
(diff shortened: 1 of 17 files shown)


apps/gitness/src/user-management/user-management-container.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { parseAsInteger, useQueryState } from 'nuqs'
+import { PageResponseHeader } from '../types'
+
+export const UserManagementPageContainer = () => {
+  const queryClient = useQueryClient()
+
+  const { query: _currentQuery, sort } = useCommonFilter<AdminListUsersQueryQueryParams['sort']>()
+  // const [query, _] = useQueryState('query', { defaultValue: currentQuery || '' })
+  const [page, setPage] = useQueryState('page', parseAsInteger.withDefault(1))
+  const changePage = (pageNum: number) => setPage(pageNum)
```
abhinavrastogi-harness, line 19: how is this different from just calling `setPage` directly wherever needed?

apps/gitness/src/user-management/user-management-container.tsx
```diff
@@ -0,0 +1,112 @@
...
+        queryClient.invalidateQueries({ queryKey: ['adminListUsers'] })
+      },
+      onError: error => {
+        console.error(error)
+      }
+    }
+  )
+
+  const handleUpdateUser = (data: { email: string; displayName: string; userID: string }) => {
+    updateUser({
```
abhinavrastogi-harness, line 68: can we just do this and void the `onSuccess` callbacks?

```
const { data, refetch: refetchUsers } = useAdminListUsersQuery(...)
const { mutate: updateUser } = useAdminUpdateUserMutation(...)

handleUpdateUser {
   await updateUser(...)
   refetchUsers()
}
```

packages/playground/src/App.tsx
```diff
@@ -322,7 +322,7 @@ const router = createBrowserRouter([
             ]
           },
           {
-            path: 'user-mamagement',
+            path: 'user-management',
```
abhinavrastogi-harness, line 325: let's change this to `/users`

packages/playground/src/components/user-management/users-list.tsx
```diff
@@ -154,7 +155,7 @@ export const UsersList = ({ users, onDelete, onEdit, onRemoveAdmin, onResetPassw
                 <TableCell className="content-center my-6">
                   <div className="flex gap-1.5 items-center justify-end">
                     <Text wrap="nowrap" size={1} truncate className="text-tertiary-background">
-                      {timeAgo(user.created)}
+                      {timeAgo(user.created!)}
```
abhinavrastogi-harness, line 158: this `!` is essentially avoid a nullcheck here, right? how does `timeAgo` handle a non-empty param?

packages/playground/src/layouts/SandboxRoot.tsx
```diff
@@ -159,6 +159,17 @@ export const SandboxRoot: React.FC<SandboxRootProps> = ({ currentUser, currentSp
               </NavLink>
             </Navbar.Group>
           </Navbar.Content>
+          {currentUser?.admin && (
+            <Navbar.Content>
+              <Navbar.Group>
+                <NavLink to={`/users`} className="p-2 hover:bg-primary/5 rounded-md">
```
abhinavrastogi-harness, line 165: this doesn't match the route in App.tsx

apps/gitness/src/user-management/user-management-container.tsx
sans-harness: Since we're passing it down as a prop to the child component, I think passing a callback rather than the `setState` itself is a better pattern.

vardanbansal-harness (approved):

PR #395, opened 2024-11-07
Pipeline download and copy to clipboard

srdjan-harness:


shaurya-harness (approved):

apps/gitness/src/pages/pipeline-edit/components/pipeline-studio-toolbar.tsx
```diff
@@ -23,9 +25,14 @@ export const PipelineStudioToolbar = ({
       {view === 'yaml' && (
         <Topbar.Right>
           <PipelineStudioToolbarActions
-            onCopyClick={() => undefined}
-            onDownloadClick={() => undefined}
-            onEditClick={() => undefined}
+            onCopyClick={() => {
+              copy(yamlRevision.yaml).then(() => {
+                // TODO: toast
```
abhinavrastogi-harness, line 30: instead of toast, we should show a tick mark on the copy button itself

abhinavrastogi-harness (approved):

PR #403, opened 2024-11-07
fix: settings page should not trigger logout

abhinavrastogi-harness:
- user profile should not trigger logout
- cancelling repo creation should not crash
- handle api errors on repo list page
- disable automatic retries for GET calls

### DIFF
--- a/apps/gitness/src/pages/repo-sandbox/repo-sandbox-list.tsx
+++ b/apps/gitness/src/pages/repo-sandbox/repo-sandbox-list.tsx
@@ -1,4 +1,4 @@
-import { Link } from 'react-router-dom'
+import { Link, useNavigate } from 'react-router-dom'
 import { parseAsInteger, useQueryState } from 'nuqs'
 import { Button, Spacer, Text } from '@harnessio/canary'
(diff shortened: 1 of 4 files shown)


apps/gitness/src/pages/repo-sandbox/repo-sandbox-list.tsx
```diff
@@ -26,22 +26,45 @@ const LinkComponent = ({ to, children }: { to: string; children: React.ReactNode
...
   const totalPages = parseInt(headers?.get(PageResponseHeader.xTotalPages) || '')
 
   const renderListContent = () => {
     if (isFetching) return <SkeletonList />
 
+    if (isError)
+      return (
+        <NoData
+          title="Error"
+          description={[error.message || '']}
```
sans-harness, line 58: NIT: Maybe we could provide a default error message in case `error.message` is null/undefined?

sans-harness (approved):

apps/gitness/src/pages/repo-sandbox/repo-sandbox-list.tsx
abhinavrastogi-harness: Good idea. Will add in a follow up PR.

PR #427, opened 2024-11-13
fix: [pipe-23098]: Can do empty comment on reply

cjlee01:
Can do empty comment on reply


vardanbansal-harness (approved):

vardanbansal-harness:
Curious though what would be the significance of an empty comment?

cjlee01:
> Curious though what would be the significance of an empty comment?

We don’t allow empty comments in GitHub or gitness rn
@vardanbansal-harness

PR #443, opened 2024-11-14
feat: [PIPE-23372]: Add pretty check back in gitness playground and views

praneshg239:


abhinavrastogi-harness (approved):
i see only playground and views packages being enabled for prettier here.
what about other packages?

praneshg239:
> i see only playground and views packages being enabled for prettier here. what about other packages?

I removed it for gitness, playground and views earlier and now I added them back in this PR. For forms and unified-pipeline packages prettier will be added in the follow up PRs.

PR #445, opened 2024-11-14
poc: zustand state with mocking in playground

abhinavrastogi-harness:
this is a contrived, but working example of how we can achieve a few things:

1. define view state entirely in the `views` package, along with callbacks for any user interactions
2. `playground` can pass in mock data very simply, with no dependence on zustand or any other state management library
3. `gitness` can implement the same strongly-typed interface in a zustand store to implement encapsulated business logic
4. includes a sample for a different API which hides the zustand store contract from the component

Here's a real-world example of hwo complex a data store can get for a single page: https://github.com/harness/canary/pull/433/files#diff-5be8368c828f150a436361aa4a6c239c1e3febde07d493f85938a663ea3c6d3f

Passing ALL these values and methods as separate props could be very verbose.

### DIFF
--- a/apps/gitness/package.json
+++ b/apps/gitness/package.json
@@ -41,7 +41,8 @@
     "react-dom": "^18.3.1",
     "react-hook-form": "^7.53.0",
     "yaml": "^2.5.0",
-    "zod": "^3.23.8"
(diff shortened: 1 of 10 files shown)


bot2-harness:

PR #451, opened 2024-11-15
Theme fix for Markdown Preview

vivek-harness:
I initially started this PR for fixing the theme issues, but we decided not to do it at the moment.

By the time, I already had the Markdown Preview fix in place. It's a very small change, but an important one, since the UI looked quite distorted in the Light mode before the fix.

Comparison

### DIFF
--- a/packages/views/src/components/theme-provider.tsx
+++ b/packages/views/src/components/theme-provider.tsx
@@ -26,9 +26,11 @@ export function ThemeProvider({
   storageKey = 'canary-ui-theme',
   ...props
 }: ThemeProviderProps) {
+  const getMediaQuery = () => window.matchMedia('(prefers-color-scheme: dark)')
+
   const [theme, setThemeState] = useState<Theme>(() => (localStorage.getItem(storageKey) as Theme) || defaultTheme)
   const [systemTheme, setSystemTheme] = useState<'dark' | 'light'>(() =>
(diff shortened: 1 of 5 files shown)


packages/views/src/components/theme-provider.tsx
```diff
@@ -26,24 +26,40 @@ export function ThemeProvider({
   storageKey = 'canary-ui-theme',
   ...props
 }: ThemeProviderProps) {
+  const getMediaQuery = () => window.matchMedia('(prefers-color-scheme: dark)')
+
   const [theme, setThemeState] = useState<Theme>(() => (localStorage.getItem(storageKey) as Theme) || defaultTheme)
+  const [systemTheme, setSystemTheme] = useState<'dark' | 'light'>(() => (getMediaQuery().matches ? 'dark' : 'light'))
```
sans-harness, line 32: maybe create an enum for themes?

vivek-harness: The system theme will always be either Dark or Light, it basically represents your currently selected laptop theme.

But, once we have multiple combinations for theme selection, we'll come up with a structured way to represent them.

vivek-harness:
Closed this PR as most of the changes already got merged as part of another PR. Created #452 for the actual change that needs to be merged.

PR #453, opened 2024-11-17
fix: [PIPE-23403]: Add import sort prettier plugin to sort the imports

praneshg239:
Added `@ianvs/prettier-plugin-sort-imports` prettier sort plugin. This will sort the imports in the following order.

1. At the top, portion for copyright comments or some other comments in future
2. Linebreak
3. React or React named imports (react-dom, react-router-dom)
4. Linebreak
5. Third party imports (lodash, immer, @ tanstack etc,..)
6. Linebreak
7. Monorepo packages (`@harnessio/*` packages) 
8. Linebreak
9. Local imports

Imported packages and imports within the brackets from a module also gets sorted alphabetically.

### Before
Screenshot 2024-11-17 at 9 04 04 PM

### After
Screenshot 2024-11-17 at 9 04 36 PM


.prettierrc.yml
```diff
@@ -8,4 +8,20 @@ trailingComma: none
 bracketSpacing: true
 bracketSameLine: true
```
knagurski, line 9: Since we're doing a big set of changes already, what's the appetite for changing this to `false`? I find brackets on the same line to make visually determining props and children difficult to do at a glance.

```typescript
<MyComp
  arg1="something"
  arg2="something"
  arg3="something"
  arg4="something"
  arg5="something"
  arg6="something">
  something
</MyComp>
```

Is much harder to read (in my opinion) than...
```typescript
<MyComp
  arg1="something"
  arg2="something"
  arg3="something"
  arg4="something"
  arg5="something"
  arg6="something"
>
  something
</MyComp>
```

Visually scanning, the latter is much easier to immediately see where props end and children begin.

praneshg239: I agree to it. I faced this "searching for closing bracket" at times.

abhinavrastogi-harness: agreed. let's make this change

abhinavrastogi-harness (approved):

PR #457, opened 2024-11-18
Bug fixes / Add fevicon

sans-harness:
- Adds fevicon to `gitness`. 
- Fixes bugs in the `reviewers-list` sidebar on the `compare-pr` page. 

Earlier: 

1. Users with the same display name get selected together
2. User can search for themselves in the list

https://github.com/user-attachments/assets/b054b434-ce4b-4203-af06-7521e618024d

Now: 

https://github.com/user-attachments/assets/1203a063-956d-4f24-97d5-1a570c3d55cf


apps/gitness/public/harness.svg
abhinavrastogi-harness, line 1: hmm.. not sure if we want to use a svg file.. few reasons:
1. only ~75% of users' browsers support svg favicons today (see https://caniuse.com/?search=svg%20favicon)
2. this is B&W.. we probably want to use the standard blue logo, unless we intentionally want to distinguish the OSS version from enterprise. 🤔

abhinavrastogi-harness (commented):
let's do separate PRs for favicon and other bugfixes. just for a cleaner git history.

cjlee01 (approved):

PR #460, opened 2024-11-20
Add lint staged to all packages and make its concurrency as 1

praneshg239:
References:
Problem faced: https://github.com/lint-staged/lint-staged/issues/929#issuecomment-1789598499
https://github.com/lint-staged/lint-staged/issues/929#issuecomment-1195595223
https://github.com/lint-staged/lint-staged/issues/929#issuecomment-1211287774

Additional changes:
Added lint staged to all the packages where it is missing currently.

### DIFF
--- a/packages/yaml-editor/package.json
+++ b/packages/yaml-editor/package.json
@@ -19,9 +19,9 @@
     "yalc-push": "yalc push",
     "prepublishOnly": "pnpm build",
     "pretty": "prettier --check ./src",
-    "lint": "eslint ./src"
+    "lint": "eslint ./src",
+    "pre-commit": "lint-staged"
   },
-  "dependencies": {},
   "peerDependencies": {
     "@monaco-editor/react": "4.6.0",
     "monaco-editor": "0.50.0",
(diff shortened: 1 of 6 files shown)


packages/yaml-editor/package.json
```diff
@@ -42,27 +41,35 @@
...
     "mini-css-extract-plugin": "^2.9.0",
     "monaco-editor-webpack-plugin": "^7.1.0",
     "sass-loader": "^14.2.1",
     "style-loader": "^4.0.0",
     "ts-loader": "^9.5.1",
     "typescript": "^4.9.5",
     "url-loader": "^4.1.1",
-    "webpack": "^5.92.1",
-    "webpack-cli": "^5.1.4",
-    "webpack-dev-server": "^5.0.4",
```
knagurski, line 61: Do we really use Webpack _and_ Vite in this project?

praneshg239: not sure, let me check this.

knagurski: Not a problem with this PR, just curious.

knagurski (approved):
Works nicely. Would be interesting to see if we can revisit at some point to see if we can keep some concurrency. But it's not a problem at the moment, so better safe than sorry.

praneshg239:
> Works nicely. Would be interesting to see if we can revisit at some point to see if we can keep some concurrency. But it's not a problem at the moment, so better safe than sorry.

We can still run concurrent checks using `--no-stash` option from lint-staged. But, people are still complaining it saying they are facing some other issues because of that. So it is better to use current solution till we get a proper solution for concurrency in monorepos.

PR #462, opened 2024-11-20
Step form improvements

srdjan-harness:
- Templates API integration
 - Removed not used elements - like filters
 - Added group edit (only group name)
 - Align with style changes

### DIFF
--- a/apps/gitness/package.json
+++ b/apps/gitness/package.json
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "@harnessio/canary": "workspace:*",
-    "@harnessio/code-service-client": "3.2.0",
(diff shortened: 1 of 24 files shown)


apps/gitness/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "@harnessio/canary": "workspace:*",
-    "@harnessio/code-service-client": "3.2.0",
+    "@harnessio/code-service-client": "3.2.1-beta.1",
```
abhinavrastogi-harness, line 22: if the new code-service-client works correctly, let's remove the beta version, and publish 3.2.1 directly

abhinavrastogi-harness (approved):

PR #470, opened 2024-11-22
chore: fix issue on forked repos

cjlee01:
fix issue on forked repos where github actions don't run and the checks dont work


knagurski (approved):
Looks reasonable!

PR #483, opened 2024-11-26
Theming support new architecture

vivek-harness:
Initial changes needed for multi-theme support

This PR is stage 1 of the following ADR:
https://harness.atlassian.net/wiki/spaces/~712020c33cb1c71cf34b2f809a60e4efc310ff/pages/edit-v2/22141108376

https://github.com/user-attachments/assets/71a1021c-64bc-472b-8bbd-431892b2a325


apps/gitness/src/framework/context/ThemeContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+export function ThemeProvider({ children }: { children: React.ReactNode }) {
+  const { theme, setTheme } = useThemeStore(state => ({
+    theme: state.theme,
+    setTheme: state.setTheme
+  }))
+
+  useEffect(() => {
+    const root = window.document.documentElement
+
+    const getMediaQuery = () => window.matchMedia('(prefers-color-scheme: dark)')
```
knagurski, line 34: This will return a new media query each time called and it's called 3 times. Because it's new on each call, it'll not be removed on unmount. Not an issue in practice, but the fix is just to assign to a variable here instead of making it a function you always call.
```suggestion
    const mediaQuery = window.matchMedia('(prefers-color-scheme: dark)')
```

apps/gitness/src/framework/context/ThemeContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+        setTheme(`${getMediaQuery().matches ? ModeType.Dark : ModeType.Light}-${color}-${contrast}` as FullTheme)
+      }
+    }
+
+    getMediaQuery().addEventListener('change', updateSystemTheme)
+
+    return () => {
+      getMediaQuery().removeEventListener('change', updateSystemTheme)
+    }
+  }, [theme, setTheme])
```
knagurski, line 55: Not an issue, but a nice optimisation would be to extract much of the logic for this to an external function to avoid creating a new listener every time the theme changes. The media query listener should be fine being created only once. Maybe consider adding the query into a ref.

apps/gitness/src/framework/context/ThemeContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }),
+    {
+      name: 'canary-ui-theme' // LocalStorage key
+    }
+  )
+)
+
+export const useTheme = () => {
+  const { theme, setTheme } = useThemeStore(state => ({ theme: state.theme, setTheme: state.setTheme }))
+  return { theme, setTheme }
```
knagurski, line 22: Might as well just return it instead of destructering it and then building it back into a new object.
```suggestion
  return useThemeStore(state => ({ theme: state.theme, setTheme: state.setTheme }))
```

apps/gitness/src/framework/context/ThemeContext.tsx
```diff
@@ -0,0 +1,58 @@
...
+      theme: 'system-std-std' as FullTheme, // Default theme
+      setTheme: (newTheme: FullTheme) => set({ theme: newTheme })
+    }),
+    {
+      name: 'canary-ui-theme' // LocalStorage key
+    }
+  )
+)
+
+export const useTheme = () => {
```
knagurski, line 20: Can we add the return type to keep typescript happy and auto-completing?

apps/gitness/src/pages/profile-settings/profile-settings-theme-page.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { SandboxLayout } from '@harnessio/ui/views'
+
+import { useTheme } from '../../framework/context/ThemeContext'
+
+const ProfileSettingsThemePage: React.FC = () => {
+  return (
+    <SandboxLayout.Main hasLeftPanel hasHeader hasSubHeader>
+      <SandboxLayout.Content>
+        <Spacer size={10} />
+        <Text size={5} weight={'medium'}>
```
knagurski, line 11:
```suggestion
        <Text size={5} weight="medium">
```

apps/gitness/src/framework/context/ThemeContext.tsx
vivek-harness: Ahh got it, good to know! I'll fix it.

apps/gitness/src/framework/context/ThemeContext.tsx
vivek-harness: Got it, I moved it to another useEffect with no dependencies so that the event listener is assigned only once and removed once during unmount.

apps/gitness/src/framework/context/ThemeContext.tsx
vivek-harness: Done

apps/gitness/src/framework/context/ThemeContext.tsx
vivek-harness: Done

apps/gitness/src/pages/profile-settings/profile-settings-theme-page.tsx
vivek-harness: Done

PR #484, opened 2024-11-26
fix: start migrating repository viewing pages to the UI

3em:
- Adjusted the structure in ui/views
- Added a general framework for the repository data viewing page
- Minor fixes to the repository list page


packages/ui/src/components/search-files.tsx
```diff
@@ -0,0 +1,102 @@
...
+  Popover,
+  PopoverAnchor,
+  PopoverContent,
+  SearchBox,
+  Text
+} from '@/components'
+
+interface SearchFilesProps {
+  navigateToFile: (file: string) => void
+  filesList: string[] | undefined
```
knagurski, line 18:
```suggestion
  filesList?: string[]
```

packages/ui/src/components/search-files.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const handleInputChange = useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value
+    setQuery(value)
+    setIsOpen(value !== '')
+  }, [])
+
+  const filteredFiles = useMemo(() => {
+    return filesList?.filter(file => file.toLowerCase().includes(query.toLowerCase()))
+  }, [query, filesList])
```
knagurski, line 33: You can simplify things, by always making sure an array is returned. Then you can use without optional chaining below. Also, this will run whether or not a query has been entered, so you might as well return quickly in that case.

```suggestion
  const filteredFiles = useMemo<string[]>(() => {
    if (!filesList) return []
    if (!query) return filesList
    
    const queryText = query.toLowerCase()
  
    return filesList.filter(file => file.toLowerCase().includes(queryText))
  }, [query, filesList])
```

packages/ui/src/components/navbar/navbar-user/index.tsx
```diff
@@ -13,9 +13,9 @@ import {
   Icon,
   Text
 } from '@/components'
+import { TypesUser } from '@/types'
```
knagurski, line 16: I know this isn't from this PR, but wow, `TypesUser` is a terrible name.

packages/ui/src/components/command.tsx
```diff
@@ -0,0 +1,130 @@
...
+    {...props}
+  />
+))
+Command.displayName = CommandPrimitive.displayName
+
+const CommandDialog = ({ children, ...props }: DialogProps) => {
+  return (
+    <Dialog {...props}>
+      <DialogContent className="overflow-hidden p-0">
+        <Command className="[&_[cmdk-group-heading]]:text-muted-foreground [&_[cmdk-group-heading]]:px-2 [&_[cmdk-group-heading]]:font-medium [&_[cmdk-group]:not([hidden])_~[cmdk-group]]:pt-0 [&_[cmdk-group]]:px-2 [&_[cmdk-input-wrapper]_svg]:size-5 [&_[cmdk-input]]:h-12 [&_[cmdk-item]]:px-2 [&_[cmdk-item]]:py-3 [&_[cmdk-item]_svg]:size-5">
```
knagurski, line 28: Not an issue, but this is why I don't really like tailwind. This barf of text is extremely difficult to visually reason about and decipher what it does.

packages/ui/src/components/branch-selector.tsx
```diff
@@ -0,0 +1,70 @@
...
+          variant="secondary"
+          size={size}
+          className={cn(widthClasses[width], 'flex items-center gap-1.5 overflow-hidden px-3', {
+            'bg-muted': !prefix,
+            'bg-background': prefix,
+            'border-border': prefix,
+            'border-2': prefix
+          })}
+        >
+          {prefix ? null : <Icon name="branch" size={12} className="text-tertiary-background min-w-[12px]" />}
```
knagurski, line 49: Is there an official spacing unit we can use instead of a specific pixel value of 12?

knagurski (commented):
Some thoughts for you

packages/ui/src/components/branch-selector.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+interface PageProps {
+  name: string
+  branchList?: BranchListProps[]
+  size?: 'default' | 'sm'
+  width?: 'auto' | 'sm' | 'md' | 'lg' | 'full'
+  selectBranch: (branch: string) => void
+  prefix?: string
+}
```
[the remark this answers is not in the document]
andrewgolovanov: @knagurski This is an old component—the new one got lost during the migration. I’m currently working on transferring the new component (Branch Selector) in a separate branch and will address these points during the process :+1:

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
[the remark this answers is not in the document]
shaurya-harness: I think this was done because the sidebar remains common in repo files / edit file / new file page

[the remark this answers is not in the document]
3em: @knagurski At the moment, I have been replicating the same app route structure that currently exists in v1. We added UI components responsible for rendering and integrated them into Gitness following the exact same scheme as already implemented. Why is RepoSidebar used as a persistent layout for all routes under `:spaceId/repos/:repoId/code`? It seems to be because it contains a repeating sidebar for all child routes. Since there will be children in the app router under this path, including RepoSidebar in every child page of `:spaceId/repos/:repoId/code` doesn’t seem necessary.

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const branchList: BranchListProps[] | undefined = branches?.body?.map(item => ({
+    name: item?.name || ''
+  }))
+
+  const { data: filesData } = useListPathsQuery({
+    repo_ref: repoRef,
+    queryParams: { git_ref: normalizeGitRef(selectedBranch) }
+  })
+
+  const filesList = filesData?.body?.files || []
```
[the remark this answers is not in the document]
3em: @knagurski All the logic code was simply transferred from existing files; we didn’t write it ourselves. In this case, it was taken from apps/gitness/src/pages/repo/repo-files.tsx. In this MR, we essentially created a UI transfer for the general structure of repository viewing pages, as our PP team will now be working on various pages for displaying repository data.

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -0,0 +1,121 @@
...
+      setSelectedBranch(repository.body.default_branch)
+    } else if (gitRef) {
+      setSelectedBranch(gitRef)
+    }
+  }, [repository?.body?.default_branch, gitRef])
+
+  const selectBranch = useCallback(
+    (branch: string) => {
+      setSelectedBranch(branch)
+      navigate(`/spaces/${spaceId}/repos/${repoId}/code/${branch}`)
```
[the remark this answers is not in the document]
shaurya-harness: this would probably require a fix at routes in App.tsx as well

packages/ui/src/components/branch-selector.tsx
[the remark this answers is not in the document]
knagurski: Sweet, thanks :-)

packages/ui/src/components/list-actions.tsx
```diff
@@ -34,8 +34,8 @@ function Right({ children }: { children: React.ReactNode }) {
 function Dropdown({ title, items, onChange, selectedValue }: DropdownProps) {
   return (
     <DropdownMenu>
-      <DropdownMenuTrigger className="text-tertiary-background hover:text-primary flex cursor-pointer items-center gap-1.5 duration-100 ease-in-out">
-        {selectedValue && <span className="bg-primary size-[4px] rounded-full"></span>}
+      <DropdownMenuTrigger className="flex cursor-pointer items-center gap-1.5 text-tertiary-background duration-100 ease-in-out hover:text-primary">
+        {selectedValue && <span className="size-[4px] rounded-full bg-primary"></span>}
```
knagurski, line 38: Not a blocker for this PR, but worth making sure we use self-closing tags for empty elements.
```suggestion
        {selectedValue && <span className="size-[4px] rounded-full bg-primary" />}
```

packages/ui/src/components/tabs.tsx
```diff
@@ -0,0 +1,148 @@
...
+          </div>
+        ) : (
+          children
+        )}
+      </TabsContext.Provider>
+    </TabsPrimitive.Root>
+  )
+)
+Tabs.displayName = 'Tabs'
+interface TabsListProps
```
knagurski, line 86: One for the next PR
```suggestion
interface TabsListProps

```

knagurski (approved):
Some things we can fix in the next PR 👍🏻

PR #485, opened 2024-11-26
fix(ui): styles of dropdown checkbox and filters

andrewgolovanov:
**Description:**
This pull request fixes styles in filters and the DropdownMenu subcomponent

### DIFF
--- a/packages/ui/src/components/filters/filters-bar/filter-variants/checkbox.tsx
+++ b/packages/ui/src/components/filters/filters-bar/filter-variants/checkbox.tsx
@@ -45,7 +45,7 @@ const Checkbox = ({ filter, filterOption, onUpdateFilter, searchQueries, handleS
               <Input
                 className="h-6 flex-1 border-none outline-none hover:border-none focus:border-none focus-visible:ring-0"
                 type="text"
-                placeholder={filter.selectedValues.length > 0 ? '' : 'Select one or more options...'}
+                placeholder={filter.selectedValues.length ? '' : 'Select one or more options...'}
                 value={searchQueries.filters[filter.type] || ''}
                 onChange={e => handleSearchChange?.(filter.type, e.target.value, 'filters')}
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/filters/filters-bar/filter-variants/checkbox.tsx
```diff
@@ -50,19 +42,17 @@ const Checkbox = ({ filter, filterOption, onUpdateFilter, searchQueries, handleS
...
-                  onClick={e => {
-                    e.preventDefault()
-                  }}
-                  onKeyDown={e => e.stopPropagation()}
-                />
-              </DropdownMenuItem>
+              <Input
+                className="h-6 flex-1 border-none outline-none hover:border-none focus:border-none focus-visible:ring-0"
+                type="text"
+                placeholder={filter.selectedValues.length > 0 ? '' : 'Select one or more options...'}
```
knagurski, line 48: Not an issue, but you can simplify a bit here.
```suggestion
                placeholder={filter.selectedValues.length ? '' : 'Select one or more options...'}
```

packages/ui/src/components/filters/filters-bar/filter-variants/number.tsx
```diff
@@ -47,32 +46,30 @@ const Number = ({ filter, onUpdateFilter }: NumberFilterProps) => {
...
+        placeholder="Type a value..."
+        onChange={handleInputChange}
+        onClick={e => {
+          e.stopPropagation()
+          e.preventDefault()
+        }}
+      />
+
+      {value && (
+        <button
```
knagurski, line 63: Since this is an icon-only button, can we add an `aria-label` to help describe it for screen readers?
```suggestion
        <button
          aria-label="clear filter"
```

knagurski (approved):
Adding an `aria-label` will make this even better 👍🏻

PR #486, opened 2024-11-27
Add i18n

sans-harness:
- Adds `i18n` with dropdown to toggle language between english, french and system. 

https://github.com/user-attachments/assets/e50afd74-cb4b-40f8-a117-327bd442215f


apps/gitness/src/i18n/i18n.ts
```diff
@@ -0,0 +1,57 @@
...
+    },
+    interpolation: {
+      escapeValue: false
+    }
+  })
+
+export const handleLanguageChange = (lng: string) => {
+  // set system language
+  if (lng === 'system') {
+    const navigatorLang = navigator.language.split('-')[0]
```
praneshg239, line 38: Do we need to use `navigator.userLanguage` as a fallback here? As will IE return user selection as `userLanguage`.  Can we use something like this? 
 
```ts
(navigator.language || navigator.userLanguage || 'en').split('-')[0]
```

packages/ui/src/data/navbar-menu-data.ts
```diff
@@ -1,57 +1,59 @@
 import { MenuGroupType, MenuGroupTypes } from '@components/navbar/types'
 
-export const navbarMenuData: MenuGroupType[] = [
+import { i18nextViewsInstance } from '../i18n/i18n'
+
+export const getNavbarMenuData = (): MenuGroupType[] => [
   {
     groupId: 0,
-    title: 'Devops',
+    title: i18nextViewsInstance.t('component:navbar.devops'),
```
vivek-harness, line 8: Maybe destructure { t } from i18nextViewsInstance and use it instead?

packages/ui/src/data/navbar-menu-data.ts
```diff
@@ -1,57 +1,59 @@
 import { MenuGroupType, MenuGroupTypes } from '@components/navbar/types'
 
-export const navbarMenuData: MenuGroupType[] = [
+import { i18nextViewsInstance } from '../i18n/i18n'
```
vivek-harness, line 3: Can't we use a single '../i18n' import  instead of '../i18n/i18n'?
Need to change the folder structure with i18n/index.ts instead of i18n/i18n.ts

apps/gitness/src/pages/profile-settings/profile-settings-general-container.tsx
```diff
@@ -26,7 +28,10 @@ export const SettingsProfileGeneralPage: React.FC = () => {
         setUserData({
           name: data.display_name || '',
           username: data.uid || '',
-          email: data.email || ''
+          email: data.email || '',
+          language: localStorage.getItem('i18nextLngSystem')
```
vivek-harness, line 32: Try not to repeat localStorage.getItem('i18nextLngSystem') and localStorage.getItem('i18nextLng') multiple times?

packages/ui/src/data/navbar-menu-data.ts
```diff
@@ -1,57 +1,61 @@
...
+
+const { t } = i18nextViewsInstance
```
abhinavrastogi-harness, line 5: not a blocker, but let's try passing this instance in as a prop to the components from `gitness` app.

abhinavrastogi-harness (approved):

PR #488, opened 2024-11-27
feat: initial migration of components for repo summary page

andrewgolovanov:
**Description:**
This pull request introduces the initial structure for the `Repo Summary` page and adds a new component, `Branch Selector`, along with migrating some components from canary that are required for the minimal rendering and display of this page.

**Note**: Components such as `CloneRepoDialog` and `MarkdownViewer` were not migrated in this PR because they require thorough review and are dependent on several components from canary. It makes more sense to address this step-by-step in future PRs as part of the development of this page.

image


packages/ui/src/components/repo-summary-panel.tsx
shaurya-harness, line 1: I was making some changes to this component to support description edit/add, maybe I can add them here once this PR is merged

packages/ui/src/views/repo/repo-summary/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+                  />
+                  <SearchFiles navigateToFile={navigateToFile} filesList={filesList} />
+                </ButtonGroup.Root>
+              </ListActions.Left>
+              <ListActions.Right>
+                <ButtonGroup.Root>
+                  <DropdownMenu>
+                    <DropdownMenuTrigger asChild>
+                      <Button variant="outline">
+                        Add file&nbsp;&nbsp;
```
knagurski, line 140: Probably best to use a bit of left margin on the icon or introduce a flex layout with gap to the button, rather than use non-breaking spaces to push it out

packages/ui/src/components/repo-summary.tsx
```diff
@@ -0,0 +1,154 @@
...
+      </Text>
+      <Text size={2} weight="normal" color="tertiaryBackground">
+        {timestamp}
+      </Text>
+    </ButtonGroup.Root>
+  )
+}
+
+export const Summary = ({ ...props }: PageProps) => {
+  const { latestFile, files } = props
```
knagurski, line 88: No need to use destructuring and rest to just put all the contents of props into props. Either just use it directly, or destructure in the function arguments.
```suggestion
export const Summary = (props: PageProps) => {
  const { latestFile, files } = props
```

or

```suggestion
export const Summary = ({ latestFile, files }: PageProps) => {
```

knagurski: Also, use `FC` to make sure the return type is set for you.
```suggestion
export const RepoSummary: FC<RepoSummaryProps> = ({ latestFile, files }) => {
```

packages/ui/src/components/repo-summary.tsx
knagurski, line 1: This file is very specific to the repos section and so needs to be moved to a `components` directory within the view it is used in. Only globally reusable things should be in the `ui/components`.

packages/ui/src/components/repo-summary-panel.tsx
knagurski, line 1: This is very specific to a particular view, so it needs to be moved to a `components` directory within that view. Only globally reusable components should be added to `ui/components`

packages/ui/src/components/loaders/skeleton-list.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div
+      className={cn(
+        'relative h-full w-full transition-opacity delay-500 duration-500 ease-in-out',
+        {
+          'opacity-0': !visible
+        },
+        className
+      )}
+    >
+      {listItems && listItems.length > 0 && (
```
knagurski, line 36: `listItems` is a local array with content, it will always exist and have a length over 0.
```suggestion
```

packages/ui/src/components/loaders/skeleton-list.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <StackedList.Root>
+          {listItems.map((itm, itm_idx) => (
+            <StackedList.Item key={itm_idx} isLast={listItems.length - 1 === itm_idx}>
```
knagurski, line 39: Not really an issue here since we know the contents, but always avoid using the index of an item in a key unless there is nothing else available.
```suggestion
          {listItems.map(item => (
            <StackedList.Item key={item} isLast={listItems.length === item}>
```

packages/ui/src/components/loaders/skeleton-list.tsx
```diff
@@ -0,0 +1,61 @@
...
+              <StackedList.Field
+                // Randomized percentage width for title skeleton
+                title={<Skeleton className="mb-1.5 h-4" style={{ width: getRandomPercentageWidth(20, 60) }} />}
```
knagurski, line 42: Might be an idea to have these set in a memo; if this component needs to re-render for any reason, the bars will all change length. Not an issue, just a potential oddity.

packages/ui/src/views/repo/repo-summary/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+  repoId,
+  gitRef,
+  latestCommitInfo
+}: RepoSummaryViewProps) {
+  const navigate = useNavigate()
+
+  const renderListContent = () => {
+    if (loading) return <SkeletonList />
+
+    if (!loading && repoEntryPathToFileTypeMap.size > 0) {
```
knagurski, line 92: No need to use `> 0` or check for `!loading`
```suggestion
    if (repoEntryPathToFileTypeMap.size) {
```

packages/ui/src/views/repo/repo-sidebar/index.tsx
```diff
@@ -1,11 +1,11 @@
-import { BranchListProps, BranchSelector, Button, ButtonGroup, Icon, SearchFiles } from '@/components'
+import { BranchSelector, Button, ButtonGroup, Icon, SearchFiles, type BranchSelectorItem } from '@/components'
 import { SandboxLayout } from '@/views'
 
 interface RepoSidebarProps {
   hasHeader?: boolean
   hasSubHeader?: boolean
   selectedBranch: string
-  branchList: BranchListProps[] | undefined
+  branchList: BranchSelectorItem[] | undefined
```
knagurski, line 8: Can this be made optional instead of specifying as possibly `undefined`?
```suggestion
  branchList?: BranchSelectorItem[]
```

packages/ui/src/components/loaders/skeleton-list.tsx
andrewgolovanov: @knagurski I left a comment in the code so that we can revisit this later

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -0,0 +1,225 @@
...
+  )
+
+  const handleCreateToken = () => {
+    const body = {
+      identifier: `code_token_${generateAlphaNumericHash(5)}`
+    }
+    createToken({ body })
+  }
+
+  const repoEntryPathToFileTypeMap = useMemo((): Map<string, OpenapiGetContentOutput['type']> => {
```
knagurski, line 103: Not an issue, but typing the memo or the variable is more consistent then typing the function return in this case.
```suggestion
  const repoEntryPathToFileTypeMap = useMemo<Map<string, OpenapiGetContentOutput['type']>>(() => {
```

or

```suggestion
  const repoEntryPathToFileTypeMap: Map<string, OpenapiGetContentOutput['type']> = useMemo(() => {
```

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -0,0 +1,225 @@
...
+    if (repoDetails?.content?.entries?.length === 0) return new Map()
+    const nonEmtpyPathEntries = repoDetails?.content?.entries?.filter(entry => !!entry.path) || []
```
knagurski, line 105: By changing the line above, you can simplify this since it'll have been proven that an entries array exists with something in it.
```suggestion
    const nonEmptyPathEntries = repoDetails.content.entries.filter(entry => !!entry.path)
```

knagurski: Also there's a typo with `nonEmptyPathEntries`

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -0,0 +1,225 @@
...
+                    type: item?.path
+                      ? getSummaryItemType(repoEntryPathToFileTypeMap.get(item.path))
+                      : SummaryItemType.File,
+                    name: item?.path || '',
+                    lastCommitMessage: item?.last_commit?.message || '',
+                    timestamp: item?.last_commit?.author?.when ? timeAgoFromISOTime(item.last_commit.author.when) : '',
+                    user: { name: item?.last_commit?.author?.identity?.name },
+                    sha: item?.last_commit?.sha && getTrimmedSha(item.last_commit.sha),
+                    path: `/spaces/${spaceId}/repos/${repoId}/code/${gitRef || selectedBranch}/~/${item?.path}`
+                  }) as FileProps
```
knagurski, line 140: Defaulting `name` to `''` will allow you to remove the `as FileProps` below
```suggestion
                    user: { name: item?.last_commit?.author?.identity?.name || '' },
                    sha: item?.last_commit?.sha && getTrimmedSha(item.last_commit.sha),
                    path: `/spaces/${spaceId}/repos/${repoId}/code/${gitRef || selectedBranch}/~/${item?.path}`
                  })
```

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -0,0 +1,225 @@
...
+              )
+            )
+          }
+        })
+        .catch()
+        .finally(() => {
+          setLoading(false)
+        })
+    }
+  }, [repoEntryPathToFileTypeMap.size, gitRef, repoRef, selectedBranch])
```
knagurski, line 150: Looks like there are missing dependencies here. If that's intentional (some might not be dependency-array-safe), then we can suppress the lint warning. If not, then we might have the potential for a bug.

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  useEffect(() => {
+    if (repository) {
+      setSelectedBranch(repository?.default_branch || '')
+    }
+  }, [repository])
+
+  const selectBranch = (branch: string) => {
+    setSelectedBranch(branch)
+  }
```
knagurski, line 160: Probably can drop this and just use `setSelectedBranch` directly

packages/ui/src/components/branch-selector/branch-selector-dropdown.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  return items.filter(item => item.name.toLowerCase().includes(query.toLowerCase().trim()))
+}
+
+export const BranchSelectorDropdown = ({ name, branchList, tagList, selectBranch }: BranchSelectorDropdownProps) => {
+  const [activeTab, setActiveTab] = useState<BranchSelectorTab>(BranchSelectorTab.BRANCHES)
+  const [searchQuery, setSearchQuery] = useState('')
+
+  const sourceItems = activeTab === BranchSelectorTab.BRANCHES ? branchList : (tagList ?? [])
+  const filteredItems = filterItems(sourceItems, searchQuery)
```
knagurski, line 48: It's worth adding this into a `useMemo`. As-is the items will be re-filtered on every render.

packages/ui/src/components/branch-selector/branch-selector-dropdown.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value)
+  }
+
+  const viewAllUrl = activeTab === BranchSelectorTab.BRANCHES ? '' : ''
```
knagurski, line 54: This doesn't seem to do anything 🤔

packages/ui/src/components/branch-selector.tsx
knagurski, line 1: Is this something that's useful outside of a single view or section? If not, then we should shift it to a `components` directory in the view/section it's used in.

packages/ui/src/components/branch-selector/branch-selector-dropdown.tsx
```diff
@@ -0,0 +1,189 @@
...
+  if (!query.trim()) return items
+
+  return items.filter(item => item.name.toLowerCase().includes(query.toLowerCase().trim()))
+}
+
+export const BranchSelectorDropdown = ({ name, branchList, tagList, selectBranch }: BranchSelectorDropdownProps) => {
+  const [activeTab, setActiveTab] = useState<BranchSelectorTab>(BranchSelectorTab.BRANCHES)
+  const [searchQuery, setSearchQuery] = useState('')
+
+  const sourceItems = activeTab === BranchSelectorTab.BRANCHES ? branchList : (tagList ?? [])
```
knagurski, line 47: If the active tab is not branches and tagList doesn't exist, the array will be different on every render, meaning it's not dependency-array-safe. Could we change tagList to be required? Ideally we want an empty array of tags rather than no list at all as it simplifies all the logic.

packages/ui/src/components/branch-selector/branch-selector-dropdown.tsx
```diff
@@ -0,0 +1,189 @@
...
+          showOnFocus
+        />
+      </div>
+
+      <Tabs
+        className="mt-2"
+        variant="branch"
+        value={activeTab}
+        onValueChange={value => {
+          setActiveTab(value as BranchSelectorTab)
```
knagurski, line 77: Not one for now, but we should allow a type to be set for what the tabs are so we can get some type safety instead of just a string. Would have been nice to do...

```typescript
<Tabs<BranchSelectorTab>
  onValueChange={value => {
    setActiveTab(value)
```

One for another day

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -45,10 +44,15 @@ export const RepoSidebar = () => {
...
-  const branchList: BranchListProps[] | undefined = branches?.body?.map(item => ({
-    name: item?.name || ''
-  }))
+  console.log(branches)
+  const branchList: BranchSelectorListProps[] = useMemo(() => {
+    return (
+      branches?.body?.map(item => ({
+        name: item?.name || '',
+        default: item?.name === repository?.body?.default_branch
+      })) || []
```
knagurski, line 53: Nothing wrong here, but you can make it a bit easier to follow by exiting early.
```suggestion
    if (!branches?.body) return []

    return (
      branches.body.map(item => ({
        name: item?.name || '',
        default: item?.name === repository?.body?.default_branch
      }))
```

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -45,10 +44,15 @@ export const RepoSidebar = () => {
       git_ref: normalizeGitRef(selectedBranch)
     }
   })
-
-  const branchList: BranchListProps[] | undefined = branches?.body?.map(item => ({
-    name: item?.name || ''
-  }))
+  console.log(branches)
```
knagurski, line 47: Can we kill this?
```suggestion

```

packages/ui/src/components/commit-copy-actions.tsx
```diff
@@ -1,12 +1,10 @@
 import { useEffect, useState } from 'react'
 import { Link } from 'react-router-dom'
 
+import { Icon, ShaBadge, Text } from '@/components'
+import { cn } from '@/utils/cn'
```
knagurski, line 5: Super minor, but you can drop the first `/`
```suggestion
import { cn } from '@utils/cn'
```

packages/ui/src/components/sha-badge.tsx
```diff
@@ -1,4 +1,4 @@
-import { Text } from '@harnessio/canary'
+import { Text } from '@/components'
 
 interface RootProps {
   children: React.ReactNode
```
knagurski, line 4: Looks like `React` isn't imported in this file. Can we import `{ ReactNode } from 'react'?

packages/ui/src/components/sha-badge.tsx
```diff
@@ -43,10 +43,10 @@ function Icon({ ...props }: IconProps) {
 
   return (
     <div
-      onClick={handleClick && handleClick}
       role="button"
       tabIndex={0}
```
knagurski, line 47: Instead of attaching a click handler to a `div` and try to get it to pretend to be a `button`, can we just use a `button`?

andrewgolovanov:
@knagurski In the next PR for this page, I plan to remove the following components: `Table`, `Tabs`, `Skeleton`, and `BranchSelector` from canary, views, and playground. Unfortunately, this turned out to be quite complex and extensive, so I removed simpler components, such as `SkeletonList`, `CommitCopyActions`, and `ShaBadge`.

packages/ui/src/components/sha-badge.tsx
```diff
@@ -43,10 +43,10 @@ function Icon({ ...props }: IconProps) {
...
-      className="bg-background hover:bg-muted/50 flex items-center border-l px-1.5 py-0.5"
+      onClick={handleClick && handleClick}
```
knagurski, line 48: This is a bit weird and we can simplify with the same result
```suggestion
      onClick={handleClick}
```

packages/ui/src/utils/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+  // Split the name into an array of words, ignoring empty strings
+  const words = name.split(' ').filter(Boolean)
+
+  // Get the initials from the words
+  const initials = words
+    .map(word => word[0].toUpperCase()) // Get the first letter of each word
+    .join('')
+
+  // If length is provided, truncate the initials to the desired length
+  return length ? initials.slice(0, length) : initials
```
knagurski, line 11: I wonder if this logic is enough. Product call, but if someone were to include their middle name, then I'd expect their 1st initial to be from their first name and 2nd to be from their last name (if the length was set to 2, which is probably pretty common).

So for _"Kevin Patrick Nagurski"_, I'd expect to see initials of _"KN"_, not _"KP"_.

One for another day I suppose.

andrewgolovanov: Yes, I think we’ll investigate this in more detail once we get to refactoring the component that uses this function.

packages/ui/src/utils/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Format a timestamp to medium format date (i.e: Jan 1, 2021)
+ * @param timestamp Timestamp
+ * @param dateStyle Optional DateTimeFormat's `dateStyle` option.
+ */
+export function formatDate(timestamp: number | string, dateStyle = 'medium'): string {
+  return timestamp
+    ? new Intl.DateTimeFormat(LOCALE, {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore: TS built-in type for DateTimeFormat is not correct
```
knagurski, line 27: TS's built-in style does work, but most people forget that it's part of the Intl package. This should work...
```suggestion
export function formatDate(timestamp: number | string, dateStyle: Intl.DateTimeFormatOptions['dateStyle'] = 'medium'): string {
  return timestamp
    ? new Intl.DateTimeFormat(LOCALE, {
```

packages/ui/src/utils/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+    : ''
+}
+
+// The timeDistance function calculates the time difference between two dates (in milliseconds)
+//  and returns it in a human-readable format. It can return the difference in days, hours, minutes, and seconds, depending on the onlyHighestDenomination flag.
+// ex) const date1 = new Date('2023-10-01T00:00:00').getTime();
+// const date2 = new Date('2023-10-02T01:30:45').getTime();
+// console.log(timeDistance(date1, date2));
+// Output: "1d 1h 30m 45s"
+export const timeDistance = (date1 = 0, date2 = 0, onlyHighestDenomination = false) => {
```
knagurski, line 39: Can we kill this and use `date-fns` instead?

packages/ui/src/utils/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+    minutes ? minutes + 'm' : hours || days ? '0m' : ''
+  } ${seconds}s`
+}
+
+/**
+ * Formats timestamp, such as, "1708113838167" is formatted as "1 hour ago"
+ * @param timestamp
+ * @returns formatted timestamp
+ */
+export const timeAgo = (timestamp?: number | null): string => {
```
knagurski, line 79: Here too

packages/ui/src/views/repo/components/branch-selector/branch-selector.tsx
```diff
@@ -0,0 +1,55 @@
...
+  branchList,
+  tagList,
+  size = 'default',
+  selectBranch,
+  prefix,
+  className,
+  repoId,
+  spaceId
+}) => {
+  const isTag = tagList.some(tag => tag.name === name) ?? false
```
knagurski, line 25: `some` returns a boolean anyway, so you don't need to default to `false`
```suggestion
  const isTag = tagList.some(tag => tag.name === name)
```

packages/ui/src/components/sha-badge.tsx
```diff
@@ -1,18 +1,20 @@
-import { Text } from '@harnessio/canary'
+import { ReactNode } from 'react'
+
+import { Text } from '@/components'
 
 interface RootProps {
-  children: React.ReactNode
+  children: ReactNode
   className?: string
 }
```
knagurski, line 8: No need to change in this PR, but `PropsWithChildren` is a great way to get the "official" way of saying a component accepts children. You can use it in a fews ways...

```typescript
interface RootProps extends PropsWithChildren<unknown> {
  className?: string
}
```

Or as a type
```typescript
type RootProps = PropsWithChildren<{
  className?: string
}>
```

packages/ui/src/views/repo/repo-sidebar/index.tsx
```diff
@@ -1,11 +1,14 @@
...
+import { BranchSelectorListProps } from '@views/repo'
+import { BranchSelector } from '@views/repo/components'
 
 interface RepoSidebarProps {
   hasHeader?: boolean
   hasSubHeader?: boolean
   selectedBranch: string
-  branchList: BranchListProps[] | undefined
+  branchList: BranchSelectorListProps[]
+  tagList: BranchSelectorListProps[]
```
knagurski, line 11: Would a better name for `BranchSelectorListProps` be `BranchSelectorListItem`? The `Props` suffix should only be used if the type for the props of a component and `BranchSelectorListProps` doesn't appear to be

packages/ui/src/views/repo/repo-summary/components/summary-panel.tsx
```diff
@@ -0,0 +1,75 @@
...
+  Spacer,
+  Text
+} from '@/components'
+
+interface DetailsProps {
+  id: string
+  iconName: 'tube-sign' | 'open-pr' | 'tag' | 'branch' | IconProps['name']
+  name: string
+  count: number
+}
```
knagurski, line 19: These don't appear to be props for a `Details` component, but rather items in an array, maybe `DetailsItem` would be a more appropriate name? Or just `Detail`?
```suggestion
interface DetailsItem {
  id: string
  iconName: 'tube-sign' | 'open-pr' | 'tag' | 'branch' | IconProps['name']
  name: string
  count: number
}
```

packages/ui/src/views/repo/repo-summary/components/summary.tsx
```diff
@@ -0,0 +1,135 @@
...
+  Text
+} from '@/components'
+import { CommitCopyActions } from '@/components/commit-copy-actions'
+import { getInitials } from '@/utils/utils'
+
+import { FileProps, SummaryItemType } from '../repo-summary.types'
+
+interface SummaryProps {
+  latestFile: Pick<FileProps, 'user' | 'lastCommitMessage' | 'timestamp' | 'sha'>
+  files?: FileProps[]
```
knagurski, line 26: Again, these aren't `Props`, can we reserve the `Props` suffix for types that describe the props of a component. In this case, maybe `FileItem` or `RepoFile`?

packages/ui/src/views/repo/repo-summary/components/summary.tsx
```diff
@@ -0,0 +1,135 @@
...
+      <Spacer size={5} />
+      <Table variant="asStackedList">
+        <TableHeader>
+          <TableRow>
+            <TableHead>Name</TableHead>
+            <TableHead>Last commit message</TableHead>
+            <TableHead className="text-right">Date</TableHead>
+          </TableRow>
+        </TableHeader>
+        {files && files.length > 0 ? (
```
knagurski, line 93: Not an issue, but you can simplify a wee bit

```suggestion
        {!!files?.length ? (
```

packages/ui/src/views/repo/repo-summary/components/summary.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <TableBody>
+            {files.map(file => (
+              <TableRow key={file.id} onClick={() => navigate(file.path)}>
```
knagurski, line 96: I think we might need some additional stuff to mark up a `tr` as clickable for a11y

packages/ui/src/views/repo/repo.types.ts
```diff
@@ -0,0 +1,25 @@
+export interface BranchSelectorListProps {
+  name: string
+  default?: boolean
+}
+
+export interface BranchProps {
```
knagurski, line 6: Let's avoid using the `Props` suffix unless it's being used to describe the props to a component.
```suggestion
export interface BranchSelectorListItem {
  name: string
  default?: boolean
}

export interface Branch {
```

packages/ui/src/views/repo/repo-summary/repo-summary.types.ts
```diff
@@ -0,0 +1,20 @@
...
+  Folder = 0,
+  File = 1
+}
+
+export interface UserProps {
+  name: string
+  avatarUrl?: string
+}
+
+export interface FileProps {
```
knagurski, line 11: Same here, only use `Props` when describing the props of a component.
```suggestion
export interface User {
  name: string
  avatarUrl?: string
}

export interface RepoFile {
```

packages/ui/src/views/repo/repo-summary/repo-summary.types.ts
```diff
@@ -0,0 +1,20 @@
+export enum SummaryItemType {
+  Folder = 0,
+  File = 1
```
knagurski, line 3: Totally fine to be specific, but in this case, you can kill the `= 0` and it'll work the same since enums are 0 indexed anyway.
```suggestion
  Folder,
  File
```

packages/ui/src/views/repo/components/branch-selector/branch-selector.tsx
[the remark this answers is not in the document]
andrewgolovanov: Yes, you’re right. I’ve left a TODO in the code to revisit this later and improve the validation logic.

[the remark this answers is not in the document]
andrewgolovanov: Unfortunately, in this case, it will throw an error due to the double negation.
Cursor 2024-11-28 23 17 06

Therefore, I think we can simplify it even further to just `files?.length`

packages/ui/src/views/repo/repo-summary/components/summary.tsx
andrewgolovanov: Unfortunately, in this case, it will throw an error due to the double negation.
Cursor 2024-11-28 23 17 06

Therefore, I think we can simplify it even further to just `files?.length`

andrewgolovanov:
@knagurski Big thanks for the intense review session! :fire:
Please let me know if there are any other critical issues remaining

packages/ui/src/views/repo/repo-summary/components/summary.tsx
knagurski: Absolutely 👍🏻

knagurski:
> @knagurski In the next PR for this page, I plan to remove the following components: `Table`, `Tabs`, `Skeleton`, and `BranchSelector` from canary, views, and playground. Unfortunately, this turned out to be quite complex and extensive, so I removed simpler components, such as `SkeletonList`, `CommitCopyActions`, and `ShaBadge`.

I wonder if we can delete playground entirely at this point to smooth things out a little bit. Seems a bit redundant now that it isn't used at all any more 🤔

knagurski (approved):
Really great work and thanks for sticking in there to make the changes. Things are really shaping up nicely!

PR #491, opened 2024-11-28
I18n single instance - removes views instance of i18n

sans-harness:
- Uses a single `i18n` instance passed down from `gitness`.


apps/gitness/src/i18n/i18n.ts
```diff
@@ -12,17 +14,17 @@ const languageDetectorOptions = {
...
 i18n
   .use(initReactI18next)
   .use(LanguageDetector)
   .init({
     detection: languageDetectorOptions,
-    resources: {},
+    resources,
     fallbackLng: 'en',
-    // debug: true,
+    debug: true,
```
abhinavrastogi-harness, line 27: Intentional to commit this?

apps/gitness/src/i18n/i18n.ts
```diff
@@ -33,6 +35,8 @@ i18n
   })
 
 export const handleLanguageChange = (lng: string) => {
+  console.log('locales', resources)
```
abhinavrastogi-harness, line 38: Remove pls

abhinavrastogi-harness (approved):

PR #495, opened 2024-11-29
feat: [pipe-20761]: add webhook list page

cjlee01:
add webhook list page


apps/gitness/src/pages-v2/webhook/stores/webhook-store.tsx
```diff
@@ -0,0 +1,38 @@
+import { create } from 'zustand'
+
+import { ListRepoWebhooksOkResponse } from '@harnessio/code-service-client'
+import { WebhookType } from '@harnessio/ui/views'
+
+import { timeAgoFromEpochTime } from '../../../pages/pipeline-edit/utils/time-utils'
+import { PageResponseHeader } from '../../../types'
+
+interface WebhookStore {
```
abhinavrastogi-harness, line 9: the interface should be in `ui/views` so that we can mock this data correctly in the new playground

apps/gitness/src/pages-v2/webhook/webhook-list.tsx
```diff
@@ -0,0 +1,66 @@
...
+      setWebhooks(webhookData, headers)
+    }
+  }, [webhookData, headers, setWebhooks])
+
+  useEffect(() => {
+    setQueryPage(page)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [page, queryPage, setPage])
+
+  return <SandboxWebhookListPage useWebhookStore={useWebhookStore} />
```
abhinavrastogi-harness, line 65: can we remove the terminology `Sandbox` for all new code being added?

packages/ui/src/views/webhook/index.ts
abhinavrastogi-harness, line 1: i think the folder should be called `views/webhooks` instead of `views/webhook`

packages/ui/src/views/webhook/webhook-list/repo-webhook-list.tsx
```diff
@@ -0,0 +1,93 @@
...
+  LinkComponent,
+  handleResetFilters,
+  hasActiveFilters,
+  query,
+  handleResetQuery
+}: PageProps) {
+  return (
+    <>
+      <StackedList.Root>
+        {webhooks && webhooks.length > 0 ? (
```
abhinavrastogi-harness, line 34: don't use ternary operators for rendering loading/error/no-data states. and don't depend on `length` of data to determine when data is present.
it causes flashing of irrelevant states. instead use short-circuit if conditions, and use the `isLoading`, `isError` states provided by react-query.

packages/ui/src/views/webhook/webhook-list/repo-webhook-list.tsx
```diff
@@ -0,0 +1,93 @@
...
+                }}
+              />
+            ) : (
+              <NoData
+                iconName="no-data-folder"
+                title="No webhooks yet"
+                description={['There are no webhooks in this project yet.', 'Create new webhooks.']}
+                primaryButton={{
+                  label: 'Create webhook',
+                  onClick: () => {
```
abhinavrastogi-harness, line 82: the view should just take the handler as a prop and call it.. we should not leave it as an unimplemented inline func.

apps/gitness/src/pages-v2/webhook/stores/webhook-store.tsx
cjlee01: done

packages/ui/src/views/webhook/index.ts
cjlee01: done

packages/ui/src/views/webhook/webhook-list/repo-webhook-list.tsx
cjlee01: done

packages/ui/src/views/webhook/webhook-list/repo-webhook-list.tsx
cjlee01: done

packages/ui/src/views/webhooks/webhook-list/repo-webhook-list.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export function RepoWebhookList({
+  webhooks,
+  LinkComponent,
+  handleResetFilters,
+  hasActiveFilters,
+  query,
+  handleResetQuery,
+  handleNavigate,
+  loading
```
vivek-harness, line 33: Can we handle the error state as well?

vivek-harness (approved):

packages/ui/src/views/webhooks/webhook-list/types.ts
```diff
@@ -0,0 +1,67 @@
...
+  updated?: number
+  url?: string
+  version?: number
+}
+
+export declare type EnumWebhookParent = 'repo' | 'space'
+
+export declare type EnumWebhookExecutionResult = 'fatal_error' | 'retriable_error' | 'success' | null
+
+export declare type EnumWebhookTrigger =
```
vivek-harness, line 48: Can we change the type to enum?

packages/ui/src/views/webhooks/webhook-list/repo-webhook-list.tsx
cjlee01: yes

packages/ui/src/views/webhooks/webhook-list/types.ts
cjlee01: it would make sense to keep it as this since its just from api

PR #497, opened 2024-11-29
feat: added repo files to UI

3em:
A part of the components related to the "Files" tab of the repository view has been moved to the UI:
- Tab navigation
- Sidebar with the file tree
- Table


apps/gitness/src/components/FileExplorer.tsx
```diff
@@ -27,6 +27,13 @@ const sortEntriesByType = (entries: OpenapiContentInfo[]): OpenapiContentInfo[]
   })
 }
 
+/**
+ * TODO: This code was migrated from V2.
+ *
+ * @param selectedBranch
+ * @param repoDetails
+ * @constructor
```
knagurski, line 35: These don't really describe this component. This isn't a constructor for a class, so `@constructor` doesn't apply and it takes a single param (props) with 2 items.
```suggestion
```

apps/gitness/src/components/FileExplorer.tsx
```diff
@@ -27,6 +27,13 @@ const sortEntriesByType = (entries: OpenapiContentInfo[]): OpenapiContentInfo[]
   })
 }
 
+/**
+ * TODO: This code was migrated from V2.
```
knagurski, line 31: This doesn't really describe an action that needs done 🤔

apps/gitness/src/pages-v2/repo/repo-code.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { useGetRepoRef } from '../../framework/hooks/useGetRepoPath'
+import { timeAgoFromISOTime } from '../../pages/pipeline-edit/utils/time-utils'
+import { PathParams } from '../../RouteDefinitions'
+import { getTrimmedSha, normalizeGitRef } from '../../utils/git-utils'
+import { splitPathWithParents } from '../../utils/path-utils'
+
+/**
+ * TODO: This code was migrated from V2.
+ * @constructor
+ */
```
knagurski, line 23: Again, this isn't a class, so `@constructor` doesn't apply and the TODO isn't really a thing to do 🤔 
```suggestion
```

apps/gitness/src/pages-v2/repo/repo-code.tsx
```diff
@@ -0,0 +1,148 @@
...
+  useEffect(() => {
+    if (repository && !gitRef) {
+      setSelectedBranch(repository?.default_branch || '')
+    } else if (gitRef) {
+      setSelectedBranch(gitRef)
+    }
+  }, [repository, gitRef])
+
+  const repoEntryPathToFileTypeMap = useMemo((): Map<string, OpenapiGetContentOutput['type']> => {
+    if (repoDetails?.content?.entries?.length === 0) return new Map()
```
knagurski, line 58: This will not return a new `Map` if `repoDetails`, `content` or `entries` are not present. Checking for falseyness will account for that
```suggestion
    if (!repoDetails?.content?.entries?.length) return new Map()
```

apps/gitness/src/pages-v2/repo/repo-code.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const nonEmtpyPathEntries = repoDetails?.content?.entries?.filter(entry => !!entry.path) || []
```
knagurski, line 59: Making the change above will mean that you don't need to make this conditional.
```suggestion
    const nonEmtpyPathEntries = repoDetails.content.entries.filter(entry => !!entry.path)
```

apps/gitness/src/pages-v2/repo/repo-code.tsx
```diff
@@ -0,0 +1,148 @@
...
+    return SummaryItemType.File
+  }
+
+  const getLastPathSegment = (path: string) => {
+    if (!path || /^\/+$/.test(path)) {
+      return ''
+    }
+    path = path.replace(/\/+$/, '')
+    const segments = path.split('/')
+    return segments[segments.length - 1]
```
knagurski, line 76: Not an issue, but you can combine these with `pop`
```suggestion
    return path.split('/').pop()
```

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -102,38 +107,48 @@ export const RepoSidebar = () => {
...
         tagList={[]}
+        // TODO: new props navigateToNewFolder
+        navigateToNewFolder={() => {}}
         navigateToNewFile={navigateToNewFile}
         navigateToFile={navigateToFile}
         filesList={filesList}
-        // repoDetails={repoDetails?.body}
-      />
+      >
+        {repoDetails?.body?.content?.entries?.length && (
```
knagurski, line 148: In the event that `repoDetails.body.content.length` is `0`, this will output `0` in the HTML; worth casting to a boolean to make sure it won't
```suggestion
        {!!repoDetails?.body?.content?.entries?.length && (
```

packages/ui/src/components/accordion.tsx
```diff
@@ -0,0 +1,95 @@
...
+    ref={ref}
+    className="overflow-hidden text-sm data-[state=closed]:animate-accordion-up data-[state=open]:animate-accordion-down"
+    {...props}
+  >
+    <div className={cn('pb-4 pt-0', className)}>{children}</div>
+  </AccordionPrimitive.Content>
+))
+AccordionContent.displayName = AccordionPrimitive.Content.displayName
+
+export { Accordion, AccordionItem, AccordionTrigger, AccordionContent }
```
knagurski, line 95: Might be handy to export the relevant types too

packages/ui/src/components/file-explorer.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+function FolderItem({ children, value, isActive, content, chevronClass, link }: FolderItemProps) {
+  return (
+    <AccordionItem value={value || ''} className="border-none">
+      <AccordionTrigger
+        className="w-full p-0"
+        leftChevron
+        rotateChevron
+        chevronClass={chevronClass || 'text-icons-7 group-hover:text-icons-2 group-data-[state=open]:text-icons-2'}
```
knagurski, line 23: I noticed that in `CommandList` we use `heightClassName`, but here we use `chevronClass`, can we standardise on whether we include the `Name` part across components? I'd say we should use it to fall in line with `className`
```suggestion
        chevronClassName={chevronClassName || 'text-icons-7 group-hover:text-icons-2 group-data-[state=open]:text-icons-2'}
```

packages/ui/src/components/search-box.tsx
```diff
@@ -78,11 +80,9 @@ const Root = forwardRef<HTMLInputElement, SearchBoxProps>(
     }
 
     const handleFocus = () => {
-      if (onFocus) {
-        onFocus()
-      }
+      onFocus?.()
       if (showOnFocus) {
-        handleSearch()
+        handleSearch?.()
```
knagurski, line 85: `handleSearch` will always exist since it's defined on line 71
```suggestion
        handleSearch()
```

packages/ui/src/components/search-files.tsx
```diff
@@ -15,39 +15,42 @@ import {
...
-  const handleInputChange = useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
+  const handleInputChange = useCallback((event: ChangeEvent<HTMLInputElement>) => {
     const value = event.target.value
     setQuery(value)
     setIsOpen(value !== '')
   }, [])
 
   const filteredFiles = useMemo(() => {
-    return filesList?.filter(file => file.toLowerCase().includes(query.toLowerCase()))
+    if (!filesList) return [] as string[]
```
knagurski, line 32: You can drop the type assertion
```suggestion
    if (!filesList) return []
```

packages/ui/src/components/search-files.tsx
```diff
@@ -61,39 +64,36 @@ export const SearchFiles = ({ navigateToFile, filesList }: SearchFilesProps) =>
...
-                    setIsOpen(false)
-                  }}
-                >
-                  {fileText(file)}
-                </CommandItem>
-              ))}
-            </CommandGroup>
+            {filteredFiles && !!filteredFiles.length && (
+              <CommandGroup>
+                {filteredFiles?.map((file: string, idx: number) => (
```
knagurski, line 82: `filteredFiles` will always exist and be an array, so you can simplify
```suggestion
            {!!filteredFiles.length && (
              <CommandGroup>
                {filteredFiles.map((file: string, idx: number) => (
```

packages/ui/src/components/search-files.tsx
```diff
@@ -61,39 +64,36 @@ export const SearchFiles = ({ navigateToFile, filesList }: SearchFilesProps) =>
...
+                  <CommandItem
+                    key={idx}
```
knagurski, line 84: Can we avoid using indexes as keys unless absolutely necessary? The file name itself will be unique, so would make a great key.
```suggestion
                    key={file}
```

packages/ui/src/components/search-files.tsx
```diff
@@ -15,39 +15,42 @@ import {
...
   }, [])
 
   const filteredFiles = useMemo(() => {
-    return filesList?.filter(file => file.toLowerCase().includes(query.toLowerCase()))
+    if (!filesList) return [] as string[]
+
+    return filesList.filter(file => file.toLowerCase().includes(query.toLowerCase()))
   }, [query, filesList])
 
   const fileText = useCallback(
```
knagurski, line 37: This being a callback won't actually help performance at all. `useCallback` is great when you need to stabilise a reference to a function when passing between components (or to a hook), but doesn't offer anything in this case.

If you want to optimise this, then you can preprocess the files list in a `useMemo` so that they start off in the right format and only change if the filesList changes.

packages/ui/src/components/sha-badge.tsx
```diff
@@ -44,13 +44,20 @@ function Icon({ ...props }: IconProps) {
   const { children, handleClick } = props
 
   return (
-    <button
-      className="flex items-center border-l bg-background px-1.5 py-0.5 hover:bg-muted/50"
+    <div
+      role="button"
```
knagurski, line 48: Instead of trying to get a `div` to act like a `button`, let's use an actual `button`

packages/ui/src/components/skeleton-list.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <div
+      className={cn(
+        'relative h-full w-full transition-opacity delay-500 duration-500 ease-in-out',
+        {
+          'opacity-0': !visible
+        },
+        className
+      )}
+    >
+      {listItems && listItems.length > 0 && (
```
knagurski, line 36: `listItems` is guaranteed to exist and have items, so no need to selectively render.
```suggestion
```

packages/ui/src/components/skeleton-list.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <StackedList.Root>
+          {listItems.map((itm, itm_idx) => (
+            <StackedList.Item key={itm_idx} className="py-4" isLast={listItems.length - 1 === itm_idx}>
```
knagurski, line 39: Since we know everything in the `listItems` array is unique, we can avoid using the index.
```suggestion
          {listItems.map(item => (
            <StackedList.Item key={item} className="py-4" isLast={listItems.length === item}>
```

packages/ui/src/views/layouts/SandboxLayout.tsx
```diff
@@ -17,10 +19,7 @@ function Root({ children }: { children: React.ReactNode }) {
 function LeftPanel({ children, className }: { children: React.ReactNode; className?: string }) {
```
knagurski, line 19: Looks like there's a few references to `React` in this, but it's not imported, can we tidy up by importing `ReactNode`?

packages/ui/src/views/layouts/SandboxLayout.tsx
```diff
@@ -88,32 +88,36 @@ function Main({
...
+  const paddingLeftStyle = {
+    paddingLeft:
+      hasLeftPanel && hasLeftSubPanel
+        ? `calc(220px + ${leftSubPanelWidth}px)`
+        : hasLeftPanel
+          ? '220px'
+          : hasLeftSubPanel
+            ? `${leftSubPanelWidth}px`
+            : undefined
+  } as CSSProperties
```
knagurski, line 105: Instead of overriding the type with `as`, set the variable type. `as` tells TS to ignore what it knows and trust you; setting the variable type tells TS to make sure you're correct.
```suggestion
  const paddingLeftStyle: CSSProperties = {
    paddingLeft:
      hasLeftPanel && hasLeftSubPanel
        ? `calc(220px + ${leftSubPanelWidth}px)`
        : hasLeftPanel
          ? '220px'
          : hasLeftSubPanel
            ? `${leftSubPanelWidth}px`
            : undefined
  }
```

packages/ui/src/views/layouts/SandboxRoot.tsx
```diff
@@ -14,6 +14,7 @@ import { SandboxLayout } from '../index'
  * Returns the complete menu model based on an array of IDs,
  * using the navbarMenuData variable as a reference.
  * @param data
+ * @param t
```
knagurski, line 17: Without types, `@param` doesn't really add much

packages/ui/src/views/repo/components/summary/index.tsx
knagurski, line 1: Not a blocker here, but let's reserve `index` files for pulling things together for more organised exports rather than for logic or components. Having a bunch of files open in your IDE all called `index.tsx` makes it hard to find what you're looking for.

packages/ui/src/views/repo/repo-files/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  isShowSummary,
+  latestFile,
+  children
+}: RepoFilesProps) => {
+  const content = useMemo(() => {
+    if (!isDir) return children
+
+    if (loading) return <SkeletonList />
+
+    if (isShowSummary && !!files.length) return <Summary latestFile={latestFile} files={files} />
```
knagurski, line 31: Not an issue, but you don't need to cast to boolean here
```suggestion
    if (isShowSummary && files.length) return <Summary latestFile={latestFile} files={files} />
```

packages/ui/src/views/repo/repo-sidebar/index.tsx
knagurski, line 1: It looks like there are some fixable ESLint issues in this file. When committing, ESLint fix should run and fix them; does it not work for you?

packages/ui/src/views/repo/repo-sidebar/index.tsx
```diff
@@ -1,57 +1,82 @@
...
+              >
+                <Icon size={16} name="add-folder" className="text-icons-3" />
+              </Button>
+              <Button
+                size="icon"
+                variant="ghost"
+                className="rounded-none border-l border-borders-2 p-0"
+                onClick={navigateToNewFile}
+              >
+                <Icon size={16} name="add-file" className="text-icons-3" />
```
knagurski, line 69: When only including an icon in a clickable element (link or button), we need a `aria-label` to describe what the thing does for a11y
```suggestion
              <Button
                size="icon"
                variant="ghost"
                className="rounded-none border-l border-borders-2 p-0"
                onClick={navigateToNewFolder}
                aria-label="Create new folder"
              >
                <Icon size={16} name="add-folder" className="text-icons-3" />
              </Button>
              <Button
                size="icon"
                variant="ghost"
                className="rounded-none border-l border-borders-2 p-0"
                onClick={navigateToNewFile}
                aria-label="Create new file"
              >
                <Icon size={16} name="add-file" className="text-icons-3" />
```

packages/ui/src/views/repo/components/summary/index.tsx
knagurski, line 1: This file was moved to the `components` directory in the relevant view in a previous PR as it was view-specific, but appears to have been moved back. If a component is view-specific, it should live with that view. If it is shared by more than one view in the same section, it should live in a `components` directory in that section.

apps/gitness/src/pages-v2/repo/repo-code.tsx
3em: @knagurski All the code in app/gitness was fully migrated from V1 without any refactoring. Therefore, I added a TODO comment at the top of each file indicating that the code was migrated and needs refactoring. I would be happy to make fixes, but we are not familiar with the implementation details that were written previously.

packages/ui/src/components/accordion.tsx
3em: This file contains only types related to the typing of the component's props. There are no other types that would be needed outside the file, so there is no type export here.

packages/ui/src/views/repo/repo-sidebar/index.tsx
3em: I ran pnpm pre-commit, and the command made no changes (screenshot).
I also always run pnpm lint --fix and pnpm pretty --write.

image

packages/ui/src/views/repo/components/summary/index.tsx
3em: Yes, it is used in multiple sections, in Summary and Files. That’s why I moved it to shared components in `ui/src/views/repo/components`.

packages/ui/src/components/accordion.tsx
knagurski: Types for components are regularly used outside of the component themselves. For example, my favourite way of testing involves this...

```typescript
const renderComponent = (props: Partial<SomeComponentProps> = {}): RenderResult =>
  render(<SomeComponent prop1="default" prop2="default2" {...props} />)

test('it should show prop1', async () => {
  const prop1 = 'hello world'
  renderComponent({ prop1 })

  expect(await screen.findByText(prop1)).toBeInTheDocument()
})
```

You can extract a component's props using TS, but it's pain and most devs will have to Google to figure it out. Having the type exported with the component and with the naming convention of `${COMPONENT_NAME}Props` means they can get at it with no fuss.

This is a very common and encouraged approach in Harness and we need to continue it with this stuff.

apps/gitness/src/pages-v2/repo/repo-code.tsx
```diff
@@ -0,0 +1,148 @@
...
+      setSelectedBranch(repository?.default_branch || '')
+    } else if (gitRef) {
+      setSelectedBranch(gitRef)
+    }
+  }, [repository, gitRef])
+
+  const repoEntryPathToFileTypeMap = useMemo((): Map<string, OpenapiGetContentOutput['type']> => {
+    if (!repoDetails?.content?.entries?.length) return new Map()
+
+    const nonEmptyPathEntries = repoDetails?.content?.entries?.filter(entry => !!entry.path) || []
```
knagurski, line 59: This can be simplified now that you've proven the existence of the `entries` array.
```suggestion
    const nonEmptyPathEntries = repoDetails.content.entries.filter(entry => !!entry.path)
```

packages/ui/src/views/repo/repo-sidebar/index.tsx
knagurski: Ohhh... weird, it looks like we have a misconfiguration of ESLint in the project. When I run fix on save, I get all the appropriate things firing and fixes applied. Must look into that. Not an issue for now.
Screenshot 2024-12-02 at 11 41 15

packages/ui/src/views/repo/components/summary/summary.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <TableHead>Name</TableHead>
+            <TableHead>Last commit message</TableHead>
+            <TableHead className="text-right">Date</TableHead>
+          </TableRow>
+        </TableHeader>
+        <TableBody>
+          {files.map(file => (
+            <TableRow key={file.id} onClick={() => navigate(file.path)}>
+              <TableCell>
+                <ButtonGroup.Root
```
knagurski, line 96: Something we can look at later, but it looks like we're using a `ButtonGroup` as a layout component for non-buttons. Might be something we can swap to be just a `div` or something. Not a blocker, just something to think about.

packages/ui/src/views/layouts/SandboxLayout.tsx
```diff
@@ -79,56 +78,61 @@ function Main({
...
+  const paddingLeftStyle: CSSProperties = {
+    paddingLeft:
+      hasLeftPanel && hasLeftSubPanel
+        ? `calc(220px + ${leftSubPanelWidth}px)`
+        : hasLeftPanel
+          ? '220px'
+          : hasLeftSubPanel
+            ? `${leftSubPanelWidth}px`
+            : ''
+  }
```
knagurski, line 105: Just curious; is the move to `style` instead of a Tailwind class because the relevant utility class would be stripped from the output since it's determined at runtime and the Tailwind utility wouldn't find it in the source?

packages/ui/src/components/sha-badge.tsx
```diff
@@ -44,13 +40,15 @@ function Icon({ ...props }: IconProps) {
   const { children, handleClick } = props
 
   return (
-    <button
-      className="flex items-center border-l bg-background px-1.5 py-0.5 hover:bg-muted/50"
+    <Button
```
knagurski, line 43: One for another time, but for a11y reasons, we probably should only use a `button` if there is a click handler, otherwise assistive technologies will announce this as interactive, when it doesn't actually do anything. Perhaps we can double back later and change this to use something like a div instead of a button when `handleClick` isn't passed 🤔

packages/ui/src/components/search-files.tsx
```diff
@@ -12,88 +12,125 @@ import {
...
+
+const markedFileClassName = 'w-full text-foreground-8'
+
+/**
+ * Get marked file component with query
+ * @param file
+ * @param query
+ * @param matchIndex
+ */
+const getMarkedFileElement = (file: string, query: string, matchIndex: number) => {
```
knagurski, line 25: Not a blocker, but we should make sure we add return types in the future
```suggestion
const getMarkedFileElement = (file: string, query: string, matchIndex: number): ReactNode => {
```

Also, since this function does return JSX, it's worth considering making it a component.

packages/ui/src/views/layouts/SandboxLayout.tsx
3em: The issue is that Tailwind cannot generate custom classes during the build process because the value depends on props. That's why I'm using styles instead.

packages/ui/src/views/repo/components/summary/summary.tsx
3em: I have removed ButtonGroup.

packages/ui/src/views/layouts/SandboxLayout.tsx
knagurski: Yeah, that makes sense 👍🏻

packages/ui/src/components/sha-badge.tsx
3em: Added this behavior.

knagurski (approved):

PR #498, opened 2024-11-29
fix: repo description view and save

shaurya-harness:
1. added textarea to ui/components
2. description truncated for repo list
3. description view and edit/save for repo summary

Screenshot 2024-11-29 at 8 14 15 PM
Screenshot 2024-11-29 at 8 14 22 PM
Screenshot 2024-11-29 at 8 14 33 PM


packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
```
andrewgolovanov, line 1: Since we’ve migrated this component from `canary`, let’s try to remove it from `canary` right away by updating any imports that reference canary to the new path where the component now resides in `packages/ui`

shaurya-harness: sure

packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { cn } from '../utils/cn'
```
andrewgolovanov, line 3:
```suggestion
import { cn } from '@utils/cn'
```

packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
andrewgolovanov, line 5:
```suggestion
```
It seems we can get rid of this

packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,22 @@
...
+export interface TextareaProps extends React.TextareaHTMLAttributes<HTMLTextAreaElement> {}
```
andrewgolovanov, line 6:
```suggestion
export interface TextareaProps extends TextareaHTMLAttributes<HTMLTextAreaElement> {}
```

Let’s import these types directly from react

```tsx
import { TextareaHTMLAttributes } from 'react'
```

packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(({ className, ...props }, ref) => {
```
andrewgolovanov, line 8:
```suggestion
const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(({ className, ...props }, ref) => {
```

same

```tsx
import { forwardRef, TextareaHTMLAttributes } from 'react'
```

packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <textarea
+      className={cn(
+        'border-input placeholder:text-muted-foreground focus-visible:ring-ring flex min-h-[60px] w-full rounded-md border bg-transparent px-3 py-2 text-sm shadow-sm focus-visible:outline-none focus-visible:ring-1 disabled:cursor-not-allowed disabled:opacity-50',
```
andrewgolovanov, line 11: Let’s make sure to use the new color variables, which can be found in the [tailwind-config](https://github.com/harness/canary/blob/main/packages/ui/tailwind.ts#L90) and [styles](https://github.com/harness/canary/blob/main/packages/ui/src/styles.css#L99):

For borders: `borders`
For text: `foreground`
For background: `background`

Therefore, it should be:
```suggestion
        'border-borders-2 placeholder:text-foreground-5 flex min-h-[60px] w-full rounded border bg-transparent px-3 py-2 text-14 disabled:cursor-not-allowed disabled:opacity-50',
```


- `focus-visible` is currently set globally in the [styles](https://github.com/harness/canary/blob/main/packages/ui/src/styles.css#L1519), so there’s no need to apply it manually. This was done to ensure consistency across the entire project.
- `shadow-sm` – shadows are not used on input components in the updated design.
- `text-sm` => `text-14` – We use this approach to think in terms of `px` for easier understanding of text sizes. A value like sm doesn’t convey much, whereas 14 gives a clearer idea of the actual size. You can find more details about this in the [tailwind.config](https://github.com/harness/canary/blob/main/packages/ui/tailwind.ts#L248)

packages/ui/src/views/repo/repo-summary/components/summary-panel.tsx
```diff
@@ -40,15 +56,44 @@ const SummaryPanel: React.FC<SummaryPanelProps> = ({ title, details, timestamp,
...
       <Spacer size={2} />
-      {description?.length && <Text>{description}</Text>}
+      {description?.length && !isEditingDescription ? <Text className="line-clamp-3">{description}</Text> : <></>}
+      {isEditingDescription && (
+        <div>
+          <Textarea
+            defaultValue={description}
+            className="text-primary h-28"
+            value={newDesc}
+            onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => {
```
andrewgolovanov, line 74:
```suggestion
            onChange={(e: ChangeEvent<HTMLTextAreaElement>) => {
```
same
```tsx
import { useState, FC, ChangeEvent } from 'react'
```

packages/ui/src/components/textarea.tsx
andrewgolovanov: Once you update the paths, please don’t forget to remove this component from `canary`. In doing so, I believe there might be issues in `/packages/playground/`, so you may need to update the paths there as well for now. However, as far as I know, @knagurski  is already working on fully eliminating playground, so I think it will be easier for us to part with these components in the future. :slightly_smiling_face:

shaurya-harness: yeah, it might pollute this PR a bit though, will need to remove references from packages/playground and packages/views as well

packages/ui/src/components/textarea.tsx
shaurya-harness: no problem

packages/ui/src/components/textarea.tsx
```diff
@@ -0,0 +1,21 @@
+import { forwardRef, TextareaHTMLAttributes } from 'react'
+
+import { cn } from '@utils/cn'
+
+export interface TextareaProps extends TextareaHTMLAttributes<HTMLTextAreaElement> {}
```
knagurski, line 5: Not an issue, but if nothing is being added, then `type` is a better choice than `interface`
```suggestion
export type TextareaProps = TextareaHTMLAttributes<HTMLTextAreaElement>
```

knagurski (approved):

PR #507, opened 2024-12-02
Integrate 18n store

sans-harness:
- Adds a `useTranslationStore` that can now be passed down to components/views in `UI` and de-structured. 
-  Adds translations for `repo-nav`.

### DIFF
--- a/packages/ui/locales/es/views.json
+++ b/packages/ui/locales/es/views.json
@@ -1,5 +1,13 @@
 {
   "repos": {
+    "summary": "Summary",
+    "files": "Files",
+    "pipelines": "Pipelines",
+    "commits": "Commits",
+    "pull-requests": "Pull Requests",
+    "webhooks": "Webhooks",
+    "branches": "Branches",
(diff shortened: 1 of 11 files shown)


packages/ui/locales/es/views.json
```diff
@@ -1,5 +1,13 @@
 {
   "repos": {
+    "summary": "Summary",
+    "files": "Files",
+    "pipelines": "Pipelines",
+    "commits": "Commits",
+    "pull-requests": "Pull Requests",
+    "webhooks": "Webhooks",
+    "branches": "Branches",
+    "settings": "Settings",
```
abhinavrastogi-harness, line 10: these should be empty strings?

abhinavrastogi-harness (approved):
nice. looks like a good API to have

packages/ui/locales/es/views.json
sans-harness: I believe fallbacks get added  to all languages

PR #508, opened 2024-12-03
Handled loading, error and no data states in view

vivek-harness:
The loading and error states are currently handled directly in the repo-list page. Ideally, these states should be passed as props to the view, enabling the view to handle rendering for the Loading, Error, or No Data states in a more modular way.

In several pages/views, multiple ternary operators are being used, which can reduce code readability and maintainability. Let's try using if conditions instead for rendering different data states.

The source of truth for loading, error, and data states should come from the react-query hook and be passed as props from the page to the view.

Refer to the changes in this PR as a reference for effectively managing data states.

### DIFF
--- a/packages/ui/src/views/repo/repo-list/repo-list.tsx
+++ b/packages/ui/src/views/repo/repo-list/repo-list.tsx
@@ -41,69 +41,73 @@ export function RepoList({
   query,
   handleResetQuery
 }: PageProps) {
+  const noData = !(repos && repos.length > 0)
+
+  if (noData) {
+    return (
+      <StackedList.Root>
+        <div className="flex min-h-[50vh] items-center justify-center py-20">
+          {hasActiveFilters || query ? (
+            <NoData
+              iconName="no-search-magnifying-glass"
+              title="No search results"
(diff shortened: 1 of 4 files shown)


packages/ui/src/views/repo/repo-list/repo-list.tsx
```diff
@@ -41,69 +41,73 @@ export function RepoList({
...
+              title="No repositories yet"
+              description={[
+                'There are no repositories in this project yet.',
+                'Create new or import an existing repository.'
+              ]}
+              primaryButton={{
+                label: 'Create repository',
+                onClick: () => {
+                  /* TODO: add create handler */
+                }
```
sans-harness, line 76: NIT: lets not pass an empty handler?
```
onClick={noop}
```

sans-harness (approved):

packages/ui/src/views/repo/repo-list/repo-list.tsx
vivek-harness: There are a few scopes of improvement. Just made the changes related to data states for now.

PR #512, opened 2024-12-03
Repo Commits

vivek-harness:
Refactored code for removing dependencies from views and canary for the Repo Commits page.

Repo Commits v2:
Repo Commits V2

Repo Commits v1:
Repo Commits


apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { parseAsInteger, useQueryState } from 'nuqs'
+
+import { useFindRepositoryQuery, useListBranchesQuery, useListCommitsQuery } from '@harnessio/code-service-client'
+import { RepoCommitsView } from '@harnessio/ui/views'
+
+import { useGetRepoRef } from '../../framework/hooks/useGetRepoPath'
+import { PageResponseHeader } from '../../types'
+import { normalizeGitRef } from '../../utils/git-utils'
+
+const sortOptions = [{ name: 'Sort option 1' }, { name: 'Sort option 2' }, { name: 'Sort option 3' }]
```
abhinavrastogi-harness, line 12: let's not have hardcoded stuff. if api doesn't support sorting, let's just remove it.

apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -0,0 +1,63 @@
...
+      commitsList={commitData?.commits}
+      isFetchingBranches={isFetchingBranches}
+      isFetchingCommits={isFetchingCommits}
+      page={page}
+      selectBranch={selectBranch}
+      selectedBranch={selectedBranch}
+      setPage={(page: number) => setPage(page)}
+      xNextPage={xNextPage}
+      xPrevPage={xPrevPage}
+      sortOptions={sortOptions}
```
abhinavrastogi-harness, line 60: let's move this data into a store as discussed

packages/ui/src/components/node-group.tsx
```diff
@@ -0,0 +1,73 @@
+import { cn } from '@utils/cn'
```
abhinavrastogi-harness, line 1: this should probably be reviewed by the design team

packages/ui/src/views/repo/repo-commits/components/filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useCommonFilter } from '../hooks/useCommonFilter'
+
+interface FilterProps {
+  showSort?: boolean
+  sortOptions?: ListActions.DropdownItemProps[]
+  showSearch?: boolean
+}
+
+const Filter = <S,>({ showSort, sortOptions, showSearch = true }: FilterProps) => {
+  const { sort, query, handleSearch, handleDropdownChange } = useCommonFilter<S>()
```
abhinavrastogi-harness, line 15: we don't want to use `useCommonFilter` anymore. Let's remove it.
also, the sort/query etc should be maintained in gitness app via `nuqs`, and we just need to pass the props and callbacks to the view.

packages/ui/src/views/repo/repo-commits/components/filter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [value, setValue] = useState<string>()
+
+  useEffect(() => {
+    setValue(query || '')
+  }, [query])
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setValue(e?.target?.value)
+    handleSearch(e)
```
abhinavrastogi-harness, line 24: search inputs should be debounced

packages/ui/src/views/repo/repo-commits/components/pull-request-commits.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import { TypesCommit } from '../types'
+
+interface CommitProps {
+  data?: TypesCommit[]
+}
+
+export const PullRequestCommits = ({ ...props }: CommitProps) => {
+  const data = props.data
+  const commitsGroupedByDate: Record<string, TypesCommit[]> = useMemo(
```
abhinavrastogi-harness, line 14: this logic can be moved into the store

packages/ui/src/views/repo/repo-commits/components/pull-request-commits.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const totalNodes = entries.length
+
+  return (
+    <>
+      {entries.map(([date, commitData], node_idx) => (
+        <NodeGroup.Root key={date}>
+          <NodeGroup.Icon simpleNodeIcon />
+          <NodeGroup.Title>{date && <Text color="tertiaryBackground">Commits on {date}</Text>}</NodeGroup.Title>
+          <NodeGroup.Content>
+            {commitData && commitData.length > 0 && (
```
abhinavrastogi-harness, line 37: is it possible to have 0 commits at this point? if so, what does the UI look like in that scenario?

packages/ui/src/views/repo/repo-commits/components/pull-request-commits.tsx
```diff
@@ -0,0 +1,78 @@
...
+                      title={
+                        <div className="flex flex-col">
+                          <div className="max-w-[500px] truncate">{commit.title}</div>
+                          <div className="flex items-center pt-1">
+                            <div className="bg-tertiary-background h-5 w-5 rounded-full bg-cover">
+                              <Avatar className="h-5 w-5 rounded-full p-0">
+                                <AvatarFallback>
+                                  <Text size={1} color="tertiaryBackground">
+                                    {/* TODO: fix fallback string */}
+                                    {getInitials(commit.author?.identity?.name || 'Gitness')}
```
abhinavrastogi-harness, line 55: remove `Gitness`.

packages/ui/src/views/repo/repo-commits/components/repo-commits-branch-selector.tsx
```diff
@@ -0,0 +1,70 @@
...
+interface PageProps {
+  name: string
+  branchList?: BranchListProps[]
+  size?: 'default' | 'sm'
+  width?: 'auto' | 'sm' | 'md' | 'lg' | 'full'
+  selectBranch: (branch: string) => void
+  prefix?: string
+}
+
+const RepoCommitsBranchSelector = ({ ...props }: PageProps) => {
```
abhinavrastogi-harness, line 25: we already have a branch selector component on RepoSummary page. let's re-use that instead of adding a copy specific to repocommits page.

packages/ui/src/views/repo/repo-commits/hooks/useCommonFilter.tsx
abhinavrastogi-harness, line 1: this file should NOT be added. it's legacy code, and it's been replaced by `nuqs`

packages/ui/src/views/repo/repo-commits/components/pull-request-commits.tsx
abhinavrastogi-harness, line 1: this should be re-usable. let's call it `commits-list-view` or something.
it can be reused on repo commits and PR commits and branch commits

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -0,0 +1,105 @@
...
+                name: item.name || '',
+                sha: item.sha || ''
+              }))}
+              selectedBranch={selectedBranch}
+              tagList={props.tagList}
+              onSelectBranch={selectBranch}
+              repoId={props.repoId}
+              spaceId={props.spaceId}
+              useTranslationStore={props.useTranslationStore}
+            />
```
sans-harness, line 64: maybe de-structure all props?

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -0,0 +1,105 @@
...
+          )}
+
+          <Filter
+            showSearch={false}
+            sortOptions={props.sortOptions}
+            sort={props.sort}
+            setSort={props.setSort}
+            query={props.query}
+            setQuery={props.setQuery}
```
sans-harness, line 73: same ^

apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -0,0 +1,80 @@
...
+      sort={sort}
+      setSort={(sort: string) => setSort(sort)}
+      query={query}
+      setQuery={(query: string) => setQuery(query)}
+      repoId={repoId || ''}
+      spaceId={spaceId || ''}
+      tagList={[]}
+      useTranslationStore={useTranslationStore}
+    />
```
sans-harness, line 78: I am assuming this will be moved into a store as a follow up pr?

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
vivek-harness: We decided to do this on purpose. So that the dev knows they are coming from props and not a local variable. Some props that are reused may be de-structured.

apps/gitness/src/pages-v2/repo/repo-commits.tsx
vivek-harness: Right

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <NodeGroup.Content>
+            {commitData && commitData.length > 0 && (
+              <StackedList.Root>
+                {commitData.map((commit, repo_idx) => {
+                  const authorName = commit.author?.identity?.name
+
+                  return (
+                    <StackedList.Item
+                      className="hover:bg-transparent"
+                      key={commit.title}
```
abhinavrastogi-harness, line 45: using title as key is risky because two commits can have the same title. 
maybe concat with commit timestamp?

packages/ui/src/views/repo/repo-commits/components/filter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  showSort?: boolean
+  sortOptions?: ListActions.DropdownItemProps[]
+  showSearch?: boolean
+  sort: string | null
+  setSort: (sort: string) => void
+  query: string | null
+  setQuery: (query: string) => void
+}
+
+const Filter = ({ showSort, sortOptions, showSearch = true, sort, setSort, query, setQuery }: FilterProps) => {
```
abhinavrastogi-harness, line 16: do we need filters components in repo-commits? I don't think we actually support any of this on the commit list view?

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -0,0 +1,105 @@
...
+          />
+        </div>
+        <Spacer size={5} />
+
+        {isFetchingCommits ? (
```
abhinavrastogi-harness, line 78: don't do nested ternary operators. it negatively affects readability and can cause flickering if the order gets messed up.
instead do sequential ternary operators like
```
{ loading ? <Loader /> : null }
{ !loading && error ? <Error /> : null }
{ !loading && !error && data ? <Data /> : null }
```

abhinavrastogi-harness (approved):

packages/ui/src/views/repo/repo-commits/components/filter.tsx
vivek-harness: Removed it for now.

PR #513, opened 2024-12-04
feat: add Views Management System for Filters

andrewgolovanov:
**Description:**
This pull request implements a new Views system that allows users to save, manage, and reuse filter configurations. This feature enables users to persist their filter setups and switch between different saved views.

**Key Features**
- Save current filter configuration as a named view
- Manage saved views (rename, reorder, delete)
- Update existing views when filter configuration changes
- Automatic validation of view names (prevents duplicates)
- Drag-and-drop reordering of saved views
- Local storage persistence for saved views

**Technical Implementation**
- Integrated with existing filter system using `FiltersBar` component
- Added validation system for view names and filter configurations
- Implemented drag-and-drop functionality using dnd-kit
- Added local storage integration for persistence

https://github.com/user-attachments/assets/45cfe1c3-fb19-47b4-affe-28dffe303ea8

### DIFF
--- a/packages/ui/locales/en/component_old.json
+++ b/packages/ui/locales/en/component_old.json
@@ -1,5 +1,0 @@
-{
-  "filter": {
-    "save": "Save"
-  }
-}
--- a/packages/ui/locales/en/component_old.json
+++ b/packages/ui/locales/en/component_old.json
@@ -1,0 +1,5 @@
+{
+  "filter": {
+    "save": "Save"
+  }
+}
--- a/packages/ui/src/components/button.tsx
+++ b/packages/ui/src/components/button.tsx
@@ -25,8 +25,8 @@ const buttonVariants = cva(
         custom: ''
       },
       size: {
-        default: 'h-8 px-6 py-2',
-        sm: 'h-7 px-3 py-0 text-sm font-normal',
+        default: 'h-8 px-6',
+        sm: 'h-7 px-3 text-sm font-normal',
         xs: 'h-auto px-1.5 py-0.5 text-xs font-normal',
         md: 'h-9 px-7',
         lg: 'h-10 px-8',
--- a/packages/ui/src/components/filters/filters-bar/actions/views.tsx
+++ b/packages/ui/src/components/filters/filters-bar/actions/views.tsx
@@ -1,0 +1,272 @@
+import { FC, useEffect, useState } from 'react'
+
+import {
+  AlertDialog,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+  Button,
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+  Icon,
+  Input
+} from '@/components'
+
+import { FilterValue, SavedView, SortValue, ViewManagement } from '../../types'
+
+/**
+ * Views component handles the management of filter views including creation, updating, renaming and deletion.
+ * It provides UI for saving current filter/sort configurations as named views and managing existing ones.
+ */
+interface ViewsProps {
+  currentView: SavedView | null
+  savedViews: SavedView[]
+  viewManagement: ViewManagement & {
+    activeFilters: FilterValue[]
+    activeSorts: SortValue[]
+  }
+  hasChanges: boolean
+}
+
+const Views: FC<ViewsProps> = ({ currentView, viewManagement, hasChanges }) => {
+  /**
+   * Dialog states for creating new view and renaming existing one
+   */
+  const [isNewViewDialogOpen, setIsNewViewDialogOpen] = useState(false)
+  const [isRenameDialogOpen, setIsRenameDialogOpen] = useState(false)
+  const [newViewName, setNewViewName] = useState('')
+  const [nameError, setNameError] = useState('')
+
+  /**
+   * Effect handles real-time validation of view names
+   * - For new views: checks if name already exists
+   * - For renaming: checks if new name exists excluding current view
+   * - Skips validation for empty names (handled by button disable state)
+   */
+  useEffect(() => {
+    const trimmedName = newViewName.trim()
+
+    if (!trimmedName) {
+      setNameError('')
+      return
+    }
+
+    if (isRenameDialogOpen && currentView) {
+      if (trimmedName === currentView.name) {
+        setNameError('')
+        return
+      }
+
+      if (viewManagement.checkNameExists(trimmedName, currentView.id)) {
+        setNameError('A view with this name already exists')
+      } else {
+        setNameError('')
+      }
+    } else {
+      if (viewManagement.checkNameExists(trimmedName)) {
+        setNameError('A view with this name already exists')
+      } else {
+        setNameError('')
+      }
+    }
+  }, [newViewName, currentView, isRenameDialogOpen, viewManagement])
+
+  /**
+   * Resets error state when dialogs are opened/closed
+   * to ensure clean state for next interaction
+   */
+  useEffect(() => {
+    setNameError('')
+  }, [isNewViewDialogOpen, isRenameDialogOpen])
+
+  /**
+   * Handles saving a new view with current filters and sorts
+   * Validates name existence before saving
+   * Closes dialog and resets form state after successful save
+   */
+  const handleSaveNewView = () => {
+    const trimmedName = newViewName.trim()
+    if (viewManagement.checkNameExists(trimmedName)) return
+
+    viewManagement.saveView(trimmedName, viewManagement.activeFilters, viewManagement.activeSorts)
+    setNewViewName('')
+    setNameError('')
+    setIsNewViewDialogOpen(false)
+  }
+
+  /**
+   * Handles renaming an existing view
+   * Validates new name doesn't conflict with other views
+   * Updates view name and closes dialog on success
+   */
+  const handleRename = () => {
+    if (!currentView) return
+
+    const trimmedName = newViewName.trim()
+    if (viewManagement.checkNameExists(trimmedName, currentView.id)) return
+
+    viewManagement.renameView(currentView.id, trimmedName)
+    setNewViewName('')
+    setNameError('')
+    setIsRenameDialogOpen(false)
+  }
+
+  /**
+   * Updates current view with latest filter and sort configurations
+   * Used when user modifies filters/sorts and wants to update existing view
+   */
+  const handleUpdateCurrentView = () => {
+    if (!currentView) return
+
+    const updatedView = {
+      ...currentView,
+      filters: viewManagement.activeFilters,
+      sorts: viewManagement.activeSorts
+    }
+    viewManagement.updateView(updatedView)
+  }
+
+  /**
+   * Validation flags for enabling/disabling save buttons
+   * - canSaveNew: checks if new view name is valid and unique
+   * - canRename: checks if new name is valid, different from current, and unique
+   */
+  const canSaveNew = newViewName.trim() !== '' && !viewManagement.checkNameExists(newViewName.trim())
+  const canRename =
+    currentView &&
+    newViewName.trim() !== '' &&
+    newViewName.trim() !== currentView.name &&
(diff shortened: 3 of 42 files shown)


abhinavrastogi-harness (changes requested):
I think there's some mixup between two different features happening here.
In my understanding, the "View" dropdown was meant to switch between different layouts of the same content.. eg. list vs grid vs card layouts.
Vs. what you've implemented is the "Saved Filters" feature. This allows users to apply a few filters on the content, and save that for quick access later.

packages/ui/src/components/filters/filters-bar/actions/views.tsx
```diff
@@ -0,0 +1,241 @@
...
+    setIsNewViewDialogOpen(false)
+  }
+
+  /**
+   * Handles renaming existing view
+   * - Validates current view exists
+   * - Checks for duplicate names
+   * - Updates view name and closes dialog
+   */
+  const handleRename = () => {
```
abhinavrastogi-harness, line 97: same thing for all callbacks.. there should be no business logic in the `ui` package components and views. these react components should be presentation-only, and take callbacks as props for all behaviour

packages/ui/src/components/filters/filters-bar/actions/views.tsx
```diff
@@ -0,0 +1,241 @@
...
+        </DropdownMenu>
+      )}
+
+      <AlertDialog open={isNewViewDialogOpen} onOpenChange={setIsNewViewDialogOpen}>
+        <AlertDialogContent className="max-w-[460px]">
+          <AlertDialogHeader>
+            <AlertDialogTitle>New view</AlertDialogTitle>
+          </AlertDialogHeader>
+          <AlertDialogDescription variant="secondary">
+            You can save current configuration of the view. Manage and switch them through the{' '}
```
abhinavrastogi-harness, line 165: can we externalise all display strings to appropriate language files in the first cut itself? it'll reduce re-work to add i18n later.

packages/ui/src/components/filters/manage-views.tsx
```diff
@@ -0,0 +1,225 @@
...
+  AlertDialogContent,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+  Button,
+  Icon,
+  Input
+} from '@/components'
+import { DndContext } from '@dnd-kit/core'
+import { SortableContext, useSortable, verticalListSortingStrategy } from '@dnd-kit/sortable'
```
abhinavrastogi-harness, line 14: when did we add `dnd-kit` as a dependency? 
we need to evaluate all new libraries and get app-sec approvals before we start using them.

abhinavrastogi-harness: also, cc @srdjan-harness because we are using some other dnd library for the pipeline studio

packages/ui/src/components/filters/manage-views.tsx
```diff
@@ -0,0 +1,225 @@
...
+   * Validates the current state of views
+   * Checks for empty names and duplicates
+   */
+  const hasErrors = useCallback(() => {
+    const duplicatesMap = new Map<string, boolean>()
+
+    return localViews.some(view => {
+      const trimmedName = view.name.trim().toLowerCase()
+
+      // Проверка на пустое имя
```
abhinavrastogi-harness, line 142: i'd like to avoid non-english comments in the codebase. open to opinions though.

packages/ui/src/components/filters/manage-views.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+interface ManageViewsProps {
+  open: boolean
+  onOpenChange: (open: boolean) => void
+  views: SavedView[]
+  onSave: (views: SavedView[]) => void
+  onRename: (id: string, newName: string) => void
+}
+
+const ManageViews: FC<ManageViewsProps> = memo(({ open, onOpenChange, views, onSave }) => {
```
abhinavrastogi-harness, line 100: this contains a LOT of business logic, which we decided to avoid in `ui` package.

we want to follow the MVVM pattern, where the "view" layer, should only be the presentational layer.

we want to be able to re-use the views across two different backends, which is not possible if we hard-code business logic in the view layer.

packages/ui/src/views/webhooks/webhook-list/repo-webhook-list-page.tsx
```diff
@@ -1,208 +1,48 @@
...
 import useFilters from '@components/filters/use-filters'
 import { Button, ListActions, PaginationComponent, SearchBox, SkeletonList, Spacer, Text } from '@components/index'
 import { useCommonFilter } from '@hooks/useCommonFilter'
+import {
+  getFilterOptions,
+  getLayoutOptions,
+  getSortDirections,
+  getSortOptions
+} from '@views/repo/repo-list/filter-options'
 import { formatDistanceToNow } from 'date-fns'
```
abhinavrastogi-harness, line 14: same issue with `date-fns` package. this should not be added and neither should it be required.
I know this is existing code, but just making a note here for later.

packages/ui/src/views/webhooks/webhook-list/repo-webhook-list-page.tsx
```diff
@@ -1,208 +1,48 @@
...
-
 const LinkComponent = ({ to, children }: { to: string; children: React.ReactNode }) => <Link to={to}>{children}</Link>
 
 const RepoWebhookListPage: React.FC<WebhookListProps> = ({ useWebhookStore, useTranslationStore }) => {
   const { t } = useTranslationStore()
 
+  const FILTER_OPTIONS = getFilterOptions(t)
+  const SORT_OPTIONS = getSortOptions(t)
+  const SORT_DIRECTIONS = getSortDirections(t)
+  const LAYOUT_OPTIONS = getLayoutOptions(t)
```
abhinavrastogi-harness, line 28: I don't think we need to pass `t` function down to all components. We can just call the `useTranslationStore` hook everywhere directly.

abhinavrastogi-harness (changes requested):
In general I'm seeing a lot of mixed business logic in the shared views in this PR. Maybe let's discuss tomorrow before we merge this change.

packages/ui/src/components/button.tsx
```diff
@@ -25,8 +25,8 @@ const buttonVariants = cva(
         custom: ''
       },
       size: {
-        default: 'h-8 px-6 py-2',
-        sm: 'h-7 px-3 py-0 text-sm font-normal',
+        default: 'h-8 px-6',
+        sm: 'h-7 px-3 text-sm font-normal',
```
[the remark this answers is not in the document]
andrewgolovanov: @abhinavrastogi-harness I only removed the vertical padding since it doesn’t make sense, as the button’s height is already defined using `h-`
image

packages/ui/locales/en/component_old.json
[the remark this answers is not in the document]
andrewgolovanov: Unfortunately, it appeared on its own after the project was built.

packages/ui/src/components/filters/filters-bar/actions/views.tsx
andrewgolovanov: @abhinavrastogi-harness I’ve revised the approaches, so now only the local logic necessary for this component remains. It handles its own state, while all the business logic for filtering and validation has been moved to a hook called [use-view-management](https://github.com/harness/canary/pull/513/files#diff-c58401d04ae9a913353d34aeabc42330af54d16582e87e43325c039034e9c35bR43), located at `views/repo`.

packages/ui/src/components/filters/manage-views.tsx
andrewgolovanov: At the moment, I’ve left comments in the code to revisit this in a future PR.

packages/ui/src/components/filters/manage-views.tsx
andrewgolovanov: Here, all the business logic has also been moved to the [use-view-management](https://github.com/harness/canary/pull/513/files#diff-c58401d04ae9a913353d34aeabc42330af54d16582e87e43325c039034e9c35bR43) hook at the page level, which is used for validating and managing all the local states of this component.

3em (approved):

packages/ui/src/components/filters/filters-bar/actions/views.tsx
andrewgolovanov: I’m planning to address this in the next PR so that this one doesn’t block further development.

packages/ui/src/views/webhooks/webhook-list/repo-webhook-list-page.tsx
andrewgolovanov: This will be fixed in the next PR

packages/ui/src/views/webhooks/webhook-list/repo-webhook-list-page.tsx
andrewgolovanov: I’ve left a comment in the code to revisit the approach to using `date-fns` in the future and to plan for fully phasing out its usage.

packages/ui/locales/en/component_old.json
```diff
@@ -0,0 +1,5 @@
+{
```
abhinavrastogi-harness, line 1: can we delete these `component_old` files? these should NOT be needed.

packages/ui/src/components/filters/filters-bar/actions/views.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const [newViewName, setNewViewName] = useState('')
+  const [nameError, setNameError] = useState('')
+
+  /**
+   * Effect handles real-time validation of view names
+   * - For new views: checks if name already exists
+   * - For renaming: checks if new name exists excluding current view
+   * - Skips validation for empty names (handled by button disable state)
+   */
+  useEffect(() => {
```
abhinavrastogi-harness, line 50: this validation should not exist in reusable components

packages/ui/src/components/filters/filters-bar/actions/views.tsx
```diff
@@ -0,0 +1,272 @@
...
+  }, [isNewViewDialogOpen, isRenameDialogOpen])
+
+  /**
+   * Handles saving a new view with current filters and sorts
+   * Validates name existence before saving
+   * Closes dialog and resets form state after successful save
+   */
+  const handleSaveNewView = () => {
+    const trimmedName = newViewName.trim()
+    if (viewManagement.checkNameExists(trimmedName)) return
```
abhinavrastogi-harness, line 93: this validation should not exist in reusable components

packages/ui/src/views/repo/hooks/use-view-management.ts
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * Hook for managing filter views state and operations
+ * Handles saving, loading, updating and validating filter views
+ *
+ * @param storageKey - Unique key for localStorage
+ * @param setActiveFilters - Callback to update active filters
+ * @param setActiveSorts - Callback to update active sorts
+ */
+export const useViewManagement = ({
```
abhinavrastogi-harness, line 43: this hook implements logic that seems specific to harness open-source, and should be implemented in the gitness app instead.

abhinavrastogi-harness: this should also be implemented using zustand stores once it's in gitness

packages/ui/src/views/repo/hooks/use-view-management.ts
abhinavrastogi-harness, line 1: is this system multi-tenant? I mean, will this scale to manage views for multiple pages and resource types? eg. PR list, repo list, commit list, pipeline list etc.

abhinavrastogi-harness (approved):
unblocking this PR to merge, but I think we need to revisit filters later, at which point there will need to be a major refactor as we move all filtering logic to the backend.

packages/ui/src/views/repo/hooks/use-view-management.ts
andrewgolovanov: @abhinavrastogi-harness Yes, the current system is easily scalable and allows for unique filtering to be applied to each page.

PR #515, opened 2024-12-04
Add i18n to badge and pagination for repo-list

sans-harness:


packages/ui/src/components/pagination-component.tsx
```diff
@@ -19,6 +20,7 @@ interface PaginationComponentProps {
   totalPages?: number
   nextPage?: number
   previousPage?: number
+  t?: TFunction
```
andrewgolovanov, line 23: It seems like this should be required; otherwise we'll run into an error because we don't have a fallback in case this parameter isn't passed to `PaginationPrevious`.
https://github.com/harness/canary/pull/515/files#diff-4c674009901fd309153a188c5adbf7b789c559c802d290422cea1309e7ce6d9dR73

```tsx
<span>{t!('component:pagination.previous', 'Previous')}</span>
```

andrewgolovanov (approved):
@sans-harness lgtm! :rocket:

PR #519, opened 2024-12-05
Branch selector store

vivek-harness:
Added a zustand store for branch selector and used it across components.

Modifications had to be made in Repo Commits, Repo Summary and Repo Sidebar pages for using the new store. A review is needed by the author of those pages.

https://github.com/user-attachments/assets/fc1a05c6-d65c-420a-b140-f69c315926bf


apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -24,7 +25,22 @@ export default function RepoCommitsPage() {
...
+    useBranchSelectorStore()
+
+  useEffect(() => {
+    branches &&
+      !branchList &&
+      setBranchList(
+        branches.map(item => ({
+          name: item.name || '',
+          sha: item.sha || ''
+        }))
```
sans-harness, line 38: should this not have `divergence` included?

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -80,27 +89,33 @@ export const RepoSidebar = () => {
...
+  //   (branchName: BranchSelectorListItem) => {
+  //     const branch = branchList.find(b => b.name === branchName.name)
+  //     if (branch) {
+  //       setSelectedBranch(branch)
+  //       navigate(`/${spaceId}/repos/${repoId}/code/${branch.name}`)
+  //     }
+  //   },
+  //   [navigate, repoId, spaceId, branchList]
+  // )
+
```
sans-harness, line 113: remove if unnecessary

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -139,12 +154,13 @@ export const RepoSidebar = () => {
...
-        selectedBranch={selectedBranch}
-        selectBranch={selectBranch}
-        branchList={branchList}
-        tagList={[]}
+        // repoId={repoId}
+        // spaceId={spaceId}
+        // selectedBranch={selectedBranch}
+        // selectBranch={selectBranch}
+        // branchList={branchList}
+        // tagList={[]}
```
sans-harness, line 162: same^

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  spaceId: '',
+  setSpaceId: (spaceId: string) => set({ spaceId }),
+
+  prefix: '',
+  setPrefix: (prefix: string | undefined) => set({ prefix }),
+
+  className: '',
+  setClassName: (className: string | undefined) => set({ className })
+}))
```
sans-harness, line 29: why does `className` need to be a part of our store?

sans-harness (commented):
Lets remove all unused comments

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
vivek-harness: Right, shouldn't be there. Let me pass it as a prop instead.

apps/gitness/src/pages-v2/repo/repo-commits.tsx
vivek-harness: Branch Selector isn't using it. We may add more props as and when needed, right?

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
sans-harness: Im confused why we need to pass a `className` prop at all?

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
sans-harness, line 1: NIT: lets rename this to `repo-branch-list-store` ? Since its a common store to list branches across pages not just for the branch selector.

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  repoId: '',
+  setRepoId: (repoId: string) => set({ repoId }),
+
+  spaceId: '',
+  setSpaceId: (spaceId: string) => set({ spaceId }),
+
+  prefix: '',
+  setPrefix: (prefix: string | undefined) => set({ prefix })
+}))
```
sans-harness, line 26:
```
prefix?: string
```

packages/ui/src/views/repo/components/branch-selector/branch-selector-dropdown.tsx
```diff
@@ -33,7 +33,7 @@ export const getBranchSelectorLabels = (t: TFunction) => ({
 })
 
 export interface BranchSelectorDropdownProps {
-  selectedBranch: BranchSelectorListItem
+  selectedBranch?: BranchSelectorListItem
```
sans-harness, line 36: does this need to be optional?

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
vivek-harness: The button in Branch Selector is appending the className from props. I don't think the button will be styled differently specifically in a particular page, so we may remove it. Anyway, no page is passing className. It's an optional prop.

packages/ui/src/views/repo/components/branch-selector/branch-selector-dropdown.tsx
vivek-harness: Yes, if there's no selected branch, only then we update it in another page or else we retain the state.

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
vivek-harness: It's an optional prop, we can skip the default value altogether, or else default it as an empty string.

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
vivek-harness: Removed

apps/gitness/src/pages-v2/repo/stores/branch-selector-store.ts
sans-harness: Imho we should make props optional using `?` instead of explicitly adding `undefined` as a type. Just how its done across the code base so helps maintain consistency, and is more readable.  Sorry for nitpicking 😅

vivek-harness: Makes sense! I too was wondering why I used undefined 😂

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -163,7 +171,7 @@ export default function RepoSummaryPage() {
     }
 
     pathDetails({
-      queryParams: { git_ref: normalizeGitRef(selectedBranch.name) },
+      queryParams: { git_ref: normalizeGitRef(selectedBranch?.name || '') },
```
praneshg239, line 174: at line 167 we are making the loading as true and we are checking for the condition `repoEntryPathToFileTypeMap.size`. We are setting the loader to false at line 196. If the condition not satisfies, we are returning from the function. Still loader is true and it will keep showing skeleton UI. 

It would be good if we place the entire effect within the condition.

```tsx
if(repoEntryPathToFileTypeMap.size){
  setLoading(true)

  ...

  .finally(() => {
      setLoading(false)
   })


}
```

apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -24,7 +25,22 @@ export default function RepoCommitsPage() {
...
   })
 
-  const [selectedBranch, setSelectedBranch] = useState<BranchSelectorListItem>({ name: '', sha: '' })
+  const { branchList, setBranchList, selectedBranch, setSelectedBranch, setRepoId, setSpaceId } =
+    useBranchSelectorStore()
+
+  useEffect(() => {
+    branches &&
+      !branchList &&
+      setBranchList(
```
abhinavrastogi-harness, line 34: we should not have this condition. it will prevent updating the branchlist once it has been set.
instead, the `setBranchList` function should take care of merging the old and new data while de-duplicating it.

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -65,12 +73,12 @@ export const RepoSidebar = () => {
     if (!fullGitRef) {
       const defaultBranch = branchList.find(branch => branch.default)
       if (defaultBranch) {
-        setSelectedBranch(defaultBranch)
+        !selectedBranch && setSelectedBranch(defaultBranch)
```
abhinavrastogi-harness, line 76: same thing here. set should not be conditional

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -73,14 +81,14 @@ export default function RepoSummaryPage() {
...
-  const branchList: BranchSelectorListItem[] = useMemo(() => {
-    if (!branches) return []
-
-    return branches.map(item => ({
-      name: item?.name || '',
-      sha: item?.sha || '',
-      default: item?.name === repository?.default_branch
-    }))
+  useEffect(() => {
+    setBranchList(
```
abhinavrastogi-harness, line 85: this can be moved to the `onSuccess` callback of the query on line 41

apps/gitness/src/pages-v2/repo/stores/repo-branch-selector-store.ts
```diff
@@ -0,0 +1,24 @@
...
+import { BranchSelectorListItem, IBranchSelectorStore } from '@harnessio/ui/views'
+
+export const useBranchSelectorStore = create<IBranchSelectorStore>(set => ({
+  size: 'default',
+  setSize: size => set({ size }),
+
+  selectedBranch: undefined,
+  setSelectedBranch: (selectedBranch?: BranchSelectorListItem) => set({ selectedBranch }),
+
+  setBranchList: (branchList?: BranchSelectorListItem[]) => set({ branchList }),
```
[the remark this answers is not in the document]
vivek-harness: By default, it shall take the value as undefined, but ya let's add it for better readability.

apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -24,7 +25,22 @@ export default function RepoCommitsPage() {
     queryParams: { page }
   })
 
-  const [selectedBranch, setSelectedBranch] = useState<BranchSelectorListItem>({ name: '', sha: '' })
+  const { branchList, setBranchList, selectedBranch, setSelectedBranch, setRepoId, setSpaceId } =
+    useBranchSelectorStore()
+
+  useEffect(() => {
+    branches &&
```
praneshg239, line 32: shall we make it as below?

```ts
if(!branchList && Array.isArray(branches)){
  branches.map(item => ({
     name: item.name || '',
     sha: item.sha || ''
  }))
}
```

praneshg239:
In the attached screencast, upon selecting the branch, entire screen shifts little right and all the elements are flickering. Can we check that too?

packages/ui/src/views/repo/components/branch-selector/branch-selector.tsx
```diff
@@ -1,61 +1,58 @@
...
-      <BranchSelectorDropdown
-        branchList={branchList}
-        tagList={tagList}
-        selectedBranch={selectedBranch}
-        onSelectBranch={onSelectBranch}
-        repoId={repoId}
-        spaceId={spaceId}
-        useTranslationStore={useTranslationStore}
-      />
+      {branchList && (
```
andrewgolovanov, line 45:
```suggestion
      {branchList && (
```

I don’t think this check is necessary, as we’ve already handled it internally. If, for some reason, the `branchList` is empty, we’ll display an appropriate message to indicate that.

Google Chrome 2024-12-05 16 54 04

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -3,35 +3,27 @@ import { PaginationComponent } from '@components/pagination-component'
...
   isFetchingCommits: boolean
-  commitsList: TypesCommit[] | null | undefined
   isFetchingBranches: boolean
-  branches?: {
-    name?: string
-    sha?: string
-  }[]
-  selectedBranch: BranchSelectorListItem
-  selectBranch: (branch: BranchSelectorListItem) => void
+  commitsList: TypesCommit[] | null | undefined
```
cjlee01, line 15:
```suggestion
  commitsList?: TypesCommit[] | null 
```

apps/gitness/src/pages-v2/pull-request/pull-request-compare.tsx
```diff
@@ -313,12 +313,23 @@ export const CreatePullRequest = () => {
     [branchList, tagsList]
   )
 
+  const { setTagList, setBranchList, setRepoId, setSpaceId } = useRepoBranchesStore()
+
+  useEffect(() => {
+    setTagList(tagsList)
+    setBranchList(branchList)
```
abhinavrastogi-harness, line 320: Can these be moved to the onsuccess callback of the query instead of a useEffect?

vivek-harness: Ohh yes, I actually checked that since you mentioned it in the previous comment, but it seems the onSuccess callback is deprecated due to performance overheads and will be removed in the next major version of react-query. We'll need to rely on the data object as we are doing now.

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -35,17 +36,27 @@ export default function RepoSummaryPage() {
   const { spaceId, repoId } = useParams<PathParams>()
   const [gitRef, setGitRef] = useState<string>('')
 
+  const {
+    branchList,
+    setBranchList,
+    setSpaceId,
+    setRepoId,
```
abhinavrastogi-harness, line 43: Hmm. Why are these separate functions?
Since a branch would depend on space and repo, I think it should be one fn

apps/gitness/src/pages-v2/pull-request/pull-request-compare.tsx
sans-harness: good point. Interesting read: https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose

apps/gitness/src/pages-v2/repo/repo-summary.tsx
vivek-harness: done

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
vivek-harness: done

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
```diff
@@ -50,8 +46,6 @@ export interface RepoSummaryViewProps {
     tag_count?: number
     pull_req_summary?: { open_count: number } | undefined
```
cjlee01, line 47: do we still need undefined here?

cjlee01 (approved):

PR #525, opened 2024-12-05
Fixed Summary Page Flicker

vivek-harness:
Handled Loading and No Data states correctly and passed the loading states as needed to the view layer.

After Fix:

https://github.com/user-attachments/assets/202cd649-9202-4a03-9559-0f5d49d26517

Before Fix:

https://github.com/user-attachments/assets/e50a3aa6-329f-4d22-89b9-9160a013700e

### DIFF
--- a/apps/gitness/src/pages-v2/repo/repo-summary.tsx
+++ b/apps/gitness/src/pages-v2/repo/repo-summary.tsx
@@ -35,9 +35,13 @@ export default function RepoSummaryPage() {
   const { spaceId, repoId } = useParams<PathParams>()
   const [gitRef, setGitRef] = useState<string>('')
 
-  const { data: { body: repository } = {}, refetch: refetchRepo } = useFindRepositoryQuery({ repo_ref: repoRef })
+  const {
+    data: { body: repository } = {},
+    refetch: refetchRepo,
+    isLoading: isLoadingRepository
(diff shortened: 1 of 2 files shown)


apps/gitness/src/pages-v2/repo/repo-summary.tsx
sans-harness, line 1: I don't think this is the right way to fix this. I don't think we need to show the entire page as loading if branches/tags/paths etc has not been loaded. They should ideally have different loading states?

sans-harness (changes requested):

apps/gitness/src/pages-v2/repo/repo-summary.tsx
vivek-harness: Ok, then just these two loadng states are needed for now: {loading || isLoadingRepoDetails}

vivek-harness: Those are the initial loading states. isLoadingRepoDetails is the one because of which we are seeing the flicker.

abhinavrastogi-harness: agreed, let;s only block on the absolutely essential data required to render the main content of the page
for eg. we dont need to block on the readme file

abhinavrastogi-harness (approved):

sans-harness (approved):

PR #534, opened 2024-12-06
feat: [PIPE-23928]: Persist pinned and recent links in localStorage

praneshg239:
### Technical chart:
image

### Video demo:
https://harness.slack.com/archives/C07QA52CWKH/p1733490169130969

### DIFF
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -146,7 +146,7 @@ export default function App() {
       path: '/',
       element: <RootWrapper />,
       children: [
-        // 🚨 Revert the temporary links before merging 🚨
+        // 🚨 Routes to fix 🚨
         {
           path: 'repos',
           element: (
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -626,10 +626,6 @@ export default function App() {
       path: 'cloud-costs',
       element: <EmptyPage pathName="Cloud Cost Management" />
     },
-    // {
-    //   path: 'databases',
-    //   element: <h1>DBs</h1>
-    // },
     {
       path: 'incidents',
       element: <EmptyPage pathName="Incidents" />
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -146,6 +146,7 @@ export default function App() {
       path: '/',
       element: <RootWrapper />,
       children: [
+        // 🚨 Revert the temporary links before merging 🚨
         {
           path: 'repos',
           element: (
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -10,7 +10,8 @@ import {
   EmptyPage,
   ForgotPasswordPage as ForgotPasswordPageV2,
   NewPasswordPage as NewPasswordPageV2,
-  OTPPage as OTPPageV2
+  OTPPage as OTPPageV2,
+  SandboxLayout
 } from '@harnessio/ui/views'
 import {
   ForgotPasswordPage,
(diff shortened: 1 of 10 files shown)


apps/gitness/src/App.tsx
```diff
@@ -145,6 +146,39 @@ export default function App() {
       path: '/',
       element: <RootWrapper />,
       children: [
+        // 🚨 Routes to fix 🚨
+        {
```
praneshg239, line 150: These links are added for testing and as discussed, these links have to be updated with proper screens.

apps/gitness/src/components/RootWrapper.tsx
```diff
@@ -1,40 +1,29 @@
...
-      />
-    </>
+    <SandboxRoot
+      t={t}
+      currentUser={currentUser}
+      pinnedMenu={pinned}
+      recentMenu={recent}
+      setPinned={setPinned}
+      setRecent={setRecent}
+      setNavLinks={setNavLinks}
```
abhinavrastogi-harness, line 24: can we pass the `useNavStore` directly to the view?

apps/gitness/src/components/stores/recent-pinned-nav-links.store.ts
```diff
@@ -0,0 +1,78 @@
...
+      let copyRecent = [...state.recent]
+
+      if (remove) {
+        copyRecent = copyRecent.filter(item => item.id !== route.id)
+      } else {
+        // Check if the route already exists in the history
+        if (state.recent.find(item => item.id === route.id)) return state
+
+        // Check if the route already exists in the pinned list
+        if (state.pinned.find(item => item.id === route.id)) return state
```
abhinavrastogi-harness, line 29: could we use a Set to avoid iterating an array for finding duplicates?
not a major issue, since it'll be a small array anyway.. but could grow if we're adding all pages in the future

packages/ui/src/components/manage-navigation/index.tsx
```diff
@@ -151,14 +151,16 @@ export const ManageNavigation = ({
...
                             </Button>
-                            <Button
-                              className="absolute right-1 top-0.5 z-20 text-icons-4 hover:text-icons-2"
-                              size="sm_icon"
-                              variant="custom"
-                              onClick={() => removeFromPinnedItems(item)}
-                            >
-                              <Icon className="w-3.5" name="x-mark" size={14} />
-                            </Button>
+                            {!item.permanentlyPinned ? (
```
abhinavrastogi-harness, line 154: i think we should not allow re-ordering these permanent items either.. maybe hide the 3-dots entirely

packages/ui/src/views/layouts/SandboxRoot.tsx
```diff
@@ -58,58 +34,42 @@ export const SandboxRoot = ({
...
-        changeRecentMenu([...new Set([nextRouteData.id, ...recentMenu])])
-      }
-    },
-    [changeRecentMenu, pinnedMenu, recentMenu]
-  )
+  useLocationChange({ t, onRouteChange: setRecent })
 
-  useLocationChange({ t, onRouteChange: handleChangeRecentMenu })
+  useEffect(() => {
+    const linksFromStorage = localStorage.getItem('nav-items')
```
abhinavrastogi-harness, line 53: if we are using a zustand store for saving pinned items, adding a `persist` middle should give us this functionality

abhinavrastogi-harness (approved):

packages/ui/src/components/manage-navigation/index.tsx
praneshg239: Sure

packages/ui/src/views/layouts/SandboxRoot.tsx
praneshg239: This is to set the very first time pinned column links for the user. As the pinned column data is in `SandboxRoot` component, we cannot set the zustand state directly in `RootWrapper`. So setting the initial pinned links this way.

apps/gitness/src/components/stores/recent-pinned-nav-links.store.ts
praneshg239: sure

PR #537, opened 2024-12-06
Fixes flicker issue on repo-list page on clearing search

sans-harness:
https://github.com/user-attachments/assets/af20925c-0501-492f-9914-302ae466c38d

### DIFF
--- a/apps/gitness/src/pages-v2/repo/repo-list.tsx
+++ b/apps/gitness/src/pages-v2/repo/repo-list.tsx
@@ -20,7 +20,7 @@ export default function ReposListPage() {
   const [queryPage, setQueryPage] = useQueryState('page', parseAsInteger.withDefault(1))
 
   const {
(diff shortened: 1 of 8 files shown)


apps/gitness/src/pages-v2/repo/repo-list.tsx
```diff
@@ -17,28 +16,27 @@ export default function ReposListPage() {
...
     },
-    { retry: false }
-  )
-
-  useEffect(() => {
-    if (repoData) {
-      setRepositories(repoData, headers)
+    {
+      retry: false,
+      onSuccess: data => {
```
abhinavrastogi-harness, line 35: do we need this change? since `onSuccess` is deprecated, let's see if it still works the same with a useEffect

abhinavrastogi-harness (approved):

abhinavrastogi-harness (approved):

PR #543, opened 2024-12-09
fix: branch tag selection and store updates from gitness

shaurya-harness:
fixes branch tag selection and store updates from gitness
added tagsList for the branch-selector in repo-commits view too


apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -23,24 +23,33 @@ import { useRepoBranchesStore } from '././stores/repo-branches-store.ts'
...
+  } = useRepoBranchesStore()
 
   const repoRef = useGetRepoRef()
   const { spaceId, repoId } = useParams<PathParams>()
   const { fullGitRef, gitRefName, fullResourcePath } = useCodePathDetails()
   const navigate = useNavigate()
 
   useEffect(() => {
-    !selectedBranchTag && setSelectedBranchTag({ name: fullGitRef || '', sha: '' })
+    !selectedBranchTag && setSelectedBranchTag({ name: gitRefName || '', sha: '' })
```
vivek-harness, line 43: Do we really need !selectedBranchTag condition? Or can we directly set the state?

vivek-harness: If we need the condition, we should probably add it in the setter method in store.

vivek-harness (approved):

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
shaurya-harness: yeah it can be directly set, fullGitRef from route would be the source of truth

PR #558, opened 2024-12-10
Add eslint rules for i18next and prevent adding string literals in JSX

praneshg239:
Adding eslint rules for i18next support. It prevents adding strings as JSX children.

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,8 @@
     "plugin:import/typescript",
     "plugin:@typescript-eslint/recommended",
 
+    "plugin:i18next/recommended",
+
     "plugin:react/recommended",
     "plugin:react-hooks/recommended",
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -49,6 +49,12 @@
     "jsx-a11y/click-events-have-key-events": "warn",
     "jsx-a11y/no-autofocus": "warn",
 
+    /*
+     * It is a temporary change.
+     * It will be updated as "error" after fixing all the components.
+     */
+    "i18next/no-literal-string": "warn",
+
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -12,6 +12,8 @@
     "plugin:import/typescript",
     "plugin:@typescript-eslint/recommended",
 
+    "plugin:i18next/recommended",
```
shaurya-harness, line 15: test

knagurski:
I think we'll need to set this as `warn` initially, otherwise we'll need to pull the brakes on everything until resolved. We'll also need to exclude `apps/design-system` from using these rules since it doesn't (yet) use i18n.

praneshg239:
> I think we'll need to set this as `warn` initially, otherwise we'll need to pull the brakes on everything until resolved. We'll also need to exclude `apps/design-system` from using these rules since it doesn't (yet) use i18n.

sure @knagurski. But it would be good if we keep a deadline for making all the changes to update it as `error` again. Any thoughts on this?

knagurski:
> > I think we'll need to set this as `warn` initially, otherwise we'll need to pull the brakes on everything until resolved. We'll also need to exclude `apps/design-system` from using these rules since it doesn't (yet) use i18n.
> 
> sure @knagurski. But it would be good if we keep a deadline for making all the changes to update it as `error` again. Any thoughts on this?

Yeah, totally agree. We should have a work stream to extract them ASAP and guard against new strings going in until such time that we can make it an error. Earlier we make it just part of the expected standard, the less work we'll have overall.

pnpm-lock.yaml
abhinavrastogi-harness, line 1: something is wrong here, if adding just one package is causing 8500 lines of change in the lockfile

PR #559, opened 2024-12-10
Fix UI package scope

knagurski:
This PR removes unnecessary CSS build tasks, switches to use the built CSS from the ui package and adds missing `files` entry in ui `package.json`

### DIFF
--- a/packages/ui/vite.config.ts
+++ b/packages/ui/vite.config.ts
@@ -30,7 +30,16 @@ export default defineConfig({
       formats: ['es']
     },
     rollupOptions: {
-      external: external
+      external: external,
+      output: {
+        assetFileNames: chunkInfo => {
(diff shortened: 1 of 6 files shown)


packages/ui/vite.config.ts
```diff
@@ -30,7 +30,16 @@ export default defineConfig({
       formats: ['es']
     },
     rollupOptions: {
-      external: external
+      external: external,
+      output: {
+        assetFileNames: chunkInfo => {
+          if (chunkInfo.name === 'style.css') {
+            return 'styles.css'
```
abhinavrastogi-harness, line 37: why do we need to rename the output? cant we just use `style.css` when importing everywhere?

abhinavrastogi-harness (approved):

PR #560, opened 2024-12-10
MFE Gitness

vivek-harness:
https://github.com/user-attachments/assets/bbf93eba-8ed4-4a4d-9818-3c966a3ab643


packages/ui/src/views/layouts/Floating1ColumnLayout.tsx
```diff
@@ -101,7 +101,7 @@ const HighlightedFloatingLayout = ({ children, className, theme = 'blue' }: High
         />
       </div>
       <span
-        className="pointer-events-none absolute inset-0 bg-[url('/images/signin/noise.png')] bg-[size:100px_100px] bg-repeat opacity-70 mix-blend-overlay"
```
praneshg239, line 104: check with Calvin

packages/views/src/components/pull-request/pull-request-diff-viewer.tsx
```diff
@@ -284,7 +284,7 @@ const PullRequestDiffViewer = ({
                   </Text>
                 </div>
                 <div className="flex items-center gap-3 border-t p-4">
-                  <div className='size-6 rounded-full bg-tertiary-background bg-[url("../images/user-avatar.svg")] bg-cover'></div>
```
praneshg239, line 287: check with Calvin

packages/ui/src/styles.css
```diff
@@ -261,6 +261,225 @@
     --canary-box-shadow-pagination: var(--canary-harness-grey-400) / 0.1;
   }
 
+  /* Need .light and .dark classes for MFE */
+  .light {
```
abhinavrastogi-harness, line 265: let's combine this with `.light-std-std, .light`

apps/gitness/src/AppMFE.css
```diff
@@ -0,0 +1 @@
+@import '@harnessio/ui/styles.css';
```
abhinavrastogi-harness, line 1: do we need this file? I see that `AppMFE.tsx` is directly importing `@harnessio/ui/styles.css`

apps/gitness/src/AppMFE.tsx
```diff
@@ -0,0 +1,335 @@
...
+import { useEffect } from 'react'
+import { I18nextProvider } from 'react-i18next'
+import { createBrowserRouter, Outlet, RouterProvider, useLocation, useNavigate } from 'react-router-dom'
+import ReactShadowRoot from 'react-shadow-root'
+
+import { QueryClientProvider } from '@tanstack/react-query'
+import { NuqsAdapter } from 'nuqs/adapters/react-router'
+
+import styles from '!!raw-loader!@harnessio/ui/styles.css'
+import { TooltipProvider } from '@harnessio/canary'
```
abhinavrastogi-harness, line 12: we should not be using anything from `@harnessio/canary`. only the `ui` package should be used. we plan to remove the `canary` package

apps/gitness/src/AppMFE.tsx
```diff
@@ -0,0 +1,335 @@
...
+  const { setTheme } = useThemeStore()
+  useEffect(() => {
+    if (theme === 'Light') {
+      setTheme('light-std-std')
+    } else {
+      setTheme('dark-std-std')
+    }
+  }, [theme])
+
+  const router = createBrowserRouter(
```
abhinavrastogi-harness, line 134: we need to remove this and use routes from `routes.tsx` instead

apps/gitness/src/AppMFE.tsx
```diff
@@ -0,0 +1,335 @@
...
+          break
+      }
+      return response
+    }
+  })
+
+  const { theme } = useMFEThemeContext()
+
+  const { setTheme } = useThemeStore()
+  useEffect(() => {
```
abhinavrastogi-harness, line 126: instead of `useEffect` here, can we initialise the ThemeStore with the correct default value instead?
or is this needed for theme switching to work?

apps/gitness/src/AppMFE.tsx
```diff
@@ -0,0 +1,335 @@
...
+    ],
+    {
+      basename: `/ng${renderUrl}`
+    }
+  )
+
+  return (
+    <div id="gitness-shadow-root">
+      <ReactShadowRoot>
+        <style>{`${styles}`}</style>
```
abhinavrastogi-harness, line 313: 🤯 👏

abhinavrastogi-harness: is it a good idea to render a multi-MB string via React?
could we instead use DOM APIs to append the `style` element as a child to the shadow dom parent?

apps/gitness/src/components/breadcrumbs/breadcrumbs-mfe.tsx
```diff
@@ -0,0 +1,57 @@
...
+  BreadcrumbLink,
+  BreadcrumbList,
+  BreadcrumbPage,
+  BreadcrumbSeparator,
+  Topbar
+} from '@harnessio/ui/components'
+
+import { CustomHandle } from '../../routes'
+
+function BreadcrumbsMFE({ selectedProject }: { selectedProject: string }) {
```
abhinavrastogi-harness, line 15: @vardanbansal-harness can you review this change? do we need a complete separate copy of the breadcrumb component? or could we manage this with just the new routes + handles?

apps/gitness/src/components/breadcrumbs/breadcrumbs-utils.ts
```diff
@@ -11,58 +11,76 @@ const repoRoutePathNameMap = [
...
+  }
+
+  const pathToMatch = path.replace('/:spaceId', '')
+
+  const isMatch = matchPath({ path: pathToMatch, end }, pathname)
+
+  return isMatch
+}
+
+export function getBreadcrumbMatchers(pathname: string, isMFE: boolean) {
```
abhinavrastogi-harness, line 26: is this even used anymore? I thought this file was part of the old breadcrumbs. we should not need this anymore with the new routing mechanism.

apps/gitness/src/components/breadcrumbs/breadcrumbs.tsx
abhinavrastogi-harness, line 1: yea, this file should probably not be used anymore. we should delete this, and use the new router breadcrumbs instead.

apps/gitness/src/pages-v2/repo/repo-layout.tsx
```diff
@@ -2,13 +2,20 @@ import { Outlet } from 'react-router-dom'
...
 const RepoLayout = () => {
+  const { scope } = useMFEContext()
+  const isMFE = useIsMFE()
+
   return (
     <>
-      <div className="top-breadcrumbs sticky z-40 bg-background-1">
-        <RepoSubheader useTranslationStore={useTranslationStore} />
+      <div className="top-0 bg-background-1 sticky" style={{ zIndex: 19 }}>
+        {!isMFE ? null : <BreadcrumbsMFE selectedProject={scope.projectIdentifier || '...'} />}
```
abhinavrastogi-harness, line 17: breadcrumbs should not be rendered in page layouts. use router

apps/gitness/webpack.config.cjs
```diff
@@ -0,0 +1,62 @@
...
+          eager: true
+        },
+        'react-dom': {
+          singleton: true,
+          requiredVersion: false,
+          eager: true
+        }
+      }
+    }),
+    new HtmlWebpackPlugin({
```
abhinavrastogi-harness, line 57: why do we need this? MFE will not have any html output, right?

packages/ui/src/components/dialog.tsx
```diff
@@ -46,7 +47,7 @@ const DialogContent = React.forwardRef<React.ElementRef<typeof DialogPrimitive.C
     })
 
     return (
-      <DialogPortal>
+      <DialogPortal container={getGitnessShadowRoot()}>
```
abhinavrastogi-harness, line 50: we need to find a way to avoid hardcoding a reference to gitness in `ui` package

packages/ui/src/components/node-group.tsx
```diff
@@ -33,10 +33,11 @@ function Icon({
     <div className="col-start-1 row-start-1 self-start">
       <div
         className={cn(
-          'border-tertiary-background/30 bg-background text-primary relative z-20 flex h-6 w-6 place-content-center place-items-center rounded-full border p-1',
+          'border-tertiary-background/30 bg-background text-primary relative flex h-6 w-6 place-content-center place-items-center rounded-full border p-1',
           { 'bg-transparent border-none size-1 p-0 mt-2 shadow-commit-list-bullet': simpleNodeIcon },
           className
         )}
+        style={{ zIndex: 15 }}
```
abhinavrastogi-harness, line 40: avoid inline styles. better options in order of priority:
1. lets check with @knagurski if we can enable tailwind in gitness app
2. create semantic classes like `layer-low`, `layer-medium`, `layer-high` to avoid proliferation of 20 different z layers.
3. create a specific class for this style in a local css file

apps/gitness/src/components/breadcrumbs/breadcrumbs-mfe.tsx
vardanbansal-harness: Agreed, I don't see the need for it. All it does, in addition to what we already have, is render `selectedProject` as a fixed breadcrumb, which I think can be managed in the existing component we have.

`<span className="ml-1">{selectedProject}</span>`

apps/gitness/src/components/breadcrumbs/breadcrumbs-utils.ts
vardanbansal-harness: Agreed, this should not be required anymore.

apps/gitness/src/AppMFE.css
praneshg239:
```ts
import styles from '!!raw-loader!@harnessio/ui/styles.css'
```

This statement is to import the styles as raw strings in order to pass to shadow dom.

In this case, we can test whether importing CSS Variables alone in `AppMFE.css` is suffice or not, excluding all other styles. So that we can prevent importing the styles twice.

apps/gitness/src/AppMFE.tsx
praneshg239: `TooltipProvider` is not available in `ui` package. Once that is moved, we can  update it.

vivek-harness: @praneshg239 I think we can move TooltipProvider to UI and import from the UI package. We can ask XD team for a review.

packages/ui/src/components/node-group.tsx
knagurski: Any reason to not just use something like `z-20`? `15` seems oddly specific 🤔

packages/ui/src/components/repo-subheader.tsx
```diff
@@ -16,7 +16,13 @@ export enum RepoTabsKeys {
 
 export const repoTabsKeysArr = Object.values(RepoTabsKeys)
 
-export const RepoSubheader = ({ useTranslationStore }: { useTranslationStore: () => TranslationStore }) => {
+export const RepoSubheader = ({
+  useTranslationStore,
+  showPipelinesTab = true
+}: {
+  useTranslationStore: () => TranslationStore
+  showPipelinesTab?: boolean
```
knagurski, line 24: Generally it's a good pattern to default to `false` for optional booleans, that way they can be controlled with existence as a prop, rather than a value.
```suggestion
  hidePipelinesTab = false
}: {
  useTranslationStore: () => TranslationStore
  hidePipelinesTab?: boolean
```

When calling you can just use `<RepoSubheader hidePipelinesTab ... />` instead of `<RepoSubheader showPipelinesTab={false} ... />`

packages/ui/src/components/node-group.tsx
praneshg239: @knagurski z-index 20 is being used by Harness's sidenav module drawer. So, we have to use z-index values less than 20 for all elements in MFE to prevent overlapping.

packages/ui/src/components/repo-subheader.tsx
praneshg239: Agree. Will make the change.

packages/ui/src/components/node-group.tsx
knagurski: Would `z-10` work instead?

apps/gitness/src/AppMFE.tsx
vivek-harness: Yes, this is needed for theme switching to work in MFE.

PR #572, opened 2024-12-12
Bug bash

vivek-harness:
**Fixed the following UI issues:**
- Chevron sizes are different on dropdowns
- Content on page is not aligned with the tabs and breadcrumbs
- Search box is not vertically centered between heading and tabs
- Tab hover radius is not same as selected radius
- Search highlight should be white
- Branch selector list item content clips vertically

### DIFF
--- a/packages/ui/src/styles.css
+++ b/packages/ui/src/styles.css
@@ -1610,7 +1610,7 @@
 }
 
 mark {
-  background: hsl(var(--canary-background-accent));
+  background: hsl(var(--canary-background-05));
   color: hsl(var(--canary-foreground-06));
 }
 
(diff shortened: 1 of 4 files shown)


praneshg239 (approved):

apps/gitness/src/pages-v2/repo/repo-layout.tsx
```diff
@@ -6,7 +6,7 @@ import { useTranslationStore } from '../../i18n/stores/i18n-store'
 const RepoLayout = () => {
   return (
     <>
-      <div className="fixed top-0 z-30 ml-56 w-full bg-background-1">
+      <div className="breadcrumbs">
```
praneshg239, line 9: This needs to be checked why classes are not getting added

vivek-harness: It used to apply previously, probably there's some change in Tailwind config or else there's some build issue.

PR #577, opened 2024-12-12
fix: repo code/summary branches and commit functionality

shaurya-harness:
fixes :

repo code/summary/commits - branch selector : fetch ‘desc’ order so default branch is first.
adds branch count / tag count to store
path breadcrumbs for file editor, commit modal redirection on new file
divergence for current branch in summary view
Screenshot 2024-12-12 at 8 29 24 PM
Screenshot 2024-12-12 at 8 29 36 PM

wrap long commits for FileLastChangeBar
Screenshot 2024-12-13 at 7 52 26 PM

### DIFF
--- a/apps/gitness/src/components-v2/file-content-viewer.tsx
+++ b/apps/gitness/src/components-v2/file-content-viewer.tsx
@@ -7,6 +7,7 @@ import { CodeEditor } from '@harnessio/yaml-editor'
 
 import GitCommitDialog from '../components-v2/git-commit-dialog'
 import GitBlame from '../components/GitBlame'
+import { useThemeStore } from '../framework/context/ThemeContext'
 import { useDownloadRawFile } from '../framework/hooks/useDownloadRawFile'
 import { useGetRepoRef } from '../framework/hooks/useGetRepoPath'
 import useCodePathDetails from '../hooks/useCodePathDetails'
@@ -42,6 +43,9 @@ export default function FileContentViewer({ repoContent }: FileContentViewerProp
   const [view, setView] = useState<ViewTypeValue>(getDefaultView(language))
   const [isDeleteFileDialogOpen, setIsDeleteFileDialogOpen] = useState(false)
   const { selectedBranchTag } = useRepoBranchesStore()
+  const { theme } = useThemeStore()
+  // TODO: temporary solution for matching themes
+  const monacoTheme = (theme ?? '').startsWith('dark') ? 'dark' : 'light'
 
   /**
    * Toggle delete dialog open state
@@ -68,10 +72,10 @@ export default function FileContentViewer({ repoContent }: FileContentViewerProp
 
   const themeConfig = useMemo(
     () => ({
-      defaultTheme: 'dark',
+      defaultTheme: monacoTheme,
       themes
     }),
-    []
+    [monacoTheme]
   )
 
   const handleDownloadFile = () => {
@@ -125,7 +129,7 @@ export default function FileContentViewer({ repoContent }: FileContentViewerProp
         <CodeEditor
           language={language}
           codeRevision={{ code: fileContent }}
-          onCodeRevisionChange={() => {}}
+          onCodeRevisionChange={() => undefined}
           themeConfig={themeConfig}
           options={{
             readOnly: true
--- a/apps/gitness/src/components-v2/file-content-viewer.tsx
+++ b/apps/gitness/src/components-v2/file-content-viewer.tsx
@@ -89,23 +89,6 @@ export default function FileContentViewer({ repoContent }: FileContentViewerProp
     navigate(`edit/${fullGitRef}/~/${fullResourcePath}`)
   }
 
-  const CodeView = () => {
-    return useMemo(
-      () => (
-        <CodeEditor
-          language={language}
-          codeRevision={{ code: fileContent }}
-          onCodeRevisionChange={() => {}}
-          themeConfig={themeConfig}
-          options={{
-            readOnly: true
-          }}
-        />
-      ),
-      [language, fileContent, themeConfig]
-    )
-  }
-
   return (
     <>
       <GitCommitDialog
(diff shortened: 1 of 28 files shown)


apps/gitness/src/components-v2/file-content-viewer.tsx
```diff
@@ -89,6 +89,23 @@ export default function FileContentViewer({ repoContent }: FileContentViewerProp
...
+          onCodeRevisionChange={() => {}}
+          themeConfig={themeConfig}
+          options={{
+            readOnly: true
+          }}
+        />
+      ),
+      [language, fileContent, themeConfig]
+    )
+  }
```
knagurski, line 107: This is a component that's entirely within another component and will be redefined on every render. Let's move this out to its own thing, or inline it where needed.

shaurya-harness: this is required for Monaco Code editor to update when fileContent changes from repo-code page itself - ex. commiting changes to current file on the currentBranch itself.

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
```diff
@@ -210,7 +219,10 @@ export function RepoSummaryView({
                   currentBranch={{
                     ...selectedBranchTag,
                     // TODO: it is required to transfer the real data that the currently selected branch should contain
```
andrewgolovanov, line 221:
```suggestion
```
same

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
```diff
@@ -210,7 +219,10 @@ export function RepoSummaryView({
...
-                    behindAhead: { ahead: 10, behind: 20 }
+                    behindAhead: {
```
andrewgolovanov, line 220: Since I initially thought that this data would be coming directly from the branch, I structured the props this way. In this case, we can simplify and update the props and types for the component.

For example:
```tsx
<BranchInfoBar
     defaultBranchName={repository?.default_branch}
     spaceId={spaceId}
     repoId={repoId}
     currentBranchDivergence={currentBranchDivergence}
 />
```

BranchInfoBar component
```tsx
interface BranchInfoBarProps {
  defaultBranchName?: string
  spaceId: string
  repoId: string
  currentBranchDivergence: {
    ahead: number
    behind: number
  }
}

export const BranchInfoBar: FC<BranchInfoBarProps> = ({
  defaultBranchName = 'main',
  spaceId,
  repoId,
  currentBranchDivergence = {
    ahead: 0,
    behind: 0
  }
}) => {
  const { ahead, behind } = currentBranchDivergence
  const hasBehind = behind > 0
  const hasAhead = ahead > 0
```

apps/gitness/src/components-v2/file-content-viewer.tsx
knagurski: Having a component defined within a component means that it will be redefined on every render and React will lose track and have to rerender in its entirety every time. Have a look here for a good explanation https://youtu.be/2sAdzy90GtE

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
shaurya-harness: updated with same

apps/gitness/src/components-v2/file-content-viewer.tsx
shaurya-harness: right I understand, what I meant was that this unmount and mount was the only way I could get monaco CodeEditor to update and show changed after a file was changed from edit mode to view mode.
will sync with @srdjan-harness to see if it can be fixed from the editor side

andrewgolovanov (approved):

packages/ui/src/views/repo/components/branch-info-bar.tsx
```diff
@@ -7,23 +7,21 @@ interface BranchInfoBarProps {
...
   repoId,
-  currentBranch
+  currentBranchDivergence
 }) => {
-  const { behindAhead } = currentBranch
-  const hasBehind = behindAhead.behind > 0
-  const hasAhead = behindAhead.ahead > 0
+  const { behind, ahead } = currentBranchDivergence
+  const hasBehind = behind && behind > 0
+  const hasAhead = ahead && ahead > 0
```
knagurski, line 24: Not an issue, but looks like a good case for checking for truthyness instead of existence and being over 0.
```suggestion
  const hasBehind = !!behind
  const hasAhead = !!ahead
```

apps/gitness/src/pages-v2/repo/repo-summary.tsx
```diff
@@ -53,15 +62,38 @@ export default function RepoSummaryPage() {
...
+    repo_ref: repoRef,
+    queryParams: { include_commit: false, sort: 'date', order: 'asc', limit: 20, page: 1 }
+  })
+
+  const { branch_count, default_branch_commit_count, pull_req_summary, tag_count } = repoSummary || {}
+
   const { data: { body: branches } = {} } = useListBranchesQuery({
     repo_ref: repoRef,
-    queryParams: { include_commit: false, sort: 'date', order: 'asc', limit: 20, page: 1, query: '' }
+    queryParams: { include_commit: false, sort: 'date', order: orderSortDate.DESC, limit: branch_count }
```
knagurski, line 74: What if there were thousands of branches?

apps/gitness/src/components-v2/file-content-viewer.tsx
shaurya-harness: reverting this as @srdjan-harness fixed the changes from Editor side in this PR - https://github.com/harness/canary/pull/585

apps/gitness/src/pages-v2/repo/repo-summary.tsx
shaurya-harness: yeah I've asked Hitesh to share if the api can support huge list of branches, otherwise will change it to query based approach.
the current implementation just unblocks us but is also similar to how github handles it.

shaurya-harness: reverted this after discussion with Hitesh, will raise a new PR with search query hooked to the api

PR #584, opened 2024-12-13
feat: import repo page

shaurya-harness:
feat: import repo page

Screenshot 2024-12-13 at 3 26 04 PM
Screenshot 2024-12-16 at 1 43 52 PM
Screenshot 2024-12-16 at 2 47 42 PM


packages/ui/src/views/repo/repo-import/index.tsx
knagurski, line 1: Can we avoid using `index` files for components/types/logic etc, only use them as indexes to get at things more easily. Otherwise it's super annoying to find what you need quickly without having to traverse the directories. Ideally a file should export a single component and the name of the file should match that.

packages/ui/src/views/repo/repo-list/repo-list.tsx
```diff
@@ -100,25 +100,27 @@ export function RepoList({
...
-                }
-                description={<Stats stars={repo.stars} pulls={repo.pulls} />}
-                right
-                label
-                secondary
-              />
+              {!repo.importing ? (
+                <StackedList.Field
+                  title={
+                    <>
```
knagurski, line 114: Doesn't look like this fragment is needed

packages/ui/src/views/repo/repo-list/repo-list.tsx
```diff
@@ -100,25 +100,27 @@ export function RepoList({
...
+                      Updated <em>{repo.timestamp}</em>
```
knagurski, line 115: Can we get this into the strings?

shaurya-harness: there's quite a bit of strings in this component, I will move them all

packages/ui/src/views/repo/repo-import/index.tsx
shaurya-harness: updated

packages/ui/src/views/repo/repo-list/repo-list-page.tsx
```diff
@@ -14,7 +15,16 @@ import { sortRepositories } from '../utils/sorting/repos'
 import { RepoList } from './repo-list'
 import { RepoListProps } from './types'
 
-const LinkComponent = ({ to, children }: { to: string; children: ReactNode }) => <Link to={to}>{children}</Link>
+const LinkComponent = ({ to, children, disabled }: { to: string; children: ReactNode; disabled?: boolean }) => (
```
abhinavrastogi-harness, line 18: why is this needed?
if needed, LinkComponent should be defined in the app layer and passed to the view.
View should not be defining a LinkComponent

shaurya-harness: yes not required, probably transferred from old views of playground, removing this

packages/ui/src/views/repo/repo-list/repo-list.tsx
shaurya-harness: fixed

vardanbansal-harness (approved):

PR #585, opened 2024-12-13
Multiple fixes in code and other editors

srdjan-harness:
All changes are related to editors (CodeEditor, YamlEditor, ...)

- Fix/improve re-rendering issue.
- Fix/improve theme selection - removed not need code and exposed theme prop.

Following changes improves and fix the issue but they are not final implementations.
- Replace hardcode colors with css var and tokens.  
  This has to be improved. We should create our own tokens in monaco theme and remove hardcoded values.

- Align code editors theme with app theme.
  We have only two themes and we are picking which one to render this way: 
  `const monacoTheme = (theme ?? '').startsWith('dark') ? 'dark' : 'light'`


apps/gitness/src/components-v2/file-content-viewer.tsx
```diff
@@ -42,6 +43,9 @@ export default function FileContentViewer({ repoContent }: FileContentViewerProp
   const [view, setView] = useState<ViewTypeValue>(getDefaultView(language))
   const [isDeleteFileDialogOpen, setIsDeleteFileDialogOpen] = useState(false)
   const { selectedBranchTag } = useRepoBranchesStore()
+  const { theme } = useThemeStore()
```
knagurski, line 46: We should probably alter `useThemeStore` to _always_ have a theme instead of potentially being `undefined`. Not a problem for this PR

apps/gitness/src/components-v2/file-editor.tsx
```diff
@@ -34,18 +35,21 @@ export const FileEditor = ({ repoDetails, defaultBranch }: FileEditorProps) => {
...
   const [isCommitDialogOpen, setIsCommitDialogOpen] = useState(false)
   const { selectedBranchTag } = useRepoBranchesStore()
+  const { theme } = useThemeStore()
+  // TODO: temporary solution for matching themes
+  const monacoTheme = (theme ?? '').startsWith('dark') ? 'dark' : 'light'
 
   const themeConfig = useMemo(
     () => ({
-      defaultTheme: 'dark',
+      defaultTheme: monacoTheme,
```
knagurski, line 49: Looks like you're having to repeat yourself in a bunch of places. It might make sense to make a theme-aware editor component that you can reuse. Or extract a custom hook to get the theme.

```typescript
import { useMemo } from 'react'

import { ThemeDefinition } from '@harnessio/yaml-editor'

import type { useThemeStore } from '../framework/context/ThemeContext.tsx'
import { themes } from '../pages/pipeline-edit/theme/monaco-theme.ts'

export interface UseThemeConfigPayload {
  defaultTheme: string
  themes: ThemeDefinition[]
}

export function useThemeConfig(): UseThemeConfigPayload {
  const { theme } = useThemeStore()

  return useMemo(
    () => ({
      defaultTheme: (theme || '').startsWith('dark') ? 'dark' : 'light',
      themes
    }),
    [theme]
  )
}

```

Then you can just call...
```typescript
const themeConfig = useThemeConfig()
```
in all the places you need it

shaurya-harness (approved):

apps/gitness/src/components-v2/file-editor.tsx
srdjan-harness: nice, will be done in the one of the follow up PRs.

PR #600, opened 2024-12-16
Integrate create rules flow to UI package

sans-harness:
https://github.com/user-attachments/assets/bd09cea4-b294-4d5d-a19d-6b5f67be234c

Minor fixes: 

Screenshot 2024-12-17 at 10 39 50 AM


packages/ui/src/views/repo/repo-branch-rules/components/repo-branch-rules-data.tsx
```diff
@@ -0,0 +1,64 @@
+import { MergeStrategy } from '@views/repo/pull-request'
+
+import { BranchRuleId } from '../types'
+
+export const branchRules = [
```
abhinavrastogi-harness, line 5: any type needed on this?

packages/ui/src/views/repo/repo-branch-rules/index.tsx
abhinavrastogi-harness, line 1: filename should not be `index.tsx`
match filename to the main component being exported, and re-export from `index.ts` if needed

packages/ui/src/views/repo/repo-branch-rules/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+      patterns: [],
+      state: true,
+      default: false,
+      repo_owners: false,
+      bypass: [],
+      rules: []
+    }
+  })
+
+  const [rules, dispatch] = useReducer(
```
abhinavrastogi-harness, line 66: could we move this into the store or into the app layer? i dont think we should be tying the view to the data so tightly

packages/ui/src/views/repo/repo-branch-rules/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+        pattern: '',
+        patterns: presetRuleData?.patterns || [],
+        state: presetRuleData?.state && true,
+        default: presetRuleData?.default || false,
+        repo_owners: presetRuleData?.repo_owners || false,
+        bypass: presetRuleData?.bypass || [],
+        rules: []
+      })
+
+      dispatch({
```
abhinavrastogi-harness, line 97: same. this should just be a callback from the view, and move the logic into the app layer

packages/ui/src/views/repo/repo-branch-rules/reducers/repo-branch-rules-reducer.ts
```diff
@@ -0,0 +1,52 @@
+import { MergeStrategy } from '@views/repo/pull-request'
+
+import { BranchRulesAction, BranchRulesActionType, Rule } from '../types'
+
+export const branchSettingsReducer = (state: Rule[], action: BranchRulesAction): Rule[] => {
```
abhinavrastogi-harness, line 5: reducer should be in app layer maybe? in the store actually if possible

packages/ui/src/views/repo/repo-branch-rules/components/repo-branch-rules-data.tsx
sans-harness: TS seems happy for now, lets add one in a follow up PR if its not a blocker for this one.

abhinavrastogi-harness (approved):

PR #605, opened 2024-12-17
fix: Branch repo store

shaurya-harness:
moved BranchSelectorStore interface to UI package and fixed the correct types
added transform utils for setting branchList to the the store
added changes to fetch branch/tag based on query hooked to the api for the selector dropdown

### DIFF
--- a/apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
+++ b/apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
@@ -1,4 +1,5 @@
 import { TypesBranchExtended, TypesCommitDivergence } from '@harnessio/code-service-client'
+import { BranchData } from '@harnessio/ui/views'
 
 import { timeAgoFromISOTime } from '../../../pages/pipeline-edit/utils/time-utils'
 
@@ -6,7 +7,7 @@ export const transformBranchList = (
   branches: TypesBranchExtended[],
   defaultBranch?: string,
   divergence?: TypesCommitDivergence[]
-) => {
+): BranchData[] => {
   return branches.map((branch, index) => {
     const { ahead, behind } = divergence?.[index] || {}
     return {
--- a/apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
+++ b/apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
@@ -14,7 +14,7 @@ export const transformBranchList = (
       name: branch.name || '',
       sha: branch.commit?.sha || '',
       timestamp: branch.commit?.committer?.when ? timeAgoFromISOTime(branch.commit.committer.when) : '',
-      default: branch.name === defaultBranch || false,
+      default: branch.name === defaultBranch || branch.is_default || false,
       user: {
         name: branch.commit?.committer?.identity?.name || '',
         avatarUrl: ''
@@ -22,7 +22,7 @@ export const transformBranchList = (
       behindAhead: {
         behind: behind || 0,
         ahead: ahead || 0,
-        default: defaultBranch === branch.name
+        default: defaultBranch === branch.name || branch.is_default || false
       }
     }
   })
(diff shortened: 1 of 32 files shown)


apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
```diff
@@ -0,0 +1,29 @@
+import { TypesBranchExtended, TypesCommitDivergence } from '@harnessio/code-service-client'
+
+import { timeAgoFromISOTime } from '../../../pages/pipeline-edit/utils/time-utils'
+
+export const transformBranchList = (
```
abhinavrastogi-harness, line 5: i think we should add a return type on this function. it should be available to be imported from the `ui` package

packages/ui/src/views/repo/components/branch-selector/branch-selector-dropdown.tsx
```diff
@@ -59,14 +52,15 @@ export const BranchSelectorDropdown: FC<BranchSelectorDropdownProps> = ({
...
         )}
-
-        <SearchBox.Root
-          className="mt-2 w-full"
-          placeholder={BRANCH_SELECTOR_LABELS[activeTab].searchPlaceholder}
-          value={searchQuery}
-          handleChange={handleSearchChange}
-          showOnFocus
-        />
+        <div role="presentation" onKeyDown={e => e.stopPropagation()}>
```
abhinavrastogi-harness, line 55: why are we adding this?

packages/ui/src/views/repo/components/branch-selector/branch-selector.tsx
```diff
@@ -21,16 +23,20 @@ export const BranchSelector: FC<BranchSelectorProps> = ({
...
 
   const isTag = selectedBranchTag
     ? tagList?.some(tag => tag.name === selectedBranchTag.name && tag.sha === selectedBranchTag.sha)
     : false
 
+  const [open, setOpen] = useState(false)
+
   return (
-    <DropdownMenu>
+    <DropdownMenu open={open} onOpenChange={setOpen}>
```
abhinavrastogi-harness, line 39: why does this need to be a controlled component? where are we using `setOpen`? i dont see it being passed anywhere

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -41,11 +42,13 @@ export const RepoCommitsView = (props: RepoCommitsViewProps) => {
         </Text>
         <Spacer size={6} />
         <div className="flex justify-between gap-5">
-          {!props.isFetchingBranches && branchList && (
+          {branchList && (
```
abhinavrastogi-harness, line 45: not a blocker, but I think we should always show the branch selector, but disable it with a loader until branch list is available

abhinavrastogi-harness (approved):

packages/ui/src/views/repo/components/branch-selector/branch-selector-dropdown.tsx
shaurya-harness: branch selector takes the focus to DropdownMenuItems as soon as we enter something in the searchbox, so just to trap focus to searchbox till we are typing.

apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
shaurya-harness: yes I can add the return type.

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
shaurya-harness: agreed, will fix this

shaurya-harness: branchList can change whilst we are typing in the searchbox as it is now hooked to the api based on the query, so this approach would break the component.

shaurya-harness: will remove the check for branchList as that is handled within nodata state inside the BranchSelector

packages/ui/src/views/repo/components/branch-selector/branch-selector.tsx
shaurya-harness: yes not required anymore, initially I added to trap focus but handled from within the component, removing this

PR #610, opened 2024-12-19
fix: fix repo commits

3em:
- Fixed repository commits UI
- Fixed pagination logic
- Fixed state management for branch data
- Added fake filters
- Added commits to preview display
- Fix preview layout for Files

image

image


apps/design-system/src/views/repo-commits/index.tsx
```diff
@@ -0,0 +1,48 @@
+import { useCallback, useState } from 'react'
+
+import { BranchSelectorTab, IBranchSelectorStore, RepoCommitsView as RepoCommitsUiView } from '@harnessio/ui/views'
+
+import { noop, useTranslationsStore } from '../../utils.ts'
+import { repoFilesStore } from '../repo-files/components/repo-files-store.ts'
+import { repoCommitsStore } from './repo-commits-store.ts'
```
knagurski, line 7: We shouldn't need the `.ts` on imports
```suggestion
import { noop, useTranslationsStore } from '../../utils'
import { repoFilesStore } from '../repo-files/components/repo-files-store'
import { repoCommitsStore } from './repo-commits-store'
```

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
```diff
@@ -99,12 +99,18 @@ export function RepoSummaryView({
...
 
-  if (loading) return <SkeletonList />
+  if (loading)
+    return (
+      <SandboxLayout.Main fullWidth>
+        <SandboxLayout.Content>
+          <SkeletonList />
+        </SandboxLayout.Content>
+      </SandboxLayout.Main>
+    )
```
knagurski, line 109: As a general rule, always add parens when the body of a condition can't fit on the same line as the `if`
```suggestion
  if (loading) {
    return (
      <SandboxLayout.Main fullWidth>
        <SandboxLayout.Content>
          <SkeletonList />
        </SandboxLayout.Content>
      </SandboxLayout.Main>
    )
  }
```

Makes it easier to visually scan and see the scope.

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -29,55 +30,97 @@ interface RepoCommitsViewProps {
...
+                description={[
+                  isDirtyList
+                    ? 'There isn’t any commit history to show here for the selected user, time range, or current page.'
+                    : "Your commits will appear here once they're made. Start committing to see your changes reflected."
+                ]}
+                primaryButton={
+                  isDirtyList
+                    ? { label: 'Clear filters', onClick: handleResetFiltersAndPages }
+                    : // TODO: add onClick for Creating new commit
+                      { label: 'Create new commit' }
```
knagurski, line 106: If we have time, can we get these into the translations?

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -29,55 +30,97 @@ interface RepoCommitsViewProps {
...
+  const FILTER_OPTIONS = getFilterOptions(t)
+  const SORT_OPTIONS = getSortOptions(t)
+  const SORT_DIRECTIONS = getSortDirections(t)
+  const filterHandlers = useFilters()
+
+  const isDirtyList = page !== 1
+
+  const handleResetFiltersAndPages = () => {
+    setPage(1)
+  }
```
knagurski, line 57: We probably want to reset to the 1st page on change of the filtering as well as clearing filters, otherwise the user could get stuck on a non-existent page if they change page with a vague query and then narrow down.

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
```diff
@@ -29,55 +30,97 @@ interface RepoCommitsViewProps {
...
+                iconName={isDirtyList ? 'no-search-magnifying-glass' : 'no-data-commits'}
+                title={isDirtyList ? 'No commits history' : 'No commits yet'}
+                description={[
+                  isDirtyList
+                    ? 'There isn’t any commit history to show here for the selected user, time range, or current page.'
+                    : "Your commits will appear here once they're made. Start committing to see your changes reflected."
+                ]}
+                primaryButton={
+                  isDirtyList
+                    ? { label: 'Clear filters', onClick: handleResetFiltersAndPages }
```
knagurski, line 104: Looks like the handler only sets to the first page, I take it that's a temporary measure?

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
```diff
@@ -1,72 +1,68 @@
...
               <StackedList.Root>
                 {commitData.map((commit, repo_idx) => {
                   const authorName = commit.author?.identity?.name
 
                   return (
                     <StackedList.Item
-                      className="hover:bg-transparent"
-                      key={commit.title}
+                      className="!cursor-default items-start py-3"
+                      key={`${commit.title}-${repo_idx}`}
```
knagurski, line 46: Could we just use the SHA from the commit to simplify things?

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
```diff
@@ -1,72 +1,68 @@
...
   const totalNodes = entries.length
 
   return (
-    <>
+    <div>
       {entries.map(([date, commitData], node_idx) => (
-        <NodeGroup.Root key={date}>
+        <NodeGroup.Root
+          className="grid-cols-[4px_1fr] gap-x-[22px] gap-y-3.5 pb-6 last:pb-0"
+          key={`${date}-${node_idx}`}
```
knagurski, line 33: Is the date not unique?

packages/ui/src/views/repo/repo-commits/repo-commits-view.tsx
3em: Added filter reset to `handleResetFiltersAndPages`

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
```diff
@@ -1,72 +1,68 @@
...
           group[date] = (group[date] || []).concat(commit)
           return group
-        },
-        {} as Record<string, TypesCommit[]>
-      ) || {},
-    [data]
-  )
+        }, {})
+
+    return Object.entries(commitsGroupedByDate) as [string, TypesCommit[]][]
```
knagurski, line 23: Looks like you don't need the type override here, TS seems to infer it properly.
```suggestion
    return Object.entries(commitsGroupedByDate)
```

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
3em: I somehow caught this case during testing, so I added the index as well.

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
3em: Perhaps this is an issue specific to my IDE

image

packages/ui/src/views/repo/repo-commits/components/commits-list.tsx
knagurski: Ohh... I get it, date might be an empty string (should be impossible in practice, but `when` is optional in the types 🤔). I don't think React cares if the `key` is blank, as long as it's unique and exists

PR #611, opened 2024-12-19
Update light-std-std styles.css

cxramina:
Update for --canary-success token. Previously, the value for the light theme incorrectly included a dark mode variable.

Before:
image

After:
image


knagurski (approved):
🥳

PR #612, opened 2024-12-19
feat: [PIPE-24233]: pipeline create dialog

srdjan-harness:
Screenshot 2024-12-19 at 15 14 30

### DIFF
--- a/apps/gitness/src/pages-v2/pipeline/create-pipeline-dialog.tsx
+++ b/apps/gitness/src/pages-v2/pipeline/create-pipeline-dialog.tsx
@@ -1,0 +1,64 @@
+import { useMemo } from 'react'
+import { useNavigate, useParams } from 'react-router-dom'
+
+import { useCreatePipelineMutation, useFindRepositoryQuery, useListBranchesQuery } from '@harnessio/code-service-client'
+import { CreatePipelineDialog as CreatePipelineDialogView, CreatePipelineFormType } from '@harnessio/ui/views'
+
+import { useGetRepoRef } from '../../framework/hooks/useGetRepoPath'
+import { useGetSpaceURLParam } from '../../framework/hooks/useGetSpaceParam'
+import { PathParams } from '../../RouteDefinitions'
+import { apiBranches2BranchNames, apiRepository2DefaultBranchName } from '../repo/transform-utils/branch-transform'
+
+interface CreatePipelineDialogProps {
+  open: boolean
+  onClose: () => void
+}
+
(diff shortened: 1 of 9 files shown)


vardanbansal-harness (approved):

apps/gitness/src/pages-v2/pipeline/create-pipeline-dialog.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const navigate = useNavigate()
+
+  const { data: { body: branches } = {}, isLoading: isLoadingBranchNames } = useListBranchesQuery({
+    repo_ref: repoRef,
+    queryParams: {}
+  })
+  const branchNames = useMemo(() => apiBranches2BranchNames(branches), [branches])
+
+  const { data: { body: repositoryData } = {}, isLoading: isLoadingDefaultBranch } = useFindRepositoryQuery({
+    repo_ref: repoRef
```
abhinavrastogi-harness, line 30: branchlist and default branch may already be available in the branches store. we should check there first before making an api call or blocking the create pipeline form.
also, if we do make the api call, we should update the store.

apps/gitness/src/pages-v2/pipeline/create-pipeline-dialog.tsx
```diff
@@ -0,0 +1,64 @@
...
+    try {
+      await savePipeline({
+        repo_ref: repoRef,
+        body: { config_path: yamlPath, default_branch: branch, identifier: name }
+      })
+
+      navigate(`/spaces/${spaceId}/repos/${repoId}/pipelines/${name}/edit`)
+    } catch (e) {
+      //TODO: Handle error by showing toast
+      console.log(e, 'error')
```
abhinavrastogi-harness, line 48: instead of console log or toast, let's just show the error inline in the modal right away.

apps/gitness/src/pages-v2/repo/transform-utils/branch-transform.ts
```diff
@@ -28,3 +33,11 @@ export const transformBranchList = (
...
   })
 }
+
+export function apiBranches2BranchNames(apiBranches?: ListBranchesOkResponse) {
+  return apiBranches?.map(apiBranch => apiBranch.name ?? '')
+}
+
+export function apiRepository2DefaultBranchName(apiRepo?: RepoRepositoryOutput, defaultBranch = 'main') {
+  return apiRepo?.default_branch ?? defaultBranch
+}
```
abhinavrastogi-harness, line 43: i believe these should not be needed. we already have a branch store which handles this

PR #614, opened 2024-12-19
feat: Refactoring Breadcrumbs

vardanbansal-harness:
Refactor Breadcrumbs implementation based on `handle` prop. (see https://reactrouter.com/6.28.0/hooks/use-matches#breadcrumbs)

https://github.com/user-attachments/assets/4f708f8c-e725-4eda-9a43-d6f553257cd7


apps/gitness/src/routes.tsx
abhinavrastogi-harness, line 1: the path `/admin/default-settings` seems to be missing from this file. it works in the older routes.
let's make sure we didn't miss moving any paths.

apps/gitness/src/AppV1.tsx
abhinavrastogi-harness, line 1: the naming is confusing. 
`v1` was used for the old pages we had (the ones we are deleting now)
i'd consider `v2` to be the current architecture.
maybe this file should be `AppV2` instead, to go along with the new pages and components.

apps/gitness/src/styles.css
```diff
@@ -3,3 +3,11 @@
 @import '@harnessio/canary/src/styles.css'; */
 
 @import 'highlight.js/styles/atom-one-dark.css';
+
+:root {
+  --breadcrumbs-height: 55px;
+}
+
+.top-breadcrumbs {
```
abhinavrastogi-harness, line 11: can we move this to a tailwind utility  instead of a custom class? it'll be more consistent

apps/gitness/src/routes.tsx
vardanbansal-harness: Done, added it back and checked for other routes as well.

apps/gitness/src/AppV1.tsx
vardanbansal-harness: Makes sense, done.

abhinavrastogi-harness (approved):

PR #622, opened 2024-12-20
fix: fix webhooks list and create / edit form

3em:
- Fixed Webhooks list UI and create /edit form
- Added delete popup and delete handlers
- Added Webhooks preview

Pasted Graphic 4
Pasted Graphic 5
Pasted Graphic 6
image


apps/gitness/src/pages-v2/webhooks/webhook-list.tsx
```diff
@@ -78,7 +106,26 @@ export default function WebhookListPage() {
 
   useEffect(() => {
     setQueryPage(page)
+
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [page, queryPage, setPage])
-  return <RepoWebhookListPage useWebhookStore={useWebhookStore} useTranslationStore={useTranslationStore} />
+
+  return (
+    <div>
```
andrewgolovanov, line 114:
```suggestion
    <>
```
I think we can just use a Fragment here.

apps/design-system/src/views/repo-webhooks-list/index.tsx
sans-harness, line 1: We should not be adding code to `index.ts`. Lets use `index` only to import/export files.

andrewgolovanov (approved):

PR #627, opened 2024-12-30
Create design-tokens-gitness

cxramina:
For the design team, we need a dedicated folder in GitHub for testing purposes. We will create design tokens using the Token Studio plugin for Figma and then push them to Git. This environment will serve as a place to store and test these tokens until we figure out the best way to deliver all new tokens to the code properly.


abhinavrastogi-harness (approved):

packages/design-tokens-gitness
abhinavrastogi-harness, line 1: actually, if there are going to be specific to `gitness`, the open-source version of Harness, then we should keep them in `apps/gitness/design-tokens` folder instead

PR #631, opened 2025-01-03
feat: enhance PR layout with tab nav and UI improvements

andrewgolovanov:
**Description**:
This pull request introduces a refactoring of the header components for the detailed `PR` page as well as style adjustments. 

**TODO** (points that need to be finalized and fixed in separate PR's)

- Refactor Page Construction: in `apps/gitness`, revise the way the page is built. Currently, the page is constructed entirely from individual components. Instead, we should use a single component named `PullRequestConversationView` and pass it the required props, similar to the approach used on other pages.
- In the `Overview section` on the `Conversation tab`, the commit history need to be displayed (currently missing), and the components and styles require refactoring.
- A state needs to be added for the dropdown where we select how we plan to merge this PR (currently, it works as static mock data).
- Add functionality to the `form` buttons (on the Write tab) to use and edit md markup (right now they are static), and check the styles for the Preview tab.

**Screenshots**:
Before:
Arc 2025-01-09 21 17 26

After:
Arc 2025-01-09 21 24 27

### DIFF
--- a/packages/ui/src/views/layouts/PullRequestLayout.tsx
+++ b/packages/ui/src/views/layouts/PullRequestLayout.tsx
@@ -1,6 +1,6 @@
 import { NavLink, Outlet } from 'react-router-dom'
 
-import { Badge, Icon, Spacer } from '@components/index'
+import { Badge, Icon, Spacer, Tabs, TabsList, TabsTrigger } from '@components/index'
 import { TranslationStore } from '@views/repo'
 import { PullRequestHeader } from '@views/repo/pull-request/components/pull-request-header'
 import { IPullRequestStore } from '@views/repo/pull-request/pull-request.types'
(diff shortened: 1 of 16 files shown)


packages/ui/src/views/layouts/PullRequestLayout.tsx
```diff
@@ -14,72 +15,89 @@ interface PullRequestLayoutProps {
...
   const { pullRequest } = usePullRequestStore()
   const { t } = useTranslationStore()
 
-  const baseClasses =
-    'inline-flex items-center justify-center px-3 py-1 px-4 items-center gap-2 bg-background hover:text-primary h-[36px] rounded-tl-md rounded-tr-md m-0'
-  const getLinkClasses = (isActive: boolean) => {
-    return `${baseClasses} ${isActive ? 'text-primary [&svg]:text-primary tabnav-active' : 'tabnav-inactive'}`
-  }
+  const activeTab = useMemo(() => {
+    if (location.pathname.includes(PullRequestTabsKeys.COMMITS)) {
```
abhinavrastogi-harness, line 37: using `location` for determining active tab is a regression. react-router already provides the active path, which was being used earlier. 
why do we need to make this change?

abhinavrastogi-harness (dismissed):

PR #639, opened 2025-01-08
Nav and user fixes

sans-harness:
- Sets spaceId as a parent route.
- Updates navbar routes to ensure they link to the correct destinations.
- Fixes the profile-settings and admin-user-settings pages, placing them at the appropriate routes and making them accessible through the navbar.
- Adjusts project-user-settings behavior: if a project is selected, it opens that project; if no project is selected, the first project is opened by default.

https://github.com/user-attachments/assets/fea8177b-72ec-4f32-b504-a6b68e669b0d


abhinavrastogi-harness:
Notices a few issues in the demo recording:

- Breadcrumbs look broken for Account scope pages. I think we should add a first segment so that it looks like `Account / Users`
Screenshot 2025-01-08 at 9 02 46 AM

- Same issue on User Profile. Breadcrumbs should be `User / Settings / General`
Screenshot 2025-01-08 at 9 03 40 AM

- When you select the first project automatically, breadcrumbs are not showing the project selector
Screenshot 2025-01-08 at 9 04 49 AM

- there's an extra `/` before the project selector when you're on a non-first project.
Screenshot 2025-01-08 at 9 04 32 AM

sans-harness:
https://github.com/user-attachments/assets/500372a2-3f67-4651-abc9-e61aea0e3dc8

apps/gitness/src/components-v2/app-shell.tsx
```diff
@@ -68,7 +71,7 @@ const AppShell = () => {
    * Map mock data menu by type to Settings and More
    */
   const { moreMenu, settingsMenu } = useMemo(() => {
-    const navbarMenuData = getNavbarMenuData(t)
+    const navbarMenuData = getNavbarMenuData(t, space_ref ?? spaces[0].path)
```
abhinavrastogi-harness, line 74: `spaces[]` might be empty if the API call is not complete yet, so `spaces[0]` can throw an exception.
we should add a length check on `spaces`

apps/gitness/src/pages-v2/project/project-member-list.tsx
```diff
@@ -19,7 +19,7 @@ import { orderSortDate } from '../../types'
 import { usePrincipalListStore } from '../account/stores/principal-list-store'
 import { useMemberListStore } from './stores/member-list-store'
 
-export function ProjectMemberListPage() {
+export const ProjectMemberListPage = () => {
```
abhinavrastogi-harness, line 22: why this change?

apps/gitness/src/routes.tsx
```diff
@@ -58,237 +58,296 @@ export const routes: CustomRouteObject[] = [
...
+                  breadcrumb: () => <Text>Members</Text>
+                }
+              }
+            ]
           }
         ]
       },
+
+      {
+        path: ':spaceId/pipelines',
```
abhinavrastogi-harness, line 321: this should also be moved to a child of `:spaceId` object

abhinavrastogi-harness (approved):

apps/gitness/src/pages-v2/project/project-member-list.tsx
sans-harness: reverted. Im just more used to the ES6 arrow fn notation.

PR #644, opened 2025-01-09
feat: advanced theme selector

abhinavrastogi-harness:
Adds advanced theme selector with Mode, Color and Contrast options in the dropdown menu.

Screenshot 2025-01-09 at 9 38 17 AM

### DIFF
--- a/packages/ui/src/components/navbar/navbar-user/index.tsx
+++ b/packages/ui/src/components/navbar/navbar-user/index.tsx
@@ -1,9 +1,12 @@
+import { useMemo } from 'react'
 import { Link } from 'react-router-dom'
 
 import {
   Avatar,
   AvatarFallback,
   AvatarImage,
+  ColorType,
+  ContrastType,
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
@@ -16,8 +19,10 @@ import {
   DropdownMenuSubTrigger,
   DropdownMenuTrigger,
   FullTheme,
+  getModeColorContrastFromFullTheme,
   Icon,
   IThemeStore,
+  ModeType,
   Text
 } from '@/components'
 import { TypesUser } from '@/types'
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/navbar/navbar-user/index.tsx
```diff
@@ -107,15 +113,41 @@ export const NavbarUser = ({
...
           <DropdownMenuPortal>
             <DropdownMenuSubContent>
+              <DropdownMenuItem disabled>Color Scheme</DropdownMenuItem>
+              <DropdownMenuSeparator />
               <DropdownMenuRadioGroup
-                value={theme}
+                value={mode}
                 onValueChange={value => {
-                  setTheme(value as FullTheme)
+                  setTheme(`${value}-${color}-${contrast}` as FullTheme)
```
knagurski, line 121: Looks like this is repeated a few times, might be good to extract a small utility function

```typescript
function formatFullTheme(theme: string, color: string, contrast: string): FullTheme {
  return `${theme}-${color}-${contrast}` as FullTheme
}
```

packages/ui/src/components/navbar/navbar-user/index.tsx
```diff
@@ -107,15 +113,41 @@ export const NavbarUser = ({
...
                 }}
               >
-                <DropdownMenuRadioItem value="light-std-std">Light</DropdownMenuRadioItem>
-                <DropdownMenuRadioItem value="dark-std-std">Dark</DropdownMenuRadioItem>
-                <DropdownMenuRadioItem value="system-std-std">System</DropdownMenuRadioItem>
+                <DropdownMenuRadioItem value={ModeType.System}>System</DropdownMenuRadioItem>
```
knagurski, line 124: Worth extracting all the strings into the i18n files?

knagurski (approved):
A couple of non-critical suggestions

packages/ui/src/components/navbar/navbar-user/index.tsx
abhinavrastogi-harness: the function param is different in all three places, so we'll end up with the same duplication again 🤔

PR #648, opened 2025-01-10
Fix settings page and rules form

3em:
- Made fixes on the General Settings page
- Fixed the display of rules
- Updated the form for creating and editing Rules
- Extracted reusable components: DeleteAlertDialog, ButtonWithOptions, MoreActionsTooltip, MultiSelect
- Fixed most of the code issues identified by eslint throughout the UI section

image
image


sans-harness:
The include/exclude button for the target patterns field should be rendered inside the input field as a React node. The current "include/exclude" button on the right-hand side should be replaced with a simple "Add" button. Users will first select whether to include or exclude a pattern within the input field and then click the "Add" button to add the pattern.

cc: @abhinavrastogi-harness

.eslintrc.json
```diff
@@ -58,7 +58,8 @@
 
     "@typescript-eslint/no-explicit-any": "warn",
 
-    "tailwindcss/no-custom-classname": "warn"
+    "tailwindcss/no-custom-classname": "off",
+    "tailwindcss/enforces-negative-arbitrary-values": "off"
```
abhinavrastogi-harness, line 62: why do we need to disable any of these rules? I think we should discuss/review these as they can impact the long-term quality of the codebase.
cc @praneshg239 @knagurski

packages/ui/src/components/more-actions-tooltip.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export interface MoreActionsTooltipProps {
+  actions: ActionData[]
+}
+
+/**
+ * A component for displaying a button that, when clicked, shows a tooltip with a list of actions.
+ */
+export const MoreActionsTooltip: FC<MoreActionsTooltipProps> = ({ actions }) => {
```
abhinavrastogi-harness, line 29: this kind of a component brings up interesting thoughts.

while this is useful to standardise all context menus, it also limits customisation which we get from composing lower layer components directly.

for eg. this component doesn't allow for icons, keyboard shortcuts or checkbox/radio items to be shown. Whereas the underlying DropdownItem component supports that. This means over time as requirements change, we will end up with multiple versions of this Tooltip component, with different features.

Hard to say which approach is better. cc @knagurski thoughts?

packages/ui/src/components/multi-select.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  useEffect(() => setSearch(searchValue ?? ''), [searchValue])
+
+  useEffect(() => {
+    const debouncedChangeSearch = debouncedChangeSearchRef.current
+
+    return () => {
+      debouncedChangeSearch.cancel()
+    }
+  }, [])
```
abhinavrastogi-harness, line 55: i see this pattern repeated multiple times.. of creating refs for debounce functions which are then cancelled in a useEffect.
why is this needed?

packages/ui/src/components/multi-select.tsx
abhinavrastogi-harness, line 1: can we add screenshots in the PR, and docs in the `design-system` app for new components?
cc @knagurski

packages/ui/src/components/text.tsx
```diff
@@ -80,7 +81,7 @@ interface TextProps extends React.ComponentProps<'span'> {
   /**
    * Sets the text size.
    */
-  size?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
+  size?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13
```
abhinavrastogi-harness, line 84: do we really use 13 different sizes of a text component?

packages/ui/src/icons/green-tick-nobg.svg
abhinavrastogi-harness, line 1: if the color is expected to be applied via css (which is indeed the right behavior in this case), we can change the name to be something like `tick-circle.svg`.
color and bg state ideally don't need to represented in the filename

abhinavrastogi-harness:
> The include/exclude button for the target patterns field should be rendered inside the input field as a React node. The current "include/exclude" button on the right-hand side should be replaced with a simple "Add" button. Users will first select whether to include or exclude a pattern within the input field and then click the "Add" button to add the pattern.
> 
> cc: @abhinavrastogi-harness

agreed. we've gone back-and-forth on this multiple times now. we had taken a call to change the UX to remove the split button here. Let's not bring it back. The designs need to be updated here to reflect the product decision.

.eslintrc.json
praneshg239: Agree, we made the `tailwindcss/no-custom-classname` rule a `warn` for the time being. It will soon be updated to `error`, which will prevent adding duplicate styles that can already be achieved through Tailwind.

praneshg239: May I know for which scenario we want to disable `tailwindcss/enforces-negative-arbitrary-values` rule?

.eslintrc.json
```diff
@@ -45,7 +45,7 @@
     "react/react-in-jsx-scope": "off",
 
     "jsx-a11y/click-events-have-key-events": "warn",
-    "jsx-a11y/no-autofocus": "warn",
+    "jsx-a11y/no-autofocus": "off",
```
praneshg239, line 48: disabling `jsx-a11y/no-autofocus` is acceptable. In some scenarios, setting autofocus is necessary for better UX. For example, forms in modals or dialogs can focus on the first input when they appear.

.eslintrc.json
3em: @abhinavrastogi-harness @praneshg239  

tailwindcss/no-custom-classname is designed to prevent the use of custom class names, but we do use them in some places. Therefore, I disabled this rule as it’s not needed for the project.

Regarding tailwindcss/enforces-negative-arbitrary-values, I ended up removing this rule. We had some custom negative values, but I fixed their format in the classes in the new commit, so we don’t need to add this rule anymore.

3em:
> The include/exclude button for the target patterns field should be rendered inside the input field as a React node. The current "include/exclude" button on the right-hand side should be replaced with a simple "Add" button. Users will first select whether to include or exclude a pattern within the input field and then click the "Add" button to add the pattern.
> 
> cc: @abhinavrastogi-harness

Here is the current new design (screenshot) implemented in this PR. In the button to the right of the input field, the user selects "Include" or "Exclude" and performs the corresponding action with the value entered in the input.

image

packages/ui/src/components/multi-select.tsx
3em: @abhinavrastogi-harness Screenshots of the Settings section and the Rules form have been added. What other screenshots should I include?

I will add these sections to the design system in the next PR, as I’m currently working on some changes to these sections after discussions with the designers.

packages/ui/src/components/text.tsx
3em: @abhinavrastogi-harness Actually, there was a discussion a while ago about completely moving away from this component and eventually removing it, as it can essentially be replaced with just tags and classes. However, since a large number of developers are still using it, I added the missing size to it for now.

3em:
> agreed. we've gone back-and-forth on this multiple times now. we had taken a call to change the UX to remove the split button here. Let's not bring it back. The designs need to be updated here to reflect the product decision.

@abhinavrastogi-harness 
For this issue, I’ll create a separate thread in Slack where our managers and designers can discuss it with you and adjust the design as needed. However, for now, can we keep the current implementation in this PR so it doesn’t get delayed? Once the new design is approved, our team will implement the necessary changes to this button.

praneshg239 (approved):

praneshg239:
Approved eslint changes.

PR #652, opened 2025-01-10
fix: Ensure App Shell renders for all applicable routes 

vardanbansal-harness:
Currently, routes which are sibling of `/` do not render with App shell, which includes side nav and breadcrumbs. This PR fixes it.

**Sample**:

**Before**:

image

**After**:

image


sans-harness (approved):
LGTM

PR #659, opened 2025-01-13
Create, edit delete and list labels

sans-harness:
- Adds the create-edit-delete flow on the labels page in a project scope. 

https://github.com/user-attachments/assets/e839aea2-7863-4492-9833-be0bfc54397f

### DIFF
--- a/packages/ui/locales/en/component.json
+++ b/packages/ui/locales/en/component.json
@@ -111,9 +111,7 @@
     "input": "Search files...",
     "noFile": "No file found."
   },
-  "There are no labels in this project yet": {
-    " Create a new label to get started": ""
-  },
+
   "layout": {
     "table": "Table",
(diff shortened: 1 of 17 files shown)


packages/ui/locales/en/component.json
```diff
@@ -111,6 +111,9 @@
     "input": "Search files...",
     "noFile": "No file found."
   },
+  "There are no labels in this project yet": {
```
vardanbansal-harness, line 114: isn't this supposed to be a `key`? Can it have spaces?

vardanbansal-harness (approved):

packages/ui/locales/en/component.json
sans-harness: added by mistake 😅, removed.

PR #674, opened 2025-01-15
Add base Starlight app

knagurski:
This PR adds a new Astro Starlight app to the `apps` directory to begin work on a Starlight-powered docs experience.


c-joetaylor-harness (approved):
On the assumption that the lint check fail is non-blocking, approved.

PR #684, opened 2025-01-16
Add search pagination to labels

sans-harness:
- combines label stores and dedupes. Implements architecture discussed in standup on 1/15. 
- Adds pagination, empty states, and pagination. 

https://github.com/user-attachments/assets/2a4cae7d-dd04-41e0-a47f-feb99860c72e


apps/gitness/src/pages-v2/project/project-labels-list-container.tsx
```diff
@@ -75,11 +100,34 @@ export const ProjectLabelsList = () => {
...
+  // temporary solution to fetch all label values - should be reworked once BE changes are done
+  useEffect(() => {
+    async function fetchAllLabelValues(storeLabels: ILabelType[]) {
+      if (!space_ref || !storeLabels) return
+
+      const valuesByKey: Record<string, any> = {}
+
+      for (const label of storeLabels) {
+        try {
+          const response = await fetch(`/api/v1/spaces/${space_ref}/labels/${label.key}/values`)
```
abhinavrastogi-harness, line 117: we can do a small optimisation: check the count of values first, and only call API if count > 0.

apps/gitness/src/pages-v2/project/stores/labels-store.tsx
```diff
@@ -1,12 +1,30 @@
 import { create } from 'zustand'
 
-import { ILabelsStore, ILabelType } from '@harnessio/ui/views'
+import { ILabelsStore, ILabelType, LabelValueType } from '@harnessio/ui/views'
 
 export const useLabelsStore = create<ILabelsStore>(set => ({
-  labels: [],
+  spaceLabels: [],
+  repoLabels: [],
```
abhinavrastogi-harness, line 7: why do we still need to maintain two separate arrays?
if the schema of the repo labels and space labels is the same, we can simplify this significantly by just maintaining one list, and overwriting if anything changes.

packages/ui/src/views/project/project-labels/project-labels-list.tsx
```diff
@@ -23,29 +47,51 @@ export const ProjectLabelsListView: React.FC<ProjectLabelPageProps> = ({
...
-        <LabelsListView
-          labels={labels}
-          createdIn={createdIn}
-          handleDeleteLabel={handleDeleteLabel}
-          handleEditLabel={handleEditLabel}
-          useTranslationStore={useTranslationStore}
+        {isLoadingSpaceLabels ? (
+          <SkeletonList />
+        ) : (
+          <LabelsListView
```
abhinavrastogi-harness, line 74: i'm curious why the SkeletonList is outside the ListView, but the NoData state is inside..

abhinavrastogi-harness (approved):
I think some parts of this can be simplified further, but should not be a blocker

Issue #694, opened 2025-01-16
Add Accordion component docs


knagurski:
Document and refactor the `Accordion` component and subcomponents.

PR #737, opened 2025-01-20
feat: Namespace `Accordion` components and add Accordion docs

knagurski:
This PR refactors `Accordion` related components into a common namespace and adds initial docs.

fixes: #694


apps/portal/src/components/docs-page/example.tsx
```diff
@@ -18,9 +18,8 @@ const Example: FC<ExampleProps> = ({ code, scope }) => {
     <div className="my-12 flex flex-col justify-start gap-0 pb-3">
       <LiveProvider code={code} scope={scopeWithLayout}>
         <h2>Preview</h2>
-        <div className="border-borders-4 bg-background-4 dark-std-std mb-0 flex rounded-lg border px-6 py-3 shadow-md">
-          <LivePreview />
-        </div>
+        <LivePreview className="border-borders-4 bg-background-4 dark-std-std mb-0 rounded-lg border p-6 shadow-md" />
```
knagurski, line 21: Moved classes here as the `LivePreview` component adds it's own `div` around things, so better to style that instead.

c-joetaylor-harness (approved):

PR #738, opened 2025-01-20
feat: [pipe-24076]: add link to commit details page

cjlee01:
clicking commit title brings you to the commit details page of that commit 

https://github.com/user-attachments/assets/7aa0f642-51c1-4cfe-a8de-fe9a03e094e7

### DIFF
--- a/packages/ui/src/components/commit-copy-actions.tsx
+++ b/packages/ui/src/components/commit-copy-actions.tsx
@@ -6,17 +6,10 @@ import copy from 'clipboard-copy'
 
 export const CommitCopyActions = ({
   sha,
-  inCompare = false,
-  inPr = false,
-  inDetails = false,
-  commitPath
+  toCommitDetails
 }: {
   sha: string
-  inCompare?: boolean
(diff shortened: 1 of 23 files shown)


packages/ui/src/components/commit-copy-actions.tsx
```diff
@@ -20,7 +35,41 @@ export const CommitCopyActions = ({ sha }: { sha: string }) => {
...
-        <span className="text-13 text-foreground-3">{sha.substring(0, 7)}</span>
+        <span
+          role="button"
+          tabIndex={0}
+          onClick={() => {
+            let path
+            if (commitPath) {
+              path = `${commitPath}/${sha}`
+            } else if (inCompare || inDetails) {
+              path = `../../commits/${sha}`
```
shaurya-harness, line 46: we are trying to avoid relative paths, maybe we can create a route for this as well?

cjlee01: fixed

PR #747, opened 2025-01-21
Portal getting started

c-joetaylor-harness:
image


knagurski (approved):
We should check out how we can use SVG for the favicon at some point, but not important at the moment.

PR #760, opened 2025-01-21
React 17

vivek-harness:
This PR is for downgrading from React 18 to React 17 to maintain consistency with NGUI and avoiding runtime issues.

https://github.com/user-attachments/assets/84987b12-0cf8-4a0e-9c88-ea0924555cbb


apps/gitness/src/hooks/useDebouncedQueryState.ts
```diff
@@ -19,18 +19,24 @@ type UseDebouncedQueryStateParams = [string] | [string, string] | [string, UseDe
  * useDebouncedQueryState('key', { defaultValue: 'value', delay: 500 })
  */
 export function useDebouncedQueryState(...args: UseDebouncedQueryStateParams): [string, (value: string) => void] {
+  const [searchParams, setSearchParams] = useSearchParams()
   const [key, options] = args
   const defaultValue = typeof options === 'string' ? options : (options?.defaultValue ?? '')
   const delay = typeof options === 'object' && options?.delay !== undefined ? options.delay : 300
 
-  const [query, setQuery] = useQueryState(key, { defaultValue })
+  const query = searchParams.get(key) || defaultValue
```
abhinavrastogi-harness, line 27: we can probably create a custom hook which has the same API as `nuqs`, but uses `useSearchParams` internally.

vivek-harness: Route communication is no longer needed between NGUI and MFE, probably the same react and react-dom versions are taking care of it.

Yes, we may create a custom hook as a replacement to nuqs once all other changes for React 17 are in place.

vardanbansal-harness (approved):

PR #766, opened 2025-01-21
fix: Add "asLink" prop to allow rendering non-link breadcrumbs

vardanbansal-harness:
- This PR adds support for `asLink` prop that handles rendering of non-link elements in Breadcrumbs, for e.g. Project Selector.
- Also fixes the flow where list of spaces if fetched if not available in App context while rendering the Project selector. Additionally, also removes duplicate api calls from Landing page.
- `<AppProvider /` now wraps around `<AppShell />` instead, since few routes outside of the App shell, such as `/signin` etc., do not need access to App context.

### DIFF
--- a/apps/gitness/src/components-v2/breadcrumbs/project-dropdown.tsx
+++ b/apps/gitness/src/components-v2/breadcrumbs/project-dropdown.tsx
@@ -1,5 +1,7 @@
+import { useEffect } from 'react'
 import { useNavigate, useParams } from 'react-router-dom'
 
+import { membershipSpaces, TypesSpace } from '@harnessio/code-service-client'
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -17,7 +19,24 @@ function ProjectDropdown(): JSX.Element {
   const routes = useRoutes()
   const { spaceId } = useParams<PathParams>()
   const navigate = useNavigate()
-  const { spaces } = useAppContext()
+  const { spaces, setSpaces } = useAppContext()
+
+  useEffect(() => {
+    if (!spaces?.length) {
+      membershipSpaces({
+        queryParams: { page: 1, limit: 100, sort: 'identifier', order: 'asc' }
+      })
+        .then(({ body: memberships }) => {
+          const spaceList = memberships.filter(item => item?.space).map(item => item.space as TypesSpace)
+          setSpaces(spaceList)
+        })
+        .catch(() => {
+          // Optionally handle error or show toast
+          navigate(routes.toSignIn())
+        })
+    }
+  }, [spaces, membershipSpaces, navigate])
+
   return (
     <DropdownMenu>
       <DropdownMenuTrigger className="flex items-center gap-x-1.5">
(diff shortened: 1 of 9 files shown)


apps/gitness/src/components-v2/breadcrumbs/project-dropdown.tsx
```diff
@@ -17,7 +19,24 @@ function ProjectDropdown(): JSX.Element {
...
+  const { spaces, setSpaces } = useAppContext()
+
+  useEffect(() => {
+    if (!spaces?.length) {
+      membershipSpaces({
+        queryParams: { page: 1, limit: 100, sort: 'identifier', order: 'asc' }
+      })
+        .then(({ body: memberships }) => {
+          const spaceList = memberships.filter(item => item?.space).map(item => item.space as TypesSpace)
+          setSpaces(spaceList)
```
abhinavrastogi-harness, line 31: doesn't this risk an infinite loop? you're setting state in a useEffect that triggers when the same state changes.
instead, we should move this to a different useEffect, or maybe to the AppContext if possible.

apps/gitness/src/components-v2/breadcrumbs/project-dropdown.tsx
```diff
@@ -17,7 +19,24 @@ function ProjectDropdown(): JSX.Element {
...
+        })
+        .catch(() => {
```
abhinavrastogi-harness, line 33: all errors should not trigger a navigation to signin. i think that is already handled via global 401 handling.
here we can simple disable the dropdown maybe for now.

vardanbansal-harness: Done.

apps/gitness/src/components-v2/breadcrumbs/project-dropdown.tsx
vardanbansal-harness: Yes definitely, my bad. Moved it to AppContext.

apps/gitness/src/framework/context/AppContext.tsx
```diff
@@ -26,6 +26,19 @@ export const AppProvider: React.FC<{ children: ReactNode }> = ({ children }) =>
   const [spaces, setSpaces] = useState<TypesSpace[]>([])
   const [currentUser, setCurrentUser] = useLocalStorage<TypesUser>('currentUser', {})
 
+  useEffect(() => {
+    membershipSpaces({
+      queryParams: { page: 1, limit: 100, sort: 'identifier', order: 'asc' }
+    })
+      .then(({ body: memberships }) => {
+        const spaceList = memberships.filter(item => item?.space).map(item => item.space as TypesSpace)
+        setSpaces(spaceList)
```
sans-harness, line 35: we might also need support for pagination here

sans-harness (approved):

apps/gitness/src/framework/context/AppContext.tsx
vardanbansal-harness: Definitely, will discuss how to handle this.

PR #804, opened 2025-01-22
Pipeline graph documentation

srdjan-harness:


packages/views/src/components/form-inputs/common/InputLabel.tsx
abhinavrastogi-harness, line 1: we should remove this file if possible..
we want to remove the entire `canary` and `views` packages.
even gitness depends on these via the run pipeline form today.

packages/views/src/components/form-inputs/common/InputLabel.tsx
```diff
@@ -7,21 +7,22 @@ export interface InputLabelProps {
 }
 
 function InputLabel(props: InputLabelProps): JSX.Element {
-  const { label, description, required } = props
```
harish-viswa, line 10: Test

PR #811, opened 2025-01-23
Filters package

abhinavrastogi-harness:


praneshg239 (approved):

packages/filters/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "files": [
+    "dist"
+  ],
+  "exports": {
+    ".": {
+      "import": "./dist/index.js"
+    }
+  },
+  "peerDependencies": {
+    "react": ">=17.0.0 <19.0.0",
```
praneshg239, line 26: it should be `^17.0.2`

packages/filters/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "react-dom": ">=17.0.0 <19.0.0",
```
praneshg239, line 27: same here

PR #837, opened 2025-01-27
Enhancements on Filters

harish-viswa:
1. Added support for defaultValue
2. Added resetFilters support inside dropdown

### DIFF
--- a/packages/filters/src/Filter.tsx
+++ b/packages/filters/src/Filter.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+
 import { useFiltersContext } from './Filters'
 import { defaultStringParser } from './parsers'
 
@@ -12,7 +13,7 @@ export interface FilterProps<T, K extends keyof T> {
   children: (props: {
     onChange: (value: T[K]) => void
     value?: Parser<T[K]> extends undefined ? string : T[K]
-    removeFilter: (filterKey: K) => void
+    removeFilter: (filterKey?: K) => void
   }) => React.ReactNode
   parser?: Parser<T[K]>
   sticky?: boolean
@@ -33,6 +34,12 @@ const Filter = <T, K extends keyof T>({
     updateFilter(filterKey as string, serializedValue, value)
   }
 
+  // If no filter key is provided,
+  // filterKey provided to component will be used
+  const _removeFilter = (fkey?: K) => {
+    removeFilter(fkey ?? filterKey)
+  }
+
   // Retrieves the raw and parsed filter value
   const rawValue = getFilterValue(filterKey as string)
   const parsedValue = rawValue as T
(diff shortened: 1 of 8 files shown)


packages/filters/src/Filter.tsx
```diff
@@ -33,14 +34,24 @@ const Filter = <T, K extends keyof T>({
...
   const parsedValue = rawValue as T
 
   // Render the children with the injected props
   return (
     <div id="filter" className={className}>
-      {children({ onChange: handleChange, value: parsedValue as T[K], removeFilter })}
+      {children({
+        onChange: handleChange,
+        value: parsedValue as T[K],
+        removeFilter: _removeFilter
```
avinashmadhwani02, line 53: Just add a inline function here, instead of creating _removeFilter. 
_removeFilter is not readable. You can rename the method too -- wrappedRemoveFilter

```suggestion
        removeFilter: (key: k) => removeFilter(key || filterKey)
```

packages/filters/src/Filters.tsx
```diff
@@ -99,11 +100,22 @@ const Filters = forwardRef(function Filters<T extends Record<string, unknown>>(
     const config = {} as Record<FilterKeys, FilterConfig>
 
     for (const key in initialFiltersConfig) {
-      const isSticky = allFiltersSticky ? true : initialFiltersConfig[key].isSticky
+      const { defaultValue, parser, isSticky: _isSticky } = initialFiltersConfig[key]
```
avinashmadhwani02, line 103: avoid naming with underscore.

packages/filters/src/Filters.tsx
```diff
@@ -139,9 +151,15 @@ const Filters = forwardRef(function Filters<T extends Record<string, unknown>>(
...
-      Object.keys(map).filter(filter => map[filter as FilterKeys].state !== FilterStatus.HIDDEN) as FilterKeys[]
-    )
+    const newFiltersOrder = Object.keys(map).filter(
+      filter => map[filter as FilterKeys].state !== FilterStatus.HIDDEN
+    ) as FilterKeys[]
+    setFiltersOrder(newFiltersOrder)
+
+    const query = createQueryString(newFiltersOrder, map)
+    debug('Updating URL with query: %s', query)
+    updateURL(new URLSearchParams(query))
```
avinashmadhwani02, line 161: we should add a check - if there's no query, updateURL shouldn't be called

packages/filters/src/Filters.tsx
```diff
@@ -211,18 +229,30 @@ const Filters = forwardRef(function Filters<T extends Record<string, unknown>>(
     Object.keys(updatedFiltersMap).forEach(key => {
       const isSticky = filtersConfig[key as FilterKeys]?.isSticky
       const defaultValue = filtersConfig[key as FilterKeys]?.defaultValue
+      const parser = filtersConfig[key as FilterKeys]?.parser
+
+      const serializedDefaultValue = defaultValue ?? parser?.serialize(defaultValue)
```
avinashmadhwani02, line 234: do you think we should save serialized defaultValue in filtersConfig? parsing again will not be required.

packages/filters/src/Filter.tsx
harish-viswa: I will rename the method, would be more readable as a separate function I guess

avinashmadhwani02 (approved):

PR #843, opened 2025-01-27
feat: add APIs and navigations

sans-harness:
- Currently the `file-create` view breaks for an empty repo. (see attached video) Lets pick this up as a seperate PR. 

cc: @shaurya-harness 

https://github.com/user-attachments/assets/c986fe2b-d9b3-452f-a652-abec0e8647c0

### DIFF
--- a/packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
+++ b/packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
@@ -21,6 +21,7 @@ interface RepoEmptyViewProps {
   sshUrl: string
   gitRef: string
   handleCreateToken: () => void
+  toProfileKeys?: () => string
 }
 
 export const RepoEmptyView: React.FC<RepoEmptyViewProps> = ({
@@ -29,7 +30,8 @@ export const RepoEmptyView: React.FC<RepoEmptyViewProps> = ({
   httpUrl,
   sshUrl,
   gitRef,
-  handleCreateToken
+  handleCreateToken,
+  toProfileKeys
 }) => {
   const getInitialCommitMarkdown = () => {
     return `
@@ -84,7 +86,7 @@ git push -u origin main
             </ButtonGroup>
             <p className="mt-2">
               You can also manage your git credential{' '}
-              <StyledLink to="/profile-settings/keys" relative="path">
+              <StyledLink to={toProfileKeys?.() ?? ''} relative="path">
                 here
               </StyledLink>
             </p>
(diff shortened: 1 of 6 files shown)


packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -56,7 +66,10 @@ git push -u origin main
           iconName="no-repository"
           title="This repository is empty"
           description={['We recommend every repository include a', 'README, LICENSE, and .gitignore.']}
-          primaryButton={{ label: 'New file' }}
+          primaryButton={{
+            label: 'New file',
+            to: `${projName ? `/${projName}` : ''}/repos/${repoName}/code/new/${gitRef}/~/`
```
abhinavrastogi-harness, line 71: can we pass in the routes function instead of hardcoding a path in the view?

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -70,11 +83,11 @@ git push -u origin main
           <Input label="SSH" value={sshUrl} readOnly rightElement={<CopyButton name={sshUrl} />} />
           <ControlGroup>
             <ButtonGroup>
-              <Button>Generate Clone Credentials</Button>
+              <Button onClick={handleCreateToken}>Generate Clone Credentials</Button>
             </ButtonGroup>
             <p className="mt-2">
               You can also manage your git credential{' '}
-              <StyledLink to="/" relative="path">
+              <StyledLink to="/profile-settings/keys" relative="path">
```
abhinavrastogi-harness, line 90: same here

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -92,7 +105,7 @@ git push -u origin main
             <MarkdownViewer source={getExistingRepoMarkdown()} />
             <p>
               You might need to{' '}
-              <StyledLink to="/" relative="path">
+              <StyledLink to="/profile-settings/keys" relative="path">
```
abhinavrastogi-harness, line 108: same

abhinavrastogi-harness (commented):
cc @vardanbansal-harness please help with route def adoption if needed

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
sans-harness: @vardanbansal-harness is it possible to define a route-constant for code since it matches all paths ? 

```
path: '*',
element: <RepoCode />
```

vardanbansal-harness: Yeah we should be able to use `routes.toRepoFiles()` here but that will cover path only upto `../code`, and rest you need to append after it.

abhinavrastogi-harness (approved):
we can take up the remaining route change in a follow-up PR if needed

PR #845, opened 2025-01-27
feat: Add Document title to pages

vardanbansal-harness:
- Adds a new `usePageTitle` hook that matches the current route and renders the Document(Page) title for it. Currently, added only for Repository pages.

https://github.com/user-attachments/assets/07f04598-eb01-45a8-90b7-fa00941d9a41

- Performs some code reorg/cleanup.


apps/gitness/src/routes.tsx
```diff
@@ -51,23 +52,6 @@ import { UserManagementPageContainer } from './pages-v2/user-management/user-man
 import { CreateWebhookContainer } from './pages-v2/webhooks/create-webhook-container'
 import WebhookListPage from './pages-v2/webhooks/webhook-list'
 
-export const extractRedirectRouteObjects = (routes: CustomRouteObject[]): CustomRouteObject[] => {
```
vardanbansal-harness, line 54: Moved to `apps/gitness/src/framework/routing/utils.ts`

apps/gitness/src/framework/routing/types.ts
```diff
@@ -99,6 +99,11 @@ export interface CustomHandle {
    * Associated route name for the breadcrumb.
    */
   routeName?: string
+
+  /**
+   * Updates the document title based on route parameters.
+   */
+  documentTitle?: (params: Params<string>) => string
```
abhinavrastogi-harness, line 106: can we support an API like this?

```suggestion
  documentTitle?: string | (params: Params<string>) => string
```

this would allow the user to pass a hardcoded string where params are not needed.

apps/gitness/src/framework/routing/utils.ts
```diff
@@ -77,3 +77,20 @@ export const getRouteMapping = ({
   const routeEntries = generateRouteEntries({ routes, parentPath, parentName })
   return generateRouteNameToPathFunctions(routeEntries)
 }
+
+export const extractRedirectRouteObjects = (routes: CustomRouteObject[]): CustomRouteObject[] => {
```
abhinavrastogi-harness, line 81: this is ripe for adding some unit tests 😅

apps/gitness/src/routes.tsx
```diff
@@ -107,7 +97,8 @@ export const repoRoutes: CustomRouteObject[] = [
             element: <RepoSummaryPage />,
             handle: {
               breadcrumb: () => <Text>Summary</Text>,
-              routeName: RouteConstants.toRepoSummary
+              routeName: RouteConstants.toRepoSummary,
+              documentTitle: ({ repoId }: { repoId: string }) => `Repository | ${repoId} | Summary`
```
abhinavrastogi-harness, line 101: if every page needs to provide the complete path, it seems to defeat the purpose. could we have something like this:

```
{
  path: 'repos',
  handle: {
    title: 'Repository'
  },
  children: [
   {
      path: ':id',
      handle: {
        title: (id) => id
      }
    } 
  ]
}
```

and we could combine all segments upto the matching path dynamically.

apps/gitness/src/framework/routing/types.ts
vardanbansal-harness: Done.

apps/gitness/src/routes.tsx
vardanbansal-harness: Done.

apps/gitness/src/framework/context/DocumentTitleProvider.tsx
```diff
@@ -0,0 +1,6 @@
+import useDocumentTitle from '../hooks/useDocumentTitle'
+
+export const DocumentTitleProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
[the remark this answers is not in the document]
vardanbansal-harness: Done.

apps/gitness/src/framework/routing/utils.ts
vardanbansal-harness: Definitely

apps/gitness/src/framework/hooks/useDocumentTitle.ts
```diff
@@ -0,0 +1,20 @@
...
+import { CustomHandle } from '../routing/types'
+
+const useDocumentTitle = () => {
+  const { t } = useTranslationStore()
+  const matches = useMatches()
+
+  useEffect(() => {
+    const currentMatch = matches.at(-1) // Get the last match
+    if (currentMatch) {
+      const { documentTitle } = (currentMatch?.handle || {}) as CustomHandle
```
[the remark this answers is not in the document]
vardanbansal-harness: Agreed, although with the new implementation, I've implemented it a slightly different way.

abhinavrastogi-harness (approved):

PR #861, opened 2025-01-29
Added useAPIPath hook to check for mfe condition

vivek-harness:
This hook adds required prefix to the API Path for MFE including window.apiURL along with /code and also adds routingId as a query param.

Before:
const rawURL = `${isMFE ? '/code' : ''}/api/v1/repos/${repoRef}/raw/${fullResourcePath}?git_ref=${fullGitRef}`

After:
const rawURL = useAPIPath(`/api/v1/repos/${repoRef}/raw/${fullResourcePath}?git_ref=${fullGitRef}`)

### DIFF
--- a/apps/gitness/src/hooks/useAPIPath.ts
+++ b/apps/gitness/src/hooks/useAPIPath.ts
@@ -1,9 +1,10 @@
-import { useIsMFE } from '../framework/hooks/useIsMFE'
+import { useMFEContext } from '../framework/hooks/useMFEContext'
 
 export function useAPIPath(path: string) {
-  const isMFE = useIsMFE()
+  const mfeContext = useMFEContext()
+  const isMFE = mfeContext.renderUrl !== ''
   if (isMFE) {
-    return `/code${path}`
+    return `${window.apiUrl || ''}/code${path}${path.includes('?') ? '&' : '?'}routingId=${mfeContext.scope.accountId}`
   }
   return path
 }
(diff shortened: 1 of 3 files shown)


apps/gitness/src/hooks/useAPIPath.ts
```diff
@@ -0,0 +1,9 @@
+import { useIsMFE } from '../framework/hooks/useIsMFE'
+
+export function useAPIPath(path: string) {
+  const isMFE = useIsMFE()
+  if (isMFE) {
+    return `/code${path}`
```
praneshg239, line 6: we need to add `${window.apiUrl || ''}` as a prefix.

apps/gitness/src/hooks/useAPIPath.ts
```diff
@@ -0,0 +1,10 @@
+import { useMFEContext } from '../framework/hooks/useMFEContext'
+
+export function useAPIPath(path: string) {
+  const mfeContext = useMFEContext()
+  const isMFE = mfeContext.renderUrl !== ''
+  if (isMFE) {
+    return `${window.apiUrl || ''}/code${path}${path.includes('?') ? '&' : '?'}routingId=${mfeContext.scope.accountId}`
+  }
+  return path
```
abhinavrastogi-harness, line 9: i think even if we are not in MFE, we should still respect `window.apiUrl`. but shouldn't be a blocker.

abhinavrastogi-harness (approved):

PR #867, opened 2025-01-29
feat: Update style tokens for light-std-std and introduce dark-std-high

### LINKED ISSUE #853: Update styles
New styles are available and should be implemented.

knagurski:
fixes #853


abhinavrastogi-harness (approved):
looks good.
but didn't we decide to keep the new colors as `dark-std-std`, and move the current theme to a "dim" or low-contrast mode instead?

knagurski:
> looks good. but didn't we decide to keep the new colors as `dark-std-std`, and move the current theme to a "dim" or low-contrast mode instead?

Good point. Changed to `dark-std-low` 👍🏻

PR #897, opened 2025-01-30
chore: Update tokens and ensure correct files are included in the ui package output

### LINKED ISSUE #896: Update styles
Update to new tokens

knagurski:
Fixes #896


abhinavrastogi-harness (commented):
why are we reverting back to hardcoded colors instead of using the variables?

knagurski:
> why are we reverting back to hardcoded colors instead of using the variables?

@abhinavrastogi-harness These changes are experimental right now (converted to a draft) and specific to the diff viewer. Anastasia is trying to dial it in at the moment, so we might not have "official" colours to cover what we need.

PR #917, opened 2025-01-31
Pipeline studio - integration updates

srdjan-harness:
- Add button
- onErrorChange

### DIFF
--- a/packages/pipeline-graph/src/render/render-node-content.tsx
+++ b/packages/pipeline-graph/src/render/render-node-content.tsx
@@ -1,4 +1,5 @@
 import { useGraphContext } from '../context/graph-provider'
+import { ContainerNodeType } from '../types/container-node'
 import { ContainerNode } from '../types/nodes'
 import {
   AnyNodeInternal,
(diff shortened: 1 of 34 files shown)


packages/pipeline-graph/src/render/render-node-content.tsx
```diff
@@ -11,28 +12,49 @@ export function RenderNodeContent(props: {
...
+  const { node, children, collapsed, isFirst, isLast, parentNodeType } = props
   const { nodes } = useGraphContext()
 
   const nodeContent = nodes[node.type]
 
   switch (nodeContent.containerType) {
     case ContainerNode.leaf:
       return (
-        <nodeContent.component node={node as LeafNodeInternalType<{}>} collapsed={collapsed}>
+        <nodeContent.component
```
abhinavrastogi-harness, line 27: not a blocker, but all react components should start with capital letters like `NodeContent.component`

packages/ui/src/views/pipeline-edit/components/graph-implementation/types/problem-type.ts
```diff
@@ -0,0 +1,11 @@
+export type ProblemSeverity = 'error' | 'warning' | 'info'
+
+export interface Problem<T = unknown> {
+  severity: ProblemSeverity
+  message: string
+  position: {
```
abhinavrastogi-harness, line 6: position should be optional, so that we can add non-yaml problems to the Problems component later as well

abhinavrastogi-harness (approved):

PR #901, opened 2025-01-31
feat: Adding new theme and language selectors

vardanbansal-harness:
Theme Selector: 

https://github.com/user-attachments/assets/1ebc000c-bcb8-4aac-a1cd-1b8f86386a3b

Language Selector:

https://github.com/user-attachments/assets/0e2f7e4f-9646-44cf-a735-a9d34def6666

PR #921, opened 2025-01-31
chore: PR#901 follow up

vardanbansal-harness:
Follow up changes from PR https://github.com/harness/canary/pull/901.


abhinavrastogi-harness:
there is another follow-up PR here: https://github.com/harness/canary/pull/934
can we close this if we're doing the other one? looks like there is a design change as well

sans-harness:
can we close this? cc: @vardanbansal-harness

PR #924, opened 2025-02-02
New component: Chat

c-joetaylor-harness:
# Chat Component

The `Chat` component provides a modular and accessible chat UI, consisting of various subcomponents for structure, messages, typing indicators, separators, and input fields.

## Installation

Ensure you have the necessary dependencies installed:

```sh
npm install @harnessio/ui
```

or if using pnpm:

```sh
pnpm add @harnessio/ui
```

---

### Basic Example

```jsx
import { Chat } from "@harnessio/ui/components";

export default function ChatExample() {
  return (
    <Chat.Root>
      <Chat.Body>
        <Chat.Separator title="Today" />
        <Chat.Message avatar={<Avatar />} time="10:30 AM">
          Hello! How can I help you?
        </Chat.Message>
        <Chat.Typing avatar={<Avatar />} />
      </Chat.Body>
      <Chat.Footer>
        <Chat.Input />
      </Chat.Footer>
    </Chat.Root>
  );
}

function Avatar() {
  return <div className="w-5 h-5 rounded-md"></div>;
}
```

---

### `Chat.Root`
The root container that wraps the entire chat UI.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `children` | `ReactNode` | The content inside the chat container. |

### `Chat.Body`
A flexible body section that contains messages and other elements.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `children` | `ReactNode` | The chat messages and elements inside the body. |

### `Chat.Footer`
The footer section that typically contains the input field.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `children` | `ReactNode` | The elements inside the chat footer. |

### `Chat.Message`
A single message bubble, supporting avatars, timestamps, and actions.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `self?`   | `boolean`  | Whether the message is sent by the current user. |
| `time?`   | `string`   | Timestamp of the message. |
| `avatar?` | `ReactNode` | The avatar of the sender. |
| `actions?` | `ReactNode` | Action buttons (e.g., reactions, reply). |
| `children` | `ReactNode` | The message content. |

### `Chat.Typing`
A typing indicator with animated dots.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `avatar?` | `ReactNode` | The avatar of the user typing. |

### `Chat.Separator`
A separator, typically used for displaying date breaks.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `title`   | `string`   | The text displayed in the separator. |

### `Chat.Input`
An input field for sending messages.

#### Props
| Prop      | Type       | Description                |
|-----------|-----------|----------------------------|
| `value?` | `string` | The current input value. |
| `inputRef?` | `RefObject<HTMLInputElement>` | Ref for the input field. |
| `onChange?` | `(e: ChangeEvent<HTMLInputElement>) => void` | Handler for input changes. |
| `onKeyDown?` | `(e: KeyboardEvent<HTMLInputElement>) => void` | Handler for key press events. |
| `onSend?` | `() => void` | Callback when the send button is clicked. |
| `placeholder?` | `string` | Placeholder text for the input. Defaults to `Type a message...`. |
| `disabled?` | `boolean` | Whether the input is disabled. |
| `sendIcon?` | `ReactNode` | Custom send button icon. Defaults to a chevron-up icon. |

---

### Handling Input and Sending Messages

```jsx
import { useState, useRef } from "react";
import { Chat } from "@harnessio/ui/components";

export default function ChatExample() {
  const [messages, setMessages] = useState([
    { id: 1, text: "Hello!", self: false, time: "10:00 AM" },
  ]);
  const [inputValue, setInputValue] = useState("");
  const inputRef = useRef(null);

  const sendMessage = () => {
    if (!inputValue.trim()) return;
    setMessages([...messages, { id: messages.length + 1, text: inputValue, self: true, time: "10:05 AM" }]);
    setInputValue("");
  };

  return (
    <Chat.Root>
      <Chat.Body>
        {messages.map((msg) => (
          <Chat.Message key={msg.id} self={msg.self} time={msg.time}>
            {msg.text}
          </Chat.Message>
        ))}
      </Chat.Body>
      <Chat.Footer>
        <Chat.Input
          value={inputValue}
          inputRef={inputRef}
          onChange={(e) => setInputValue(e.target.value)}
          onSend={sendMessage}
        />
      </Chat.Footer>
    </Chat.Root>
  );
}
```

---

## Notes
- The `Chat.Message` component supports optional avatars and actions.
- The `Chat.Typing` component can be used to indicate a user is typing.
- The input component (`Chat.Input`) has built-in event handlers for better flexibility.

This component is highly customizable and can be extended based on your needs.

## Screenshot
image

### DIFF
--- a/packages/ui/src/components/chat.tsx
+++ b/packages/ui/src/components/chat.tsx
@@ -1,0 +1,149 @@
+import { ChangeEvent, KeyboardEvent, ReactNode, RefObject } from 'react'
+
+import { Button, Icon, Input, Text } from '@/components'
+import { cn } from '@/utils/cn'
+
+// Root Container
+const Root: React.FC<{ children: ReactNode }> = ({ children }) => {
+  return <div className="w-full h-full overflow-y-auto flex flex-col">{children}</div>
+}
+
+// Body
+const Body: React.FC<{ children: ReactNode }> = ({ children }) => {
+  return <div className="flex-1 flex flex-col gap-6 mb-3 p-6">{children}</div>
+}
+
+// Footer
+const Footer: React.FC<{ children: ReactNode }> = ({ children }) => {
+  return <div className="sticky bottom-0 p-6 pt-0 bg-background">{children}</div>
+}
+
+// Message Component
+interface MessageProps {
+  self?: boolean
+  time?: string
+  avatar?: ReactNode
+  actions?: ReactNode
+  children: ReactNode
+}
+
+const Message: React.FC<MessageProps> = ({ self, time, avatar, actions, children }) => {
+  return (
+    <div
+      className={cn('flex gap-x-3 content-center items-start', {
+        'place-content-end': self
+      })}
+    >
+      <div className="mt-0.5">{!self && avatar}</div>
+      <div
+        className={cn('flex flex-col gap-1.5', {
+          'w-[85%] items-end': self,
+          'w-full': !self
+        })}
+      >
+        <div
+          className={cn('text-foreground-3 leading-relaxed', {
+            'px-3 py-2 bg-background-8 rounded-lg': self
+          })}
+        >
+          {children}
+        </div>
+        <div className="flex gap-3 items-center justify-between mt-1">
+          <div>{actions && <div className="flex gap-1 items-center justify-start">{actions}</div>}</div>
+          {time && (
+            <Text size={1} className="text-primary opacity-20">
+              {time}
+            </Text>
+          )}
+        </div>
+      </div>
+    </div>
+  )
+}
+
+// Typing Indicator
+interface TypingProps {
+  avatar?: ReactNode
+}
+
+const Typing: React.FC<TypingProps> = ({ avatar }) => {
+  return (
+    <div className="flex items-center gap-3 mt-3">
+      {avatar}
+      <div className="flex gap-1 text-success text-lg font-medium" aria-live="polite">
+        <span className="dot animate-bounce">&middot;</span>
+        <span className="dot animate-bounce" style={{ animationDelay: '0.15s' }}>
+          &middot;
+        </span>
+        <span className="dot animate-bounce" style={{ animationDelay: '0.3s' }}>
+          &middot;
+        </span>
+      </div>
+    </div>
+  )
+}
+
+// Separator (For Date Breaks)
+interface SeparatorProps {
+  title: string
+}
+
+const Separator: React.FC<SeparatorProps> = ({ title }) => {
+  return <div className="text-center text-xs font-medium opacity-50">{title}</div>
+}
+
+// Input Field
+interface InputFieldProps {
+  value?: string
+  inputRef?: RefObject<HTMLInputElement>
+  onChange?: (e: ChangeEvent<HTMLInputElement>) => void
+  onKeyDown?: (e: KeyboardEvent<HTMLInputElement>) => void
+  onSend?: () => void
+  placeholder?: string
+  disabled?: boolean
+  sendIcon?: React.ReactNode
+}
+
+const InputField: React.FC<InputFieldProps> = ({
+  value = '',
+  inputRef,
+  onChange = () => {},
+  onKeyDown = () => {},
+  onSend = () => {},
+  placeholder = 'Type a message...',
+  disabled = false,
+  sendIcon = <Icon name="chevron-up" size={16} />
+}) => {
+  return (
+    <div className="sticky bottom-0 bg-background flex items-center gap-2">
+      <Input
+        ref={inputRef}
+        className="flex-1 pt-6 px-4 pb-16 rounded-lg focus:ring-0"
+        value={value}
+        onChange={onChange}
+        onKeyDown={onKeyDown}
+        placeholder={placeholder}
+        aria-label="Chat input"
+      />
+      <Button
+        onClick={onSend}
+        variant="outline"
+        size="icon"
+        disabled={disabled}
+        className="absolute right-3.5 bottom-3.5 z-10 w-6 h-6"
+      >
+        {sendIcon}
+      </Button>
+    </div>
+  )
+}
+
+export const Chat = {
+  Root,
+  Body,
+  Footer,
+  Message,
+  Typing,
+  Separator,
+  Input: InputField
+}


packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
+import { ChangeEvent, KeyboardEvent, ReactNode, RefObject } from 'react'
+
+import { Button, Icon, Input, Text } from '@/components'
+import { cn } from '@/utils/cn'
+
+// Root Container
+const Root: React.FC<{ children: ReactNode }> = ({ children }) => {
```
knagurski, line 7: `React` isn't imported in this file, you can add `FC` to the existing imports.
```suggestion
const Root: FC<{ children: ReactNode }> = ({ children }) => {
```

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+  return <div className="w-full h-full overflow-y-auto flex flex-col">{children}</div>
```
knagurski, line 8: Looks like the ESLint rules are kicking up a fuss about the order of classes here.
```suggestion
  return <div className="flex size-full flex-col overflow-y-auto">{children}</div>
```

Same with other ones below. Worth running the eslint --fix command.

knagurski: Screenshot 2025-02-03 at 11 18 02

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Footer
+const Footer: React.FC<{ children: ReactNode }> = ({ children }) => {
+  return <div className="sticky bottom-0 p-6 pt-0 bg-background">{children}</div>
+}
+
+// Message Component
+interface MessageProps {
+  self?: boolean
```
knagurski, line 23: What does `self` mean?

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+      >
+        <div
+          className={cn('text-foreground-3 leading-relaxed', {
+            'px-3 py-2 bg-background-8 rounded-lg': self
+          })}
+        >
+          {children}
+        </div>
+        <div className="flex gap-3 items-center justify-between mt-1">
+          <div>{actions && <div className="flex gap-1 items-center justify-start">{actions}</div>}</div>
```
knagurski, line 52: Is the extra `div` needed?
```suggestion
          {actions && <div className="flex gap-1 items-center justify-start">{actions}</div>}
```

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+// Typing Indicator
+interface TypingProps {
+  avatar?: ReactNode
+}
+
+const Typing: React.FC<TypingProps> = ({ avatar }) => {
+  return (
+    <div className="flex items-center gap-3 mt-3">
+      {avatar}
+      <div className="flex gap-1 text-success text-lg font-medium" aria-live="polite">
```
knagurski, line 73: Not sure if we want to use `aria-live` here. It will tell screen readers to announce, but there's no text to announce. Maybe best to add something the screen reader can announce.
```suggestion
      <div className="flex gap-1 text-success text-lg font-medium" aria-live="polite">
        <div className="sr-only">{nameOfPersonTyping} is typing</div>
```

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+  )
+}
+
+// Separator (For Date Breaks)
+interface SeparatorProps {
+  title: string
+}
+
+const Separator: React.FC<SeparatorProps> = ({ title }) => {
+  return <div className="text-center text-xs font-medium opacity-50">{title}</div>
```
knagurski, line 92: Is there something we can use for this text that's semantic? Is it a title of some kind? Maybe an heading tag of some kind?

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+      />
+      <Button
+        onClick={onSend}
+        variant="outline"
+        size="icon"
+        disabled={disabled}
+        className="absolute right-3.5 bottom-3.5 z-10 w-6 h-6"
+      >
+        {sendIcon}
+      </Button>
```
knagurski, line 136: Can we add something for a screenreader to see there? Perhaps an `aria-label`?

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+  inputRef,
+  onChange = () => {},
+  onKeyDown = () => {},
+  onSend = () => {},
+  placeholder = 'Type a message...',
+  disabled = false,
+  sendIcon = <Icon name="chevron-up" size={16} />
+}) => {
+  return (
+    <div className="sticky bottom-0 bg-background flex items-center gap-2">
```
knagurski, line 118: Might want to swap this out for a `form` here to allow the user to send chat messages by pressing enter instead of having to lose focus by swapping to the button. You can then move the `onSend` to the form as an `onSubmit` and make the button `type="submit"`. Remember to kill the event. (shame we're not on React 19 with its new form actions)

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+        disabled={disabled}
+        className="absolute right-3.5 bottom-3.5 z-10 w-6 h-6"
+      >
+        {sendIcon}
+      </Button>
+    </div>
+  )
+}
+
+export const Chat = {
```
knagurski, line 141: Looks like we're exporting, but not importing anywhere to be usable. Are we ok to add to the `index.ts` file in the `components` directory?

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+        >
+          {children}
+        </div>
+        <div className="flex gap-3 items-center justify-between mt-1">
+          <div>{actions && <div className="flex gap-1 items-center justify-start">{actions}</div>}</div>
+          {time && (
+            <Text size={1} className="text-primary opacity-20">
+              {time}
+            </Text>
+          )}
```
knagurski, line 57: Might be an idea to allow the consumer pass in a `Date` object or an ISO date/time string and then use the `<time>` tag instead of a `<span>`

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+const Message: React.FC<MessageProps> = ({ self, time, avatar, actions, children }) => {
+  return (
+    <div
+      className={cn('flex gap-x-3 content-center items-start', {
+        'place-content-end': self
+      })}
+    >
+      <div className="mt-0.5">{!self && avatar}</div>
```
knagurski, line 37: If it is `self` (or no `avatar` is passed), do we want the blank `div` as a spacer, or would it be better to not render anything?
```suggestion
      {!self && avatar && <div className="mt-0.5">{avatar}</div>}
```

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+  disabled?: boolean
+  sendIcon?: React.ReactNode
+}
+
+const InputField: React.FC<InputFieldProps> = ({
+  value = '',
+  inputRef,
+  onChange = () => {},
+  onKeyDown = () => {},
+  onSend = () => {},
```
knagurski, line 112: It might make things easier to wire up if this component handled its own state and just exposed `onTyping` and `onSend`. You could then keep the state of the input contained and only report back to the parent when the user starts/stops typing (using the `onTyping` callback) and when they wish to send the message (using the `onSend` callback with the message as the argument). Less flexible, but easier to consume and less to go wrong.

packages/ui/src/components/chat.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Message Component
+interface MessageProps {
+  self?: boolean
+  time?: string
+  avatar?: ReactNode
+  actions?: ReactNode
+  children: ReactNode
+}
```
knagurski, line 28: Instead of manually typing `children`, you can use `PropsWithChildren`.
```suggestion
interface MessageProps extends PropsWithChildren {
  self?: boolean
  time?: string
  avatar?: ReactNode
  actions?: ReactNode
}
```

packages/ui/src/components/chat.tsx
c-joetaylor-harness: It means the user's message, one's self.

packages/ui/src/components/chat.tsx
c-joetaylor-harness: I deliberately made this component super dumb, I think it's better that whatever consumes it handles all the logic/state/context. It might be that we build a reusable block for that in the future.

abhinavrastogi-harness:
is this PR superceded by this one? https://github.com/harness/canary/pull/937
can we close this if we're merging the other one?

abhinavrastogi-harness (approved):

PR #928, opened 2025-02-03
fix: Pull Request labels

3em:
- Added fetching of complete label data and their values
- Implemented functionality for adding a label, a label with a value, and a label with a custom label
- Enabled label display in the Pull Request list

https://github.com/user-attachments/assets/f16567de-2d9f-4b67-b5d2-bc299ed2693c


andrewgolovanov (approved):

packages/ui/src/views/repo/pull-request/components/pull-request-list.tsx
```diff
@@ -137,7 +138,9 @@ export const PullRequestList: FC<PullRequestListProps> = ({
       </StackedList.Root>
     )
   }
-  if (!filteredData?.length) return null
+
+  if (!filteredData?.length) return <></>
+
   return (
     <StackedList.Root>
       <StackedList.Item disableHover>
```
cjlee01, line 146: quick question: if there alot of labels in the pr list, does it wrap to the next line or how is that handled ?
@3em

3em: @cjlee01 That's a good question, I was wondering about it too. 

Right now, everything wraps, but without adjusting for the icon size, which is designed as an external element. 
This needs to be fixed, but the task is more about the overall refactoring of the PR list. I'll highlight this issue to the team so that we pay attention to it and make the necessary adjustments.

cjlee01: great thanks, in gitness we just wrap to the next line but maybe we can we iterate on this and make a better user experience @3em

PR #940, opened 2025-02-03
Summary page fixes

sans-harness:
Screenshot 2025-02-03 at 2 36 11 PM

### DIFF
--- a/packages/ui/src/views/repo/components/summary/summary.tsx
+++ b/packages/ui/src/views/repo/components/summary/summary.tsx
@@ -1,7 +1,7 @@
 import { useNavigate } from 'react-router-dom'
 
 import { Icon, Spacer, Table, TableBody, TableCell, TableHead, TableHeader, TableRow, Text } from '@/components'
-import { LatestFileTypes, RepoFile, SummaryItemType, TranslationStore } from '@/views'
+import { FileStatus, LatestFileTypes, RepoFile, SummaryItemType, TranslationStore } from '@/views'
 import { FileLastChangeBar } from '@views/repo/components'
 
 interface RoutingProps {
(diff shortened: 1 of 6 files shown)


packages/ui/src/views/repo/components/summary/summary.tsx
```diff
@@ -65,10 +65,40 @@ export const Summary = ({
           {files.map(file => (
             <TableRow key={file.id} onClick={() => navigate(file.path)}>
               <TableCell>
-                <div className="flex cursor-pointer items-center gap-1.5">
+                <div
+                  className={`flex cursor-pointer items-center gap-1.5 ${
+                    file.status && file.status !== FileStatus.SAFE
+                      ? file.status === FileStatus.LOW_RISK
+                        ? 'absolute left-0 border-l-2 border-yellow-300'
+                        : file.status === FileStatus.MEDIUM_RISK
```
abhinavrastogi-harness, line 73: this would be better shared as a util function which takes the `FileStatus` enum as a param and returns the correct color class. it can then be added on both the div and the icon

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
```diff
@@ -302,6 +302,7 @@ export function RepoSummaryView({
               is_public={repository?.is_public}
               useTranslationStore={useTranslationStore}
             />
+            <Outlet />
```
abhinavrastogi-harness, line 305: instead of Outlet (the cards will not be a separate route), this should be a renderProp to the repoSummary component

packages/ui/src/views/repo/components/summary/summary.tsx
sans-harness: Lets do this as a follow up PR and merge this for now?

packages/ui/src/views/repo/repo-summary/repo-summary.tsx
sans-harness: Done!

abhinavrastogi-harness (approved):

PR #972, opened 2025-02-04
feat: Add "readonly" mode for Pipeline Graph

vardanbansal-harness:
With "readonly" `true`: 

image

With "readonly" `false` (default): 

image


apps/design-system/src/pages/view-preview/execution-logs-view-wrapper copy.tsx
abhinavrastogi-harness, line 1: name: why is it `copy.tsx`?

abhinavrastogi-harness (approved):

apps/design-system/src/pages/view-preview/execution-logs-view-wrapper copy.tsx
vardanbansal-harness: Fixed it, removed `copy` from the name.

PR #974, opened 2025-02-05
feat: support for animating the yaml loading in

abhinavrastogi-harness:
https://github.com/user-attachments/assets/1a30e226-82b4-4597-a0e6-43780132a4fd

### DIFF
--- a/packages/yaml-editor/src/components/YamlEditor.tsx
+++ b/packages/yaml-editor/src/components/YamlEditor.tsx
@@ -44,6 +44,8 @@ export interface YamlEditorProps<T> {
   }
   minimap?: boolean
   folding?: boolean
+  animateOnUpdate?: boolean
+  onAnimateEnd?: () => void
 }
 
 export const YamlEditor = function YamlEditor<T>(props: YamlEditorProps<T>): JSX.Element {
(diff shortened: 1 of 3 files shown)


packages/yaml-editor/src/components/YamlEditor.tsx
```diff
@@ -36,16 +36,19 @@ export interface YamlEditorProps<T> {
...
   themeConfig?: { rootElementSelector?: string; defaultTheme?: string; themes?: ThemeDefinition[] }
   theme?: string
+  options?: monaco.editor.IStandaloneEditorConstructionOptions
   selection?: {
     path: string
     className: string
     revealInCenter?: boolean
   }
   minimap?: boolean
   folding?: boolean
```
srdjan-harness, line 46: We can remove these and refactor as we exposed IStandaloneEditorConstructionOptions.

packages/yaml-editor/src/components/YamlEditor.tsx
```diff
@@ -91,18 +138,21 @@ export function YamlEditor<T>(props: YamlEditorProps<T>): JSX.Element {
...
           editorRef.current.executeEdits('edit', [
             {
               range: model.getFullModelRange(),
-              text: yamlRevision.yaml
+              text: animateOnUpdate ? '' : yamlRevision.yaml
             }
           ])
-          editorRef.current.pushUndoStop()
+          // editorRef.current.pushUndoStop()
+          if (animateOnUpdate) {
```
srdjan-harness, line 149: We can probably add if else here. if (!animateOnUpdate) .... else...

PR #981, opened 2025-02-05

cxramina:
Updated tokens value for dark-std-std to match original PP design


knagurski (approved):
Looks good 👍🏻

PR #990, opened 2025-02-05
add memory router for docs

abhinavrastogi-harness:


apps/portal/src/content/docs/components/tabnav.mdx
```diff
@@ -65,13 +65,14 @@ An individual tab item that links to a specific route. It accepts the same props
...
+>
+  {label}
+</TabNav.Item>
 ```
 
 ### Props
 
 - `to`: A string representing the route path to link to.
-- `label`: A string representing the label to display for the tab.
+- `label`: A string or ReactElement representing the label to display for the tab.
```
knagurski, line 77: Is this actually a `ReactNode`?

knagurski (approved):
Good solution. Nice way to nerf the navigation without breaking things

apps/portal/src/content/docs/components/tabnav.mdx
abhinavrastogi-harness: indeed it is

PR #994, opened 2025-02-06
Webhook execution details

sans-harness:
https://github.com/user-attachments/assets/c69c2077-f1f9-433f-93fb-b68d94517166

The `<CodeEditor />` does not work as expected. The content is not wrapped around. Lets take that up seperately and not block this PR.

### DIFF
--- a/packages/ui/src/utils/utils.ts
+++ b/packages/ui/src/utils/utils.ts
@@ -110,34 +110,3 @@ export function formatNumber(num: number | bigint): string {
 export interface Violation {
   violation: string
 }
-
-/**
- * Convert nanoseconds to simple human-readable format
- * @example
- * formatNs(416_328_583) // "416ms"
- * formatNs(1_234_567_890) // "1s 234ms"
- * formatNs(90_000_000_000) // "1m 30s"
- */
-export function formatNs(nanoseconds: number): string {
-  if (!nanoseconds || nanoseconds < 0) return '0s'
-
-  const seconds = nanoseconds / 1_000_000_000
-  const duration = intervalToDuration({
-    start: 0,
-    end: Math.round(seconds * 1000) // Convert to milliseconds
-  })
-
-  const parts: string[] = []
-  if (duration.days) parts.push(`${duration.days}d`)
-  if (duration.hours) parts.push(`${duration.hours}h`)
-  if (duration.minutes) parts.push(`${duration.minutes}m`)
-  if (duration.seconds) parts.push(`${duration.seconds}s`)
-
-  // Show milliseconds for durations under 1 minute
-  if (seconds < 60) {
-    const ms = Math.round((nanoseconds % 1_000_000_000) / 1_000_000)
-    if (ms > 0) parts.push(`${ms}ms`)
-  }
-
-  return parts.join(' ') || '0s'
-}
--- a/packages/ui/src/utils/utils.ts
+++ b/packages/ui/src/utils/utils.ts
@@ -1,4 +1,4 @@
-import { formatDistance, formatDistanceToNow, intervalToDuration } from 'date-fns'
+import { formatDistance, formatDistanceToNow } from 'date-fns'
 
 export const getInitials = (name: string, length = 2) => {
   // Split the name into an array of words, ignoring empty strings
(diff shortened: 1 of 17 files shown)


packages/ui/src/utils/utils.ts
```diff
@@ -110,3 +110,34 @@ export function formatNumber(num: number | bigint): string {
...
 }
+
+/**
+ * Convert nanoseconds to simple human-readable format
+ * @example
+ * formatNs(416_328_583) // "416ms"
+ * formatNs(1_234_567_890) // "1s 234ms"
+ * formatNs(90_000_000_000) // "1m 30s"
+ */
+export function formatNs(nanoseconds: number): string {
```
abhinavrastogi-harness, line 121: we already have a function that takes milliseconds and returns this same format.
can we just extend/reuse that function to accept nanoseconds as well?

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-execution-details-page.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <Button variant="default" size="md" onClick={handleRetriggerExecution} disabled={isLoading}>
+              {isLoading ? 'Re-triggering Execution' : 'Re-trigger Execution'}
+            </Button>
+          </ListActions.Right>
+        </ListActions.Root>
+
+        <Spacer size={6} />
+        <div className="flex gap-10">
+          <div className="flex gap-1">
+            <Text color="foreground-5">Triggered Event:</Text>
```
abhinavrastogi-harness, line 100: i18n missing in many places. not a blocker

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-execution-details-page.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <Text color="foreground-5">Duration:</Text>
+            <Text>{formatNs(execution?.duration ?? 0)}</Text>
+          </div>
+        </div>
+        <Spacer size={6} />
+        <WebhookExecutionEditorControlBar view={view} onChangeView={onChangeView} />
+        <CodeEditor
+          height="500px"
+          language="json"
+          codeRevision={codeEditorContent}
```
abhinavrastogi-harness, line 117: can we pretty-print the JSON instead of a single line as seen in the video?
something like `JSON.stringify(obj, null, 4)`

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
```diff
@@ -65,7 +68,11 @@ const RepoWebhookExecutionsPage: FC<RepoWebhookExecutionsPageProps> = ({
               </TableHeader>
               <TableBody>
                 {executions.map(execution => (
-                  <TableRow key={execution.id}>
+                  <TableRow
+                    key={execution.id}
+                    onClick={() => navigate(toRepoWebookExecutionDetails(`${execution.id}`))}
```
abhinavrastogi-harness, line 73: this should be a Link. `navigate` should only be used for navigations triggered via code.

abhinavrastogi-harness (commented):
in the video the editor does not seem to respect the dark theme.
and the request/response json should be formatted

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
sans-harness: I dont think a `Link` can wrap a `table-row`. The html is invalid. Also attaching a pic. It no longer recognizes separate `table-cells`. The only other place we have such a usecase is on the `summary` page which also uses an `onClick` handler. 
Screenshot 2025-02-07 at 1 34 33 PM

abhinavrastogi-harness: did we try using `asChild`?

abhinavrastogi-harness (approved):

PR #998, opened 2025-02-06
Update language dialog to accept supported languages as prop

praneshg239:


packages/ui/src/components/language-selector/language-dialog.tsx
```diff
@@ -3,11 +3,12 @@ import { FC, useEffect, useState } from 'react'
 import { Dialog, Icon } from '@/components'
 import { cn } from '@utils/cn'
 
-import { LanguageCode, LanguageDialogProps, languages } from './types'
+import { LanguageCode, LanguageDialogProps } from './types'
```
abhinavrastogi-harness, line 6: can we also delete the `languages` object if it's not being used anymore

praneshg239: sure

abhinavrastogi-harness (approved):

PR #1004, opened 2025-02-06
Fixed success icon color

vivek-harness:
Before fix:
successicon-before-fix

After fix:
success-icon-after-fix

### DIFF
--- a/packages/ui/src/views/execution/execution-status.tsx
+++ b/packages/ui/src/views/execution/execution-status.tsx
@@ -59,15 +59,15 @@ const Badge: React.FC<ExecutionStatusProps & BadgeProps> = props => {
       return minimal ? (
         <div className="flex items-center gap-1">
           <div className="size-2 rounded-full bg-success" />
-          <span className="text-success">Success</span>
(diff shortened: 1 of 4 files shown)


srdjan-harness (approved):

packages/ui/src/views/execution/execution-status.tsx
```diff
@@ -64,7 +64,7 @@ const Badge: React.FC<ExecutionStatusProps & BadgeProps> = props => {
       ) : (
         <div className="flex items-center gap-1 rounded-md">
           <div className="flex items-center gap-0.5">
-            <CanaryIcon size={12} name="success" />
+            <CanaryIcon size={12} name="success" className="text-foreground-success" />
```
abhinavrastogi-harness, line 67: how come this icon uses `text-foreground-sucess` but the text below uses `text-sucess`.
if they were the same, we chould just set the class on the parent div for both icon and text to be consistent.

vivek-harness: Fixed

PR #1006, opened 2025-02-06
feat: add Chat.CodeBlock component

abhinavrastogi-harness:
Also adds support for pressing the Enter key to send a message
Screenshot 2025-02-06 at 9 25 35 AM


c-joetaylor-harness (commented):
I'd tried to avoid hardcoding text and adding logic into the actual chat component, can this not be handled outside?

PR #1019, opened 2025-02-07
update pipeline problems bar to correct icons and colors

praneshg239:
### Error icon with error count > 0
Pasted Graphic

#### Light theme:
image

### Error icon with error count equals 0
Pasted Graphic 1

#### Light theme:
image


srdjan-harness (approved):

abhinavrastogi-harness (approved):
is the `canary` package still being used? changes should not be required there

praneshg239:
> is the `canary` package still being used? changes should not be required there

Yes @abhinavrastogi-harness, we are using canary and views package here. I added the icons in both `ui` and `canary` packages anyway.

PR #1021, opened 2025-02-07

cxramina:
Adjusted colors for new modes


praneshg239 (approved):

packages/ui/src/shared-style-variables.css
```diff
@@ -1476,13 +1485,13 @@
 
     /* Backgrounds */
     --canary-background-01: var(--canary-base-chrome-black);
-    --canary-background-02: var(--canary-base-chrome-900);
+    --canary-background-02: var(--canary-base-chrome-950);
     --canary-background-03: var(--canary-base-chrome-950);
```
praneshg239, line 1489:
```css
--canary-background-02: var(--canary-base-chrome-950);
--canary-background-03: var(--canary-base-chrome-950);
```

We have these two variant as same colors. Is this expected?

PR #1023, opened 2025-02-07
feat: Fix animation on Execution tree with logs

vardanbansal-harness:
https://github.com/user-attachments/assets/9acd6ca7-7b61-42a3-8608-f47e629fa880


abhinavrastogi-harness (commented):
the logs seem to be completely independent of the tree or the selected nodes. I don't think that's the desired behaviour

abhinavrastogi-harness (approved):

PR #1035, opened 2025-02-07

cxramina:
Changed dark-std-std to dark-std-low 

and updated dark-std-std with tokens from high version


knagurski (approved):
👍🏻

PR #1059, opened 2025-02-08
fix: timer, architecture

sans-harness:
Move logic to design system. Fix views.

### DIFF
--- a/packages/ui/src/views/execution/pipeline-status.tsx
+++ b/packages/ui/src/views/execution/pipeline-status.tsx
@@ -1,10 +1,17 @@
-import { FC, useEffect, useRef, useState } from 'react'
+import { FC } from 'react'
 
 import { ExecutionState } from '@views/repo/pull-request'
 
 import { ExecutionStatus } from './execution-status'
 
-interface PipelineStatusProps {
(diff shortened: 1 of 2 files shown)


apps/portal/src/content/docs/components/caption.mdx
abhinavrastogi-harness, line 1: bad rebase? these files should not be in your PR I think

sans-harness: Just rebased everything. Please re-check.

packages/ui/src/views/execution/pipeline-status.tsx
```diff
@@ -1,67 +1,23 @@
-import { FC, useEffect, useRef, useState } from 'react'
-
 import { ExecutionState } from '@views/repo/pull-request'
 
 import { ExecutionStatus } from './execution-status'
 
-interface PipelineStatusProps {
+const PipelineStatus = ({
```
abhinavrastogi-harness, line 5: looks like the interface got removed, and we have inline types now.. 
not a blocker, but let's come back to it later

abhinavrastogi-harness (approved):

PR #1070, opened 2025-02-10
docs: added docs for ui components

harish-viswa:


knagurski:
Can we break this up into different PRs? This is too much to review without blindly rubber-stamping

knagurski (changes requested):
These appear to be mass-produced AI-generated docs with issues. All docs additions should be their own PRs and need to be checked over to make sure they're correct (e.g. the first component is `BranchSelector`, but is in a file called `branch-chooser` with `Branch Chooser` referenced etc).

abhinavrastogi-harness:
agreed with the bocker here. we need to ensure correctness and basic quality. individual PRs would help with that. 

however, we do need to get some basic docs ready for a majority of components, because we want to open up the component library to other teams very soon. AI can be a good way to kickstart something, but we should not be blindly committing "AI Slop".

harish-viswa:
> These appear to be mass-produced AI-generated docs with issues. All docs additions should be their own PRs and need to be checked over to make sure they're correct (e.g. the first component is `BranchSelector`, but is in a file called `branch-chooser` with `Branch Chooser` referenced etc).

@knagurski  Actually the file-name is `Branch Chooser` but the component being exported is BranchSelector, we may need to update this in BranchChooser component, thought this should not be a blocker for docs

harish-viswa:
@knagurski @abhinavrastogi-harness 
I was not able to mass produce the results actually, due to free tier usage 😢 
I created two to three components as a group and reviewed each of them for
1. Usage correctness
2. Props for components and sub-components

we may have to re-look at the text content being generated which I did not give much importance to,
since it can be fine-tuned later

PR #1072, opened 2025-02-11
feat: restore components animations

### LINKED ISSUE #780: Restore component animations
Due to janky animations with a number of Radix components, it was decided that animations should be removed. The underlying Radix issue has been resolved and the removed animations should be restored.

### LINKED ISSUE #791: Namespace `Popover` components
- Popover
- PopoverAnchor
- PopoverContent
- PopoverTrigger

3em:
fixes #780
fixes #791

### DIFF
--- a/packages/ui/src/components/problems.tsx
+++ b/packages/ui/src/components/problems.tsx
@@ -35,13 +35,12 @@ const ProblemsComponent = {
   },
 
   Row: function Root({ onClick, children }: { onClick?: () => void; children: React.ReactNode }) {
-    const rowClasses = 'bg-neutral-800 text-neutral-200'
     return (
       <div
         role="button"
         tabIndex={0}
         onClick={onClick}
-        className={`width-100 flex flex-1 cursor-pointer items-center justify-between gap-2 text-nowrap px-4 py-0.5 ${rowClasses}`}
+        className="width-100 flex flex-1 cursor-pointer items-center justify-between gap-2 text-nowrap px-4 py-0.5 bg-neutral-800 text-neutral-200"
       >
         {children}
       </div>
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/problems.tsx
```diff
@@ -41,7 +41,7 @@ const ProblemsComponent = {
         role="button"
         tabIndex={0}
         onClick={onClick}
-        className={`width-100 flex flex-1 cursor-pointer items-center gap-2 text-nowrap py-0.5 justify-between px-4 ${rowClasses}`}
+        className={`width-100 flex flex-1 cursor-pointer items-center justify-between gap-2 text-nowrap px-4 py-0.5 ${rowClasses}`}
```
knagurski, line 44: Not a problem with this PR, but we should try to remove string concatenation and instead use `cn`
```suggestion
        className={cn('width-100 flex flex-1 cursor-pointer items-center justify-between gap-2 text-nowrap px-4 py-0.5', rowClasses)}
```

packages/ui/src/views/layouts/SandboxLayout.tsx
```diff
@@ -18,7 +18,7 @@ function Root({ children }: { children: ReactNode }) {
 
 function LeftPanel({ children, className }: { children: ReactNode; className?: string }) {
   return (
-    <div className="sticky top-0 z-10 h-screen">
+    <div className="sticky top-0 z-50 h-screen">
```
knagurski, line 21: Is this intentional/required? Might be better pulled out into a bug fix ticket to keep this PR focused on the restoration of the animations.

3em: @knagurski This is related to the animation since I added an opening animation for the menu in the navigation, which should open beneath the navigation. I checked that the current z-index doesn’t break anything and is only used for the `Navbar`.

packages/ui/src/components/problems.tsx
3em: This fix was just the result of running pnpm pretty --write. So now I completely removed the unnecessary rowClasses.

knagurski (approved):
Nice stuff 👍🏻

PR #1077, opened 2025-02-11
Add hover card to time-ago to show UTC/Local time

sans-harness:
1. If time ago > 3 days, show absolute time instead of 3 days ago. 
2. Adds `hovercard` to show UTC and local time when something was executed.  

https://github.com/user-attachments/assets/df26b8b1-9ec1-48d1-a012-b766f76461ad


packages/ui/src/components/time-ago-hover-card.tsx
```diff
@@ -0,0 +1,49 @@
+import { timeAgo } from '../utils/utils'
+import { Button } from './button'
+import { HoverCard, HoverCardContent, HoverCardTrigger } from './hover-card'
+
+export const TimeAgoHoverCard = ({ timeStamp }: { timeStamp: number }) => {
```
abhinavrastogi-harness, line 5: this can be a view instead

packages/ui/src/utils/utils.ts
```diff
@@ -70,14 +70,36 @@ export const timeDistance = (date1 = 0, date2 = 0, onlyHighestDenomination = fal
...
  * timeAgo(1708113838167) // Returns "1 hour ago"
  */
+
 export const timeAgo = (timestamp?: number | null): string => {
   if (timestamp === null || timestamp === undefined) {
     return 'Unknown time'
   }
 
+  const now = Date.now()
+  const threeDaysMs = 3 * 24 * 60 * 60 * 1000
```
abhinavrastogi-harness, line 80: let's make this a optional second prop, which defaults to 3 days.
hardcoded is no-beuno.

packages/ui/src/utils/utils.ts
```diff
@@ -70,14 +70,36 @@ export const timeDistance = (date1 = 0, date2 = 0, onlyHighestDenomination = fal
...
+
+  if (isOld) {
+    const date = new Date(timestamp)
+    return (
+      new Intl.DateTimeFormat(LOCALE, {
+        month: 'short',
+        day: 'numeric',
+        year: 'numeric'
+      }).format(date) +
+      ' at ' +
```
abhinavrastogi-harness, line 91: probably a product decision, but I don't mind if we remove the `at` and just use one `Intl()` to show the complete date + time string

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
```diff
@@ -99,8 +100,9 @@ const RepoWebhookExecutionsPage: FC<RepoWebhookExecutionsPageProps> = ({
                             : 'Invalid'}
                       </Badge>
                     </TableCell>
-                    <TableCell className="text-right">
-                      <Text size={2}>{timeAgo(execution.created)}</Text>
+                    <TableCell className="text-right relative">
+                      {/* Add relative positioning */}
+                      <TimeAgoHoverCard timeStamp={execution.created ?? 0} />
```
abhinavrastogi-harness, line 105: can we look into updating the `timeAgo()` function to return JSX instead of a string?
that way we don't need to change anything to adopt this hovercard across the app.

packages/ui/src/utils/utils.tsx
```diff
@@ -66,20 +69,43 @@ export const timeDistance = (date1 = 0, date2 = 0, onlyHighestDenomination = fal
...
  * Formats timestamp to relative time (e.g., "1 hour ago")
  * @param timestamp - Unix timestamp in milliseconds
+ * @param cutoffDays - Days within which to use relative time (default: 3)
  * @returns formatted relative time string
  * @example
  * timeAgo(1708113838167) // Returns "1 hour ago"
  */
-export const timeAgo = (timestamp?: number | null): string => {
+
+export const timeAgo = (timestamp?: number | null, cutoffDays: number = 3): ReactNode => {
```
abhinavrastogi-harness, line 78: i'd prefer to move the util to a different file since the file needs to be a tsx now.
generally most utils should be in ts files.

abhinavrastogi-harness (approved):

packages/ui/src/utils/utils.tsx
sans-harness: We can do `React.createElement` and return JSX from a `*.ts` file. Doing that and reverting to `.ts`.

packages/ui/src/views/repo/components/time-ago-hover-card.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const localTimeFormatter = new Intl.DateTimeFormat(undefined, {
+  hour: '2-digit',
+  minute: '2-digit',
+  second: '2-digit'
+})
+
+interface TimeAgoHoverCardProps {
+  formattedDate: string
+  timeStamp: number
```
knagurski, line 33: "timestamp" is one word, not "time stamp", so it shouldn't be camel-cased

packages/ui/src/components/hover-card.tsx
```diff
@@ -0,0 +1,27 @@
...
+    className={cn(
+      'bg-popover text-popover-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 z-50 w-64 rounded-md border p-4 shadow-md outline-none',
+      className
+    )}
+    {...props}
+  />
+))
+HoverCardContent.displayName = HoverCardPrimitive.Content.displayName
+
+export { HoverCard, HoverCardTrigger, HoverCardContent }
```
knagurski, line 27: Multi-part components should be namespaced.

```typescript
const HoverCard = {
  Root: HoverCardRoot,
  Trigger: HoverCardTrigger,
  Content: HoverCardContent
}

export { HoverCard }
```

packages/ui/src/utils/utils.ts
```diff
@@ -1,3 +1,6 @@
+import React, { ReactNode } from 'react'
```
knagurski, line 1: Importing `React` like this causes an ESLint warning since `react` doesn't have a default export. To use `React.X`, you need to import react as `import * as React from 'react'`, or we need to turn on synthetic default imports in the TS config. Best option is to just import whatever you need directly and drop the `React.` part.

PR #1086, opened 2025-02-13
fix: repo summary empty state and generateToken from NGUI MFE

shaurya-harness:
fixes:
empty repo page based on is_empty from repoMetadata
profile keys route is different for NGUI - passed navigate hook as part of customUtil
clone credential token generation api is different for NGUI, goes through gateway and is passed as a hook from NGUI - added the customHook to be used in case of MFE

### DIFF
--- a/apps/gitness/src/AppMFE.tsx
+++ b/apps/gitness/src/AppMFE.tsx
@@ -83,6 +83,9 @@ interface AppMFEProps {
   customHooks: Partial<{
     useGenerateToken: Unknown
   }>
+  customUtils: Partial<{
+    navigateToUserProfile: Unknown
+  }>
 }
 
 function decode<T = unknown>(arg: string): T {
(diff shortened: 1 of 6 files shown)


vardanbansal-harness (approved):

apps/gitness/src/AppMFE.tsx
```diff
@@ -12,7 +12,7 @@ import { PortalProvider } from '@harnessio/ui/context'
 
 import ShadowRootWrapper from './components-v2/shadow-root-wrapper'
 import { ExitConfirmProvider } from './framework/context/ExitConfirmContext'
-import { MFEContext } from './framework/context/MFEContext'
+import { MFEContext, Unknown } from './framework/context/MFEContext'
```
praneshg239, line 15: @shaurya-harness what is `Unknown` import from `MFEContext`? can you please check?

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -107,9 +117,20 @@ git push -u origin main
             <MarkdownViewer source={getExistingRepoMarkdown()} />
             <p>
               You might need to{' '}
-              <StyledLink to={toProfileKeys?.() ?? ''} relative="path">
+              <span
```
praneshg239, line 120: can we use a button instead?

PR #1091, opened 2025-02-14
fix: refactoring of the PR conversation page timeline section

3em:
- Restored data rendering in the timeline
- Fixed time display in the timeline

image


packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-timeline-item.tsx
```diff
@@ -93,7 +95,7 @@ const ItemHeader: FC<ItemHeaderProps> = memo(
           </DropdownMenu.Trigger>
           <DropdownMenu.Content className="w-[216px]" align="end">
             <DropdownMenu.Group>
-              {!hideEditDelete ? <DropdownMenu.Item onClick={onEditClick}>Edit</DropdownMenu.Item> : null}
+              {!!hideEditDelete && <DropdownMenu.Item onClick={onEditClick}>Edit</DropdownMenu.Item>}
```
knagurski, line 98: Does this flip the logic to show when `hideEditDelete` is truthy?
```suggestion
              {!hideEditDelete && <DropdownMenu.Item onClick={onEditClick}>Edit</DropdownMenu.Item>}
```

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-timeline-item.tsx
```diff
@@ -6,13 +6,15 @@ import { getInitials } from '@utils/utils'
...
-    selectStatus?: ReactNode
-  }[]
+export interface TimelineItemPropsHeaderType {
+  avatar?: ReactNode
+  name?: string
+  description?: ReactNode
+  selectStatus?: ReactNode
+}
+
+export interface TimelineItemProps {
```
knagurski, line 16: Probably no need to export `TimelineItemPropsHeaderType` since you can get it from `TimelineItemProps` using `TimelineItemProps['header']`

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-system-comments.tsx
```diff
@@ -36,205 +35,259 @@ const PullRequestSystemComments: React.FC<SystemCommentProps> = ({
...
+    merge_method,
+    merge_sha,
+    reviewer_type,
+    label,
+    label_color,
+    value
+  } = payload as GeneralPayload
+
+  const openFromDraft = old_draft === true && new_draft === false
+  const changedToDraft = old_draft === false && new_draft === true
```
knagurski, line 61: There's a lot of hoops to be jumped through here to get at some basic meta data. Feels icky. We should probably revisit and solidify the types at some point to reflect the realities of the data.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-system-comments.tsx
```diff
@@ -36,205 +35,259 @@ const PullRequestSystemComments: React.FC<SystemCommentProps> = ({
...
+      <Avatar.Fallback>{getInitials(author?.display_name || '')}</Avatar.Fallback>
+    </Avatar.Root>
+  )
+
+  const {
+    header,
+    ...restProps
+  }: Partial<Omit<TimelineItemProps, 'header'>> & {
+    header: TimelineItemPropsHeaderType
+  } = (() => {
```
knagurski, line 78: Instead of a self-executing function, can we change this to a `useMemo`? Because this function is redefined and invoked on every render, it's quite inefficient. Using a `useMemo` will ensure that things are only reevaluated when the inputs change.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-system-comments.tsx
3em: Agree, but I think there's too little specificity in the data types. However, the `GeneralPayload` type was defined by @cjlee01, and I haven't looked into why exactly it's `[key: string]: unknown`. In the current PR, I won't be able to change this type.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-system-comments.tsx
3em: Agree, my plan was to eventually move the entire component to memo. But for now, in this PR, it's better not to leave the IIFE. Thanks!

knagurski (approved):

PR #1097, opened 2025-02-18
Delete yaml-editor-playground

srdjan-harness:


vardanbansal-harness (approved):

abhinavrastogi-harness (approved):

abhinavrastogi-harness:
it might be better to combine this with https://github.com/harness/canary/pull/1096 so that git can maintain file history (it should see it as a "move" op instead of "delete" and "create" ops)

srdjan-harness:
> it might be better to combine this with #1096 so that git can maintain file history (it should see it as a "move" op instead of "delete" and "create" ops)

Demo examples are refactored. Its better to keep changes in different PRs.

PR #1128, opened 2025-02-19
fix: refactoring of the PR conversation page comments blocks

3em:
- Fixed comments rendering
- Refactored the `PullRequestOverview` component
- Part of the work on optimization and refactoring of unnecessary renders

image


apps/gitness/src/hooks/useAPIPath.ts
```diff
@@ -1,12 +1,20 @@
...
   const mfeContext = useMFEContext()
-  const isMFE = mfeContext.renderUrl !== ''
-  return function (path: string) {
-    if (isMFE) {
-      return `${window.apiUrl || ''}/code${path}${path.includes('?') ? '&' : '?'}routingId=${mfeContext.scope.accountId}`
-    }
-    return `${window.apiUrl || ''}${path}`
-  }
+
+  return useCallback(
```
abhinavrastogi-harness, line 8: why is this changed needed? this is a high-risk change.

apps/gitness/src/hooks/useGetPullRequestTab.ts
```diff
@@ -46,10 +46,11 @@ const useGetPullRequestTab = ({
...
     const matchedRoute = routeTabMapping.find(route => route.match)
-    if (matchedRoute) {
-      setPullRequestTab(matchedRoute.tab)
-    }
-  }, [routeTabMapping, spaceId, repoId, pullRequestId])
+    if (!matchedRoute) return
+
+    setPullRequestTab(matchedRoute.tab)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [spaceId, repoId, pullRequestId])
```
abhinavrastogi-harness, line 53: why was `routeTabMapping` removed from dependency list?
ignoring eslint rules is not ideal.

apps/gitness/src/pages-v2/pull-request/hooks/usePRCommonInteractions.ts
abhinavrastogi-harness, line 1: why are any of these change needed?

apps/gitness/src/hooks/useAPIPath.ts
3em: Because in any place where this hook was used, and the method was used inside useEffect or other places, it always triggered a re-render since the function was not memoized. Can you tell me please what the risk?

apps/gitness/src/hooks/useGetPullRequestTab.ts
3em: Because `routeTabMapping` (which is an array containing data that uses the `useMatch` hook) is not memoized and gets recreated on every render. In its previous implementation, this caused the `useEffect` to trigger on every render.

In the current version, when the specified dependencies in useEffect change, routeTabMapping will contain up-to-date values and remain relevant within that useEffect.

But it's not a problem — I can roll back this fix since it's not critical.

apps/gitness/src/pages-v2/pull-request/hooks/usePRCommonInteractions.ts
3em: The reason for these changes is similar: since this method is not memoized, it gets recreated on every render, causing the entire chain of components where it is passed to update.

This fix does not change any existing functionality of the method — it only memoizes it.

apps/gitness/src/hooks/useGetPullRequestTab.ts
3em: @abhinavrastogi-harness I have reverted the old code for this issue

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
```diff
@@ -34,13 +34,13 @@ const RepoWebhookExecutionsPage: FC<RepoWebhookExecutionsPageProps> = ({
   return (
     <SandboxLayout.Main className="mx-0">
       <SandboxLayout.Content className="pl-0">
-        <h1 className="text-foreground-1 mb-4 text-2xl font-medium">Order Status Update Webhook</h1>
+        <h1 className="mb-4 text-2xl font-medium text-foreground-1">Order Status Update Webhook</h1>
```
cjlee01, line 37: can we move this to en/ views.json

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
```diff
@@ -34,13 +34,13 @@ const RepoWebhookExecutionsPage: FC<RepoWebhookExecutionsPageProps> = ({
...
         <Text>
           This webhook triggers every time an order status is updated, sending data to the specified endpoint for
```
cjlee01, line 39: can we move this to en/ views.json

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
```diff
@@ -34,13 +34,13 @@ const RepoWebhookExecutionsPage: FC<RepoWebhookExecutionsPageProps> = ({
...
           real-time tracking.
         </Text>
         <FormSeparator className="my-6" />
-        <h1 className="text-foreground-1 mb-4 text-xl font-medium">Executions</h1>
+        <h1 className="mb-4 text-xl font-medium text-foreground-1">Executions</h1>
```
cjlee01, line 43: can we move this to en/ views.json

packages/ui/src/views/repo/pull-request/details/components/conversation/regular-and-code-comment.tsx
```diff
@@ -0,0 +1,341 @@
...
+                    setComment={text => setEditComments(prev => ({ ...prev, [componentId]: text }))}
+                  />
+                ) : (
+                  <ComponentViewBase commentItem={commentItem} />
+                )
+              }}
+              customHeaderData={{
+                name,
+                avatar: undefined,
+                // TODO: fix comment to tell between comment or code comment?
```
cjlee01, line 245: can remove this todo if this was extracted out

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-timeline-item.tsx
```diff
@@ -81,75 +37,104 @@ const ItemHeader: FC<ItemHeaderProps> = memo(
...
+                    {
+                      title: 'Delete comment',
+                      onClick: () => handleDeleteComment?.(),
+                      isDanger: true
+                    }
+                  ]
+                : [])
+            ]}
+          />
         )}
```
cjlee01, line 82: probs could put this in a variable above to make it more readable and cleaner

packages/ui/src/views/repo/pull-request/details/components/conversation/regular-and-code-comment.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+const getAvatar = (name?: string) => {
+  return (
+    <Avatar.Root>
+      <Avatar.Fallback>{getInitials(name || '')}</Avatar.Fallback>
+    </Avatar.Root>
+  )
+}
+
+export interface PullRequestRegularAndCodeCommentProps
```
cjlee01, line 33: should we just move this to types file

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
3em: @cjlee01 Thanks for pointing it out, but we haven't gotten to refactoring this section yet. The current changes in this PR are the result of running lint --fix. We will definitely add translations as soon as we get to this section.

packages/ui/src/views/repo/webhooks/webhook-executions/repo-webhook-executions-list-page.tsx
3em: @cjlee01 Thanks for pointing it out, but we haven't gotten to refactoring this section yet. The current changes in this PR are the result of running lint --fix. We will definitely add translations as soon as we get to this section.

packages/ui/src/views/repo/pull-request/details/components/conversation/regular-and-code-comment.tsx
3em: @cjlee01 It seems unnecessary since this interface is only used for typing the current component, and it mainly consists of extends.

apps/gitness/src/pages-v2/pull-request/hooks/usePRCommonInteractions.ts
```diff
@@ -14,6 +14,8 @@ import { generateAlphaNumericHash } from '@harnessio/views'
 import { useAPIPath } from '../../../hooks/useAPIPath'
 import { getErrorMessage } from '../pull-request-utils'
 
+let count = generateAlphaNumericHash(5)
```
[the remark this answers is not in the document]
knagurski: Also looking at the usage of `count`, it's "incremented" further down (line 128). Since this is an alpha-numeric string, it won't increment, but instead a `1` will be appended each time. Might not matter, but seems super weird.

packages/ui/src/views/repo/pull-request/details/components/conversation/regular-and-code-comment.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const { highlight, wrap, fontsize } = useDiffConfig()
+
+  const [hideReplyHeres, setHideReplyHeres] = useState<{ [key: string]: boolean }>({})
+  const [quoteReplies, setQuoteReplies] = useState<Record<number, { text: string }>>({})
+  const [editModes, setEditModes] = useState<{ [key: string]: boolean }>({})
+  const [editComments, setEditComments] = useState<{ [key: string]: string }>({})
+
+  const payload = useMemo(() => {
+    return commentItems[0]?.payload
+  }, [commentItems])
```
knagurski, line 73: Probably no need for a `useMemo` here. It won't stabilise the reference and the overhead will slow things down a wee bit.
```suggestion
  const payload = commentItems[0]?.payload
```

packages/ui/src/views/repo/pull-request/details/components/conversation/regular-and-code-comment.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+  const isCode = useMemo(() => isCodeComment(commentItems), [commentItems])
+
+  const headerData: TimelineItemProps['header'][number] = useMemo(
+    () => ({
+      avatar: getAvatar(payload?.author?.display_name),
+      name: payload?.author?.display_name
+    }),
+    [payload?.author]
```
knagurski, line 82: If you make this a wee bit more specific, you can guard against if `author` isn't dependency array safe (not saying it's an issue, just a good practice)
```suggestion
    [payload?.author?.display_name]
```

apps/gitness/src/hooks/useAPIPath.ts
abhinavrastogi-harness: this risk is high because this impacts all API calls in both standalone and MFE mode. this will need to be tested in the MFE before we can merge.
cc @praneshg239

abhinavrastogi-harness (approved):
@knagurski has some good comments. Looks good from my side apart from that.

apps/gitness/src/pages-v2/pull-request/hooks/usePRCommonInteractions.ts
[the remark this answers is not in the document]
3em: @knagurski Moved the declaration of count into a ref. Regarding the alphanumeric issue—yes, there's something odd here, but I'm not competent to comment further since this code was already in place. I merely extracted it from the component in my previous commit because it was being recreated each time.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-timeline-item.tsx
```diff
@@ -193,7 +180,7 @@ const PullRequestTimelineItem: FC<TimelineItemProps> = ({
   }, [isResolved])
 
   const renderContent = () => {
-    if (!content) return null
+    if (!content) return <></>
```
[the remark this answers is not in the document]
3em: Reverted these changes.

packages/ui/src/views/repo/pull-request/details/components/conversation/regular-and-code-comment.tsx
3em: In that case, I wrapped the entire component in `memo` since payload is currently used as a dependency in multiple hooks. This way, I can remove all individually memoized data, and the component itself will only re-render upon receiving new props.

PR #1131, opened 2025-02-19
refactor: style improvements for the User Management page 

andrewgolovanov:
**Description:**
This pull request adds style improvements as well as refactoring of current components for the `User Management` page 

**Before:**
image

**After:**
image

**Before:** 
image

**After:**
image

**Before:**
image

**After:**
image

**Before:**
image

**After:**
image

**Before:**
image

**After:**
image

**Before:**
image

**After:**
image


apps/gitness/src/pages-v2/user-management/user-management-container.tsx
```diff
@@ -131,18 +94,13 @@ export const UserManagementPageContainer = () => {
...
-      onError: error => {
-        console.error(error)
       }
     }
   )
 
-  const handleCreateUser = (data: { uid: string; email: string; display_name: string }) => {
-    createUser({
+  const handleCreateUser: IDataHandlers['handleCreateUser'] = data => {
+    return promisifyMutation(createUser, {
```
knagurski, line 103: Instead of wrapping things in `promisifyMutation`, you can just the `mutateAsync` instead of `mutate` to get the promise-based version from Tanstack Query

https://tanstack.com/query/latest/docs/framework/react/guides/mutations#promises

apps/gitness/src/pages-v2/user-management/utils/promisify-mutation.ts
knagurski, line 1: I think we can kill this entirely and just use `mutateAsync` instead of `mutate`

packages/ui/locales/en/views.json
```diff
@@ -462,5 +464,78 @@
     "create": "Create webhook",
     "edit": "Edit webhook",
     "delete": "Delete webhook"
+  },
+  "userManagement": {
+    "addNewUser": "Add a new user",
+    "userId": "User ID",
+    "userIdHint": "User ID cannot be changed once created",
+    "enterUsername": "Enter user name",
```
knagurski, line 472: I'd argue that it's `username`, not `user name`
```suggestion
    "enterUsername": "Enter username",
```

packages/ui/src/views/user-management/components/dialogs/components/create-user/create-user.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+          {createUserError && <span className="text-xs text-destructive">{createUserError}</span>}
+        </div>
+
+        <Dialog.Footer>
+          <ButtonGroup className="justify-end">
+            <Button variant="outline" type="button" onClick={onClose} disabled={isCreatingUser}>
+              {t('views:userManagement.cancel', 'Cancel')}
+            </Button>
+            <Button type="button" onClick={handleSubmit(onSubmit)} disabled={isCreatingUser}>
```
knagurski, line 77: I don't know if the `Dialog` component allows this, but is it possible wrap the content in a form tag so that submission via keyboard would be possible? Instead of the submission being a click handler, it would then be set on the form and this button could then just be set as `type="submit"`.

```typescript jsx
<Dialog.Content>
  <form onSubmit={handleSubmit(onSubmit)}>
    <Dialog.Header>...</Dialog.Header>

    {/* form stuff */}

    <Dialog.Footer>
      <ButtonGroup>
        <Button onclick={onClose}>Cancel</Button>
        <Button type="submit">Invite user</Button>
      </ButtonGroup>
    </Dialog.Footer>
  </form>
</Dialog.Content>
```

Or something like that. Ideally most/all dialogs should be confirmable using enter and dismissible using escape without first having to focus those buttons.

packages/ui/src/views/user-management/components/dialogs/components/create-user/schemas.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod'
+
+export const newUserSchema = z.object({
+  uid: z.string().min(1, { message: 'Please provide a user ID' }),
+  email: z.string().email({ message: 'Please enter a valid email address' }),
+  display_name: z.string().min(1, { message: 'Please provide a display name' })
```
knagurski, line 6: Will these display? Or will they somehow be replaced by the translations?

packages/ui/src/views/user-management/components/dialogs/components/create-user/types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type NewUserFields = z.infer<typeof newUserSchema>
+
+export interface ICreateUserDialogProps {
+  handleCreateUser: (data: NewUserFields) => void
+  open: boolean
+  onClose: () => void
+}
+
+export interface ICreateUserFormProps {
```
knagurski, line 12: I'm really not a fan of the `I` prefix on interfaces. For me, the `Props` suffix is indication enough that it is a type.
```suggestion
export interface CreateUserDialogProps {
  handleCreateUser: (data: NewUserFields) => void
  open: boolean
  onClose: () => void
}

export interface CreateUserFormProps {
```

I also prefer when the component props interfaces live with their respective components. Easier to keep them in sync and see at a glance.

packages/ui/src/views/user-management/components/dialogs/components/edit-user/edit-user.tsx
```diff
@@ -0,0 +1,106 @@
...
+        userID: user.uid,
+        email: user.email,
+        displayName: user.display_name
+      })
+    }
+  }, [user, reset])
+
+  const onSubmit: SubmitHandler<EditUserFormType> = data => {
+    handleUpdateUser(data)
+  }
```
knagurski, line 46: Do we need this intermediate function? Could we just use `handleUpdateUser` directly? i.e.
```typescript jsx
<Button type="button" onClick={handleSubmit(handleUpdateUser)} disabled={isUpdatingUser}>
```

packages/ui/src/views/user-management/components/dialogs/components/remove-admin/remove-admin.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Dialog.Content className="max-w-xl">
+        <Dialog.Header>
+          <Dialog.Title className="font-medium">
+            {t('views:userManagement.updateAdminRights', 'Are you sure you want to remove {name} as an admin?', {
+              name: user?.display_name
+            })}
+          </Dialog.Title>
+        </Dialog.Header>
+
+        {/* TODO: investigate how to make user login bold with font-weight: 600 */}
```
knagurski, line 32: Not quite sure what "user login" is referring to 🤔

packages/ui/src/views/user-management/components/dialogs/components/reset-password/reset-password.tsx
```diff
@@ -0,0 +1,76 @@
...
+                )
+              : t(
+                  'views:userManagement.resetPasswordMessage',
+                  'A new password will be generated to assist {name} in resetting their current password.',
+                  { name: user?.display_name || '' }
+                )}
+          </span>
+
+          {generatePassword && (
+            <Input
```
knagurski, line 47: Just to confirm, the user who will be generating a password is unlikely to be the user who the password is for?

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-data.ts
```diff
@@ -0,0 +1,40 @@
+import { generateAlphaNumericHash } from '@/utils/utils'
+import { DialogLabels } from '@/views/user-management/components/dialogs'
+import { useDialogs } from '@/views/user-management/providers/dialogs-provider'
+import { useUserManagementStore } from '@/views/user-management/providers/store-provider'
+import { UsersProps } from '@/views/user-management/types'
+
+export const useDialogData = () => {
+  const { useAdminListUsersStore } = useUserManagementStore()
+
+  const { setUser, setPassword, setGeteneratePassword } = useAdminListUsersStore()
```
knagurski, line 10: Looks like there's a typo in the store?
```suggestion
  const { setUser, setPassword, setGeneratePassword } = useAdminListUsersStore()
```

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-data.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  const handleDialogOpen = (user: UsersProps | null, dialogType: DialogLabels) => {
+    prepareDialogData(user, dialogType)
+    openDialog(dialogType)
+  }
+
+  return {
+    handleDialogOpen
+  }
```
knagurski, line 39: Worth pointing out that this will not be dependency array safe. Easiest way is to wrap this whole bit in a `useMemo`
```suggestion
  return useMemo(() => {
    const prepareDialogData = (user: UsersProps | null, dialogType: DialogLabels) => {
      if (user) setUser(user)

      if (dialogType === DialogLabels.RESET_PASSWORD) {
        setGeteneratePassword(false)
        setPassword(generateAlphaNumericHash(10))

        return
      }

      if (dialogType === DialogLabels.CREATE_USER) {
        setGeteneratePassword(true)
        setPassword(generateAlphaNumericHash(10))

        return
      }
    }

    const handleDialogOpen = (user: UsersProps | null, dialogType: DialogLabels) => {
      prepareDialogData(user, dialogType)
      openDialog(dialogType)
    }

    return {
      handleDialogOpen
    }
  }, [setUser, setPassword, setGeteneratePassword, openDialog])
```

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-handlers.ts
```diff
@@ -0,0 +1,46 @@
+import { DialogLabels, IDialogHandlers } from '@/views/user-management/components/dialogs'
+import { useDialogs } from '@/views/user-management/providers/dialogs-provider'
+
+export const useDialogHandlers = (handlers: IDialogHandlers) => {
+  const { closeDialog } = useDialogs()
+
+  return {
+    handleCreateUser: async (...args: Parameters<typeof handlers.handleCreateUser>) => {
+      const result = await handlers.handleCreateUser(...args)
```
knagurski, line 9: What happens if these handlers fail?

packages/ui/src/views/user-management/components/main.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Content } from '@/views/user-management/components/page-components/content'
+import { useUserManagementStore } from '@/views/user-management/providers/store-provider'
+
+export const UserManagementPageContent = ({ handlers }: Pick<IUserManagementPageProps, 'handlers'>) => {
+  const { useAdminListUsersStore } = useUserManagementStore()
+
+  const { totalPages, page: currentPage, setPage } = useAdminListUsersStore()
+
+  return (
+    <SandboxLayout.Main className="max-w-[1092px]">
```
knagurski, line 12: 1092 is a very peculiarly specific number 🤔

apps/gitness/src/pages-v2/user-management/user-management-container.tsx
```diff
@@ -152,8 +109,9 @@ export const UserManagementPageContainer = () => {
     })
   }
 
-  const handleUpdateUser = (data: { email: string; displayName: string; userID: string }) => {
-    updateUser({
+  const handleUpdateUser = (data: IUpdateUserData) => {
+    console.log('data', data)
```
andrewgolovanov, line 113:
```suggestion

```

packages/ui/src/views/user-management/components/dialogs/components/create-user/create-user.tsx
andrewgolovanov: I think here we can encounter a problem, because the `Dialog` component has its own handlers that interact with the keyboard, for example, the same Esc just closes the dialog box, I also looked at how this case is in `shadcn`, and unfortunately there is also a problem with the operation of the Enter key :thinking_face: - https://ui.shadcn.com/docs/components/dialog

we have also checked in places where `Dialog` uses a form, and even in the case of wrapping in form it unfortunately does not solve the problem

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-handlers.ts
andrewgolovanov: in this case, the error object from `reactQuery` will be returned and displayed on the screen

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-data.ts
andrewgolovanov: Thanks for your comment! I’ve checked the current usage of `handleDialogOpen` and tried to evaluate the benefits of memoization. From what I see, it’s mostly used in static components that don’t expect frequent re-renders and aren’t memoized (e.g., `NoData` and others). As a result, these components will re-render on every render anyway, regardless of whether handleDialogOpen is stable or not. Am I missing something here? just want to clarify this for my own understanding. Thanks! :raised_hands:

packages/ui/src/views/user-management/components/dialogs/components/reset-password/reset-password.tsx
andrewgolovanov: Yes that's right, the admin can generate or change the password for any other user

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-data.ts
knagurski: The two most important things about memoization are to cut down on needless expensive processing (not applicable here) and to stabilise references. Since objects, arrays and functions are always passed by reference, if they are created in the flow of a react function component (as in the case of hooks), then they will be redefined on every render.

Often that doesn't really matter; it's inefficient, but ¯\\_(ツ)_/¯. However, if that function, object or array is used in the dependency array of something like a `useMemo` or `useEffect`, then it will trigger on _every_ render. So when passing functions, objects and arrays around, it's a good practice to stabilise their reference. For objects and arrays, `useMemo` is great for that and for functions, `useCallback` is ideal. In this case you're passing back an object with a single function, so you could go either way.

To demonstrate the use-case, the example below will trigger the `useEffect` on _every_ render.

```typescript jsx
function useMyAwesomeThing () {
  const getNumber = () => Math.random() * 1000

  return { getNumber }
}

function MyComponent () {
  const { getNumber } = useMyAwesomeThing()
  
  useEffect({
    alert(getNumber())
  }, [getNumber])

  return <p>Hello</p>
}
```

You can leave `getNumber` out of the dependency array, but ESlint will complain that you left it out and you'll have to disable the lint rule to get around it. And _most importantly_, it assumes you know exactly where and how that hook will be used throughout the codebase and that no one will ever use it in a dependency array.

You can fix the above by stabilising the reference. You can either stabilise the object you're returning...
```typescript jsx
function useMyAwesomeThing () {
  return useMemo(() => {
    const getNumber = () => Math.random() * 1000

    return { getNumber }
  }, [])
}
```
or the function.
```typescript jsx
function useMyAwesomeThing () {
  const getNumber = useCallback(() => Math.random() * 1000, [])

  return { getNumber }
}
```

Either of the above will make sure that `getNumber` is dependency array safe on every render.

In your case, this might not be an issue in the slightest. But it's a good habit to assume that you have no control over where your component/hook etc will be used and plan for the worst. Generally you'll hit less bugs and cut down on needless re-renders by being mindful of when you're creating and passing new functions, objects and arrays on each render.


I made up a simple CodePen a couple of years back to demonstrate. To use it, just uncomment the commented out lines one-by-one and you'll see the effect on the renders https://codepen.io/knagurski/pen/VwVWPQJ

knagurski (approved):
I don't see any blockers. Just a few suggestions.

packages/ui/src/views/user-management/components/dialogs/components/remove-admin/remove-admin.tsx
knagurski: You can use `dangerouslySetInnerHTML` to do that https://github.com/harness/canary/pull/1131#discussion_r1985851438

packages/ui/src/views/user-management/components/dialogs/hooks/use-dialog-data.ts
andrewgolovanov: Thanks @knagurski  for the detailed explanation and examples :heart:

packages/ui/src/views/user-management/components/dialogs/components/create-user/create-user.tsx
```diff
@@ -0,0 +1,105 @@
...
+      <Dialog.Content className="max-w-xl">
+        <Dialog.Header>
+          <Dialog.Title className="font-medium">
+            {t('views:userManagement.createUser.title', 'Add a new user')}
+          </Dialog.Title>
+        </Dialog.Header>
+
+        <FormWrapper onSubmit={handleSubmit(onSubmit)}>
+          <Fieldset>
+            <ControlGroup>
```
andrewgolovanov, line 51:
```suggestion
```
The `Input` component already contains a `ContentGroup` by default, so we don't need to add it manually 
image

packages/ui/src/views/user-management/components/dialogs/components/create-user/create-user.tsx
```diff
@@ -0,0 +1,105 @@
...
+                size="md"
+                {...register('uid')}
+                label={t('views:userManagement.userId', 'User ID')}
+                caption={t('views:userManagement.userIdHint', 'User ID cannot be changed once created')}
+                placeholder={t('views:userManagement.enterName', 'Enter name')}
+                error={errors.uid?.message?.toString()}
+              />
+            </ControlGroup>
+
+            <ControlGroup>
```
andrewgolovanov, line 63:
```suggestion
```
same

packages/ui/src/views/user-management/components/dialogs/components/create-user/create-user.tsx
```diff
@@ -0,0 +1,105 @@
...
+                type="email"
+                size="md"
+                {...register('email')}
+                placeholder={t('views:userManagement.enterEmail', 'Enter email address')}
+                label={t('views:userManagement.email', 'Email')}
+                error={errors.email?.message?.toString()}
+              />
+            </ControlGroup>
+
+            <ControlGroup>
```
andrewgolovanov, line 75:
```suggestion
```
same

packages/ui/src/views/user-management/components/dialogs/components/edit-user/edit-user.tsx
```diff
@@ -0,0 +1,125 @@
...
+                error={errors.userID?.message?.toString()}
+              />
+            </ControlGroup>
+          </Fieldset>
+
+          <Fieldset>
+            <ControlGroup>
+              <Input
+                id="email"
+                size="md"
```
andrewgolovanov, line 87:
```suggestion
                size="md"
                type="email"
```

packages/ui/src/views/user-management/components/dialogs/components/edit-user/edit-user.tsx
```diff
@@ -0,0 +1,125 @@
...
+              {
+                name: user?.display_name
+              }
+            )
+          }}
+        />
+
+        <FormWrapper onSubmit={handleSubmit(handleUpdateUser)}>
+          <Fieldset>
+            <ControlGroup>
```
andrewgolovanov, line 69:
```suggestion
```

same

packages/ui/src/views/user-management/components/dialogs/components/edit-user/edit-user.tsx
```diff
@@ -0,0 +1,125 @@
...
+                className="cursor-not-allowed"
+                label={t('views:userManagement.userId', 'User ID')}
+                caption={t('views:userManagement.userIdHint', 'User ID cannot be changed once created')}
+                error={errors.userID?.message?.toString()}
+              />
+            </ControlGroup>
+          </Fieldset>
+
+          <Fieldset>
+            <ControlGroup>
```
andrewgolovanov, line 84:
```suggestion
```
same

packages/ui/src/views/user-management/components/dialogs/components/edit-user/edit-user.tsx
```diff
@@ -0,0 +1,125 @@
...
+                size="md"
+                {...register('email')}
+                label={t('views:userManagement.editUser.email', 'Email')}
+                error={errors.email?.message?.toString()}
+              />
+            </ControlGroup>
+          </Fieldset>
+
+          <Fieldset>
+            <ControlGroup>
```
andrewgolovanov, line 96:
```suggestion
```

same

PR #1145, opened 2025-02-21
fix: refactor labels

3em:
Adjusted multiple components to the design
Added missing functionalities:
- remove label at PR page
- navigate to edit labels from the label dropdown
Fixed the Icon component size: in flex containers, it could shrink 

Design changes (before/after):

image

image

image

image

image

image

image


apps/gitness/src/pages-v2/repo/labels/hooks/use-full-fill-label-store.ts
abhinavrastogi-harness, line 1: i'm not sure I follow why this file/hook exists in the first place.
cc @cjlee01 @sans-harness can we discuss once?

apps/gitness/src/pages-v2/project/labels/hooks/use-get-project-label-and-values-data.ts
```diff
@@ -123,13 +128,19 @@ export const useGetProjectLabelAndValuesData = ({
...
-  }
+  /**
+   * Set loading state to store
+   */
+  useEffect(() => {
+    if (enabled) {
+      return setIsLoading(isLoadingSpaceLabels)
+    }
+
+    setIsLoading(isLoadingValues)
```
knagurski, line 142: It _probably_ isn't an issue, but it's typically a good idea to only return cleanup functions within a `useEffect`. It can be pretty easily rewritten in a few different ways, but personally I'd consider a ternary
```suggestion
    setIsLoading(enabled ? isLoadingSpaceLabels : isLoadingValues)
```

apps/gitness/src/pages-v2/repo/labels/hooks/use-full-fill-label-store.ts
sans-harness: For sure

apps/gitness/src/pages-v2/project/labels/project-labels-list-container.tsx
```diff
@@ -24,17 +24,16 @@ export const ProjectLabelsList = () => {
   const { queryPage } = usePaginationQueryStateWithStore({ page, setPage })
   const [query, setQuery] = useQueryState('query')
 
-  const { isLoading } = useGetProjectLabelAndValuesData({ queryPage, query })
+  // To fetch labels/values and set isLoading state at useLabelsStore
+  useGetProjectLabelAndValuesData({ queryPage, query })
```
knagurski, line 28: I'm not really a fan of fire-and-forget hooks that are named as accessors. It's not immediately obvious that this kicks something off in the background; reading this, I would assume it was a mistake and could be deleted without side effect.

Would it make sense to break out the functionality you need into a separate hook, with more appropriate naming, that can be used here as well in `useGetProjectLabelAndValuesData`?

apps/gitness/src/pages-v2/project/stores/labels-store.tsx
knagurski, line 1: Not an issue (and not introduced in this PR), but since there's no JSX in this file, it should be a `.ts` file, not `.tsx`

apps/gitness/src/pages-v2/repo/labels/hooks/use-full-fill-label-store.ts
```diff
@@ -57,9 +59,16 @@ export const useFullFillLabelStore = ({ queryPage, query, enabled = true }: UseF
...
-  }
+  /**
+   * Set loading state to store
+   */
+  useEffect(() => {
+    if (enabled) {
+      return setIsLoading(isDataLoading)
+    }
+
+    setIsLoading(false)
```
knagurski, line 70: Best to only ever return teardown functions from a `useEffect`. You can simplify this by not returning early
```suggestion
    setIsLoading(enabled && isDataLoading)
```

apps/gitness/src/pages-v2/repo/labels/label-form-container.tsx
```diff
@@ -14,52 +14,31 @@ export const RepoLabelFormContainer = () => {
...
-  const { isLoading: isDataLoading, repo_ref } = useFullFillLabelStore({
-    query: labelId,
-    enabled: !!labelId
-  })
+  const { repo_ref } = useFullFillLabelStore({ query: labelId, enabled: !!labelId })
+
+  const onFormCancel = () => navigate(routes.toRepoLabels({ spaceId, repoId }))
 
   const {
     mutate,
```
knagurski, line 22: Not an issue, just a "while you're there" kinda thing. It's usually a good idea to alias the `mutate` so that it's obvious what it does when calling it. If it's called `mutate`, you have to scroll back up to the top to figure out what it does. If it's called `saveRepoLabel`, it's immediately obvious.

apps/gitness/src/pages-v2/repo/labels/label-form-container.tsx
```diff
@@ -14,52 +14,31 @@ export const RepoLabelFormContainer = () => {
...
-        },
-        values
-      }
-    })
+    mutate({ body: { label: { ...rest }, values } })
   }
 
   return (
     <LabelFormPage
+      className="w-[570px] px-0"
```
knagurski, line 35: 570 seems super arbitrary to me 🤔

packages/ui/src/components/icon.tsx
```diff
@@ -424,7 +424,14 @@ export interface IconProps {
 
 const Icon: React.FC<IconProps> = ({ name, size = 16, height, width, className }) => {
   const Component = IconNameMap[name]
-  return <Component width={width || size} height={height || size} className={className} />
+  return (
+    <Component
+      width={width || size}
+      height={height || size}
+      className={className}
+      style={{ minWidth: `${width || size}px`, minHeight: `${height || size}px` }}
```
knagurski, line 432: Just curious why this was needed?

packages/ui/src/components/label-marker.tsx
```diff
@@ -25,23 +29,71 @@ export const LabelMarkerColor = {
...
-      )}
-    >
-      <span className={cn('truncate', { 'flex-1': inTable }, { 'pr-1.5': !!value })}>{label}</span>
-      {!!value && (
-        <span
-          className={`flex h-full items-center overflow-hidden rounded-r bg-label-background-black pl-1.5 pr-[7px] ${inTable && 'max-w-[50%]'}`}
+    <div className={cn('flex max-w-full items-center gap-2 opa relative', className)}>
+      <span
+        className={cn(
+          `grid text-12 leading-4 h-5 py-px px-2 items-center font-medium rounded overflow-hidden w-fit ${LabelMarkerColor[color]}`,
```
knagurski, line 65: Probably easier to let `cn` figure this out instead of using string interpolation. 
```suggestion
          'grid text-12 leading-4 h-5 py-px px-2 items-center font-medium rounded overflow-hidden w-fit',
          LabelMarkerColor[color],
```

packages/ui/src/components/label-marker.tsx
```diff
@@ -25,23 +29,71 @@ export const LabelMarkerColor = {
...
+      {isWithDeleteButton && (
+        <Button
+          className={`absolute right-1 top-0.5 max-h-4 min-h-4 min-w-4 max-w-4 px-0 ${deleteButtonColor[color]}`}
+          variant="custom"
+          onClick={onDelete}
         >
-          <span className="max-w-full truncate">{value}</span>
-        </span>
+          <Icon name="cross" size={10} />
+        </Button>
```
knagurski, line 93: Can we add something here for screen readers to pick up? Either an `aria-label` on the `Button`, or a string as a sibling of the `Icon` with a class of `sr-only`?

packages/ui/src/components/label-marker.tsx
```diff
@@ -25,23 +29,71 @@ export const LabelMarkerColor = {
...
       )}
-    </span>
+
+      {isDynamic && <Icon name="circle-plus" size={12} className="min-w-3 text-icons-4" />}
```
knagurski, line 96: Does this have any meaning that as screen reader should be mad aware of?

packages/ui/src/components/search-box.tsx
```diff
@@ -115,17 +104,18 @@ const Root = forwardRef<HTMLInputElement, SearchBoxProps>(
     return (
       <div className={cn('relative', width === 'full' ? 'w-full' : 'w-96', className)}>
         {hasSearchIcon && (
-          <Icon name="search" size={12} className="absolute left-2.5 top-1/2 -translate-y-1/2 text-icons-1" />
+          <Icon name="search" size={12} className="absolute left-2.5 top-1/2 z-10 -translate-y-1/2 text-icons-1" />
         )}
+
         {hasShortcut && !!shortcutLetter && (
           <div className="absolute right-1.5 top-1/2 flex h-5 -translate-y-1/2 cursor-pointer items-center gap-0.5 rounded-sm border bg-background-3 px-1 text-foreground-2 duration-100 ease-in-out">
             <Icon name="apple-shortcut" size={12} />
```
knagurski, line 112: This is going to annoy everyone who's not on a Mac.

packages/ui/src/views/labels/components/label-cell-content.tsx
```diff
@@ -1,45 +1,46 @@
...
 export interface LabelCellContentProps {
   label: ILabelType
-  values: LabelValuesType
+  values?: LabelValueType[]
 }
 
 export const LabelCellContent: FC<LabelCellContentProps> = ({ label, values }) => {
-  const value = values?.[label.key]
-  const length = value ? value.length : 0
+  const isWithValues = values?.length && values.length > 0
```
knagurski, line 13: Can be simplified a bit. Not a blocker in any way
```suggestion
  const isWithValues = !!values?.length
```

packages/ui/src/views/labels/components/label-form-color-and-name-group.tsx
```diff
@@ -1,8 +1,7 @@
 import { FC } from 'react'
-import { UseFormRegister } from 'react-hook-form'
 
-import { Button, Icon, Input, Select } from '@/components'
-import { ColorsEnum, CreateLabelFormFields, TranslationStore } from '@/views'
+import { Button, Icon, Input, InputProps, Select, SelectRootProps } from '@/components'
+import { cn, ColorsEnum, TranslationStore } from '@/views'
```
knagurski, line 4: Has `cn` been accidentally added to the views? This should only be importable from `@utils`.

packages/ui/src/views/labels/components/label-form-color-and-name-group.tsx
```diff
@@ -21,62 +20,60 @@ const SelectColorMarker = {
...
+        isWithDeleteButton && 'grid-cols-[8rem_1fr_auto]',
+        className
+      )}
+    >
+      <Select.Root {...selectProps}>
+        <Select.Content>
+          {Object.values(ColorsEnum).map(color => (
+            <Select.Item key={color} value={color}>
+              <div className="flex max-w-full items-center gap-x-1.5">
+                <div className={`size-2 min-h-2 min-w-2 rounded-full ${SelectColorMarker[color]}`} />
```
knagurski, line 56:
```suggestion
                <div className={cn('size-2 min-h-2 min-w-2 rounded-full', SelectColorMarker[color])} />
```

packages/ui/src/views/labels/components/labels-list-view.tsx
```diff
@@ -55,31 +58,33 @@ export const LabelsListView: FC<LabelsListViewProps> = ({
...
             <span className="pl-[22px]">{t('views:labelData.table.name', 'Name')}</span>
           </Table.Head>
-          <Table.Head>{t('views:labelData.table.created', 'Created in')}</Table.Head>
-          <Table.Head>{t('views:labelData.table.description', 'Description')}</Table.Head>
-          <Table.Head />
+          <Table.Head className="w-1/4">{t('views:labelData.table.created', 'Created in')}</Table.Head>
+          <Table.Head className={cn('w-1/2', { 'w-5/12': isSmallWidth })}>
+            {t('views:labelData.table.description', 'Description')}
+          </Table.Head>
+          <Table.Head className="w-[54px]"></Table.Head>
```
knagurski, line 79: Does a self-closing tag work here?
```suggestion
          <Table.Head className="w-[54px]" />
```

packages/ui/src/views/labels/labels-list-page.tsx
```diff
@@ -61,62 +62,55 @@ export const LabelsListPage: FC<LabelsListPageProps> = ({
...
-              />
+              <Checkbox checked={getParentScopeLabels} onCheckedChange={setGetParentScopeLabels} id="parent-labels" />
             }
-            id="show-parent-labels"
+            id="parent-labels"
             label={t('views:labelData.showParentLabels', 'Show labels from parent scopes')}
           />
         )}
 
         {(!!spaceLabels.length || (!spaceLabels.length && isDirtyList)) && (
```
knagurski, line 79: Am I crazy for thinking this could just be...
```suggestion
        {(!!spaceLabels.length || isDirtyList) && (
```

🤔

packages/ui/src/components/label-marker.tsx
```diff
@@ -25,23 +29,71 @@ export const LabelMarkerColor = {
...
+        {isWithExtraContent && (
+          <span
+            className={cn(
+              `bg-label-background-black flex h-full items-center overflow-hidden rounded-r-[3px] pl-1.5 pr-[7px]`,
+              { 'pr-6': isWithExtraContent && isWithDeleteButton }
+            )}
+          >
+            {!!value && <span className="max-w-full truncate">{value}</span>}
+
+            {!!counter && counter}
```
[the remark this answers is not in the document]
3em: yes you are right, 0 not allowed

apps/gitness/src/pages-v2/repo/labels/hooks/use-full-fill-label-store.ts
3em: This hook was created to populate data in the store (`useLabelsStore`). It's used on the labels list page and on the repository label details page. Essentially, it's just isolated code moved into a hook to fill the store, avoiding repetitive code.

apps/gitness/src/pages-v2/project/labels/project-labels-list-container.tsx
3em: We have renamed this hook to a more understandable name – `useFillLabelStoreWithProjectLabelValuesData`

apps/gitness/src/pages-v2/project/labels/project-labels-list-container.tsx
```diff
@@ -48,22 +47,19 @@ export const ProjectLabelsList = () => {
...
-    deleteSpaceLabel({
-      key: identifier
-    })
+    deleteSpaceLabel({ key: identifier })
   }
 
   return (
     <>
       <LabelsListPage
+        className="mx-auto max-w-[812px]"
```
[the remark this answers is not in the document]
3em: We discussed it with the designers and made the container wider

packages/ui/src/components/icon.tsx
3em: The Icon component is often used in layouts in a way that unintentionally reduces its width (e.g., due to flex), which is incorrect behavior. This code ensures that the icon maintains its width in any layout.

packages/ui/src/components/search-box.tsx
knagurski: Yeah, totally, just an observation when we were in the area

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
```diff
@@ -227,33 +227,30 @@ export default function PullRequestConversationPage() {
...
-            }
-          } else {
-            setShowRestoreBranchButton(true)
-          }
-        })
-      }
+    if (!branchError) return
+
+    if (pullReqMetadata?.merged || pullReqMetadata?.closed) {
+      return setShowRestoreBranchButton(true)
```
knagurski, line 233: Best to only return cleanup functions from a `useEffect`. This will be fine, but it's probably best to separate the call and the return since they aren't really connected.
```suggestion
      setShowRestoreBranchButton(true)
      return
```

packages/ui/src/components/label-marker.tsx
```diff
@@ -1,16 +1,20 @@
 import { FC } from 'react'
 
+import { Button, Icon } from '@/components'
 import { cn } from '@utils/cn'
-import { ColorsEnum } from '@views/labels'
+import { ColorsEnum, ILabelType, LabelType } from '@views/labels'
```
knagurski, line 5: Can we find a way to avoid references to the `views` from within `components`? Ideally this should be one-way to make it clear that views bring components together into a cohesive interface, leaving components as independent building blocks. Perhaps should `LabelMarker` be a subcomponent of the Labels view?

packages/ui/src/components/label-marker.tsx
```diff
@@ -25,23 +29,73 @@ export const LabelMarkerColor = {
...
-        `inline-flex text-12 leading-5 h-5 py-px px-2 items-center font-medium rounded overflow-hidden max-w-full ${LabelMarkerColor[color]}`,
-        { 'pr-px': !!value }
-      )}
-    >
-      <span className={cn('truncate', { 'flex-1': inTable }, { 'pr-1.5': !!value })}>{label}</span>
-      {!!value && (
-        <span
-          className={`flex h-full items-center overflow-hidden rounded-r bg-label-background-black pl-1.5 pr-[7px] ${inTable && 'max-w-[50%]'}`}
+    <div className={cn('flex max-w-full items-center gap-2 opa relative', className)}>
+      <span
```
knagurski, line 63: This being a `span` (an inline element), but being set as display grid (a block level element) doesn't really make much sense to me. Is there a reason we can't use a `div`? Makes no difference, but I'm curious if there's a reason as I've seen inline elements being changed to block a lot in this codebase 🤔

packages/ui/src/components/label-marker.tsx
```diff
@@ -25,23 +29,73 @@ export const LabelMarkerColor = {
...
+  label,
+  value,
+  color,
+  counter,
+  onDelete,
+  className
+}) => {
+  const isDynamic = type === LabelType.DYNAMIC
+  const isWithExtraContent = !!value || !!counter
+  const isWithDeleteButton = !!onDelete
```
knagurski, line 59: You can totally ignore this since it makes zero difference... instead of `isWith`, `has` reads better IMO
```suggestion
  const hasExtraContent = !!value || !!counter
  const hasDeleteButton = !!onDelete
```

packages/ui/src/components/table.tsx
```diff
@@ -43,7 +43,7 @@ TableHeader.displayName = 'TableHeader'
 
 const TableBody = forwardRef<HTMLTableSectionElement, HTMLAttributes<HTMLTableSectionElement>>(
   ({ className, ...props }, ref) => (
-    <tbody ref={ref} className={cn('[&_tr:last-child]:border-0', className)} {...props} />
+    <tbody ref={ref} className={cn('[&_tr:last-child]:border-0 [&>tr:hover]:bg-background-4', className)} {...props} />
```
knagurski, line 46: To me a change of colour on hover denotes that something is clickable. This might be the ideal for the usage in the labels work, but I don't think we can make that call for _all_ tables. Can we make this opt-in as, perhaps, an extra prop?

packages/ui/src/views/repo/pull-request/components/labels/pull-request-labels-header.tsx
```diff
@@ -11,9 +12,18 @@ import {
...
+import { debounce } from 'lodash-es'
 
 import { LabelValueSelector } from './label-value-selector'
 
+const getSelectedValueId = (label?: LabelAssignmentType) => {
+  if (!!label && 'assigned_value' in label) {
+    return label.assigned_value?.id ?? undefined
+  }
+
+  return undefined
```
knagurski, line 24: No need to manually return `undefined`; if you don't return anything, you'll get the same
```suggestion
```

praneshg239 (approved):

PR #1151, opened 2025-02-21
Unit tests ts jest

sans-harness:


packages/ui/package.json
```diff
@@ -34,12 +34,14 @@
...
     "pre-commit": "lint-staged",
-    "test": "vitest --run",
+    "test": "jest",
     "test:coverage": "vitest --run --coverage",
     "test:watch": "vitest",
     "test:changed": "vitest --run --changed=origin/main",
     "typecheck": "tsc -b",
-    "extract": "i18next --config './i18n.config.ts'"
+    "extract": "i18next --config './i18n.config.ts'",
+    "jest-test": "jest src",
```
abhinavrastogi-harness, line 43: we can remove these

packages/ui/src/components/accordion/__tests__/accordion.test.tsx
```diff
@@ -0,0 +1,22 @@
+import { render, screen } from '@testing-library/react'
+
+import '@testing-library/jest-dom'
+
+import { Accordion } from '../accordion'
+
+describe('Accordion Component', () => {
+  test('renders the accordion with trigger and content', () => {
+    render(
```
abhinavrastogi-harness, line 9: better way to do it:

```suggestion
    const { getByText } = render(
```

packages/ui/src/components/index.ts
```diff
@@ -8,7 +8,7 @@ export * from './drawer'
 export * from './text'
 export * from './spacer'
 export * from './pagination/pagination'
-export * from './badge'
+export * from './badge/badge'
```
abhinavrastogi-harness, line 11: instead of this, can we move the tests to `src/components/__tests__`?

abhinavrastogi-harness (approved):

packages/ui/src/components/accordion/__tests__/accordion.test.tsx
srdjan-harness: Could be better to use screen. https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

PR #1152, opened 2025-02-22
feat: branch state management with namespaced store

andrewgolovanov:
**Description:**
This merge request fixes the search functionality for the `Branches` and `Tags` pages, as well as refactoring the `Branch Selector` component and removing `Branch Chooser`

Before:

https://github.com/user-attachments/assets/1087ba83-6c13-40be-9593-d26ecf910856

After:

https://github.com/user-attachments/assets/6a680628-cb34-46ac-9570-64e8ffc4c2fb


apps/gitness/src/pages-v2/repo/stores/repo-branches-store.ts
```diff
@@ -1,35 +1,111 @@
...
+    setSelectedBranchTag: (branch: BranchSelectorListItem) =>
+      baseStore.setState(namespace, { selectedBranchTag: branch }),
+    setSelectedRefType: (refType: BranchSelectorTab) => baseStore.setState(namespace, { selectedRefType: refType }),
+    setTagList: (tags: BranchSelectorListItem[]) => baseStore.setState(namespace, { tagList: tags }),
+    setPage: (page: number) => baseStore.setState(namespace, { page }),
+    setPaginationFromHeaders: (xNextPage: number, xPrevPage: number) =>
+      baseStore.setState(namespace, { xNextPage, xPrevPage }),
+    setSpaceIdAndRepoId: (spaceId: string, repoId: string) => baseStore.setState(namespace, { spaceId, repoId })
+  }
+}
```
sans-harness, line 111: I don't think this is the right approach. To avoid state conflicts, it would be better to adopt a `view-model` pattern, where each view that requires its own state can use an isolated store, handle its own API calls, and manage its state independently.

The current approach increases complexity within a shared store, making it more difficult for other developers to contribute and use it effectively. IMHO, this isn't a scalable solution.

PR #1156, opened 2025-02-24
fix: remove jest and fix vitest config for react 17

knagurski:
This PR removes Jest and fixes the Vitest config. This PR also adds a resolution for `monaco-editor` for use when referenced in a test and when running test coverage. This PR also extends the `Accordion` component tests and normalises some other component tests.


praneshg239 (approved):

packages/ui/src/components/accordion/__tests__/accordion.test.tsx
```diff
@@ -1,22 +1,105 @@
...
-      <Accordion.Root type="single">
-        <Accordion.Item value="item-1">
-          <Accordion.Trigger>Trigger</Accordion.Trigger>
-          <Accordion.Content>Content</Accordion.Content>
-        </Accordion.Item>
-      </Accordion.Root>
-    )
+const renderComponent = (props: Partial<ComponentProps<typeof Accordion.Root>> = {}): RenderResult =>
+  render(
+    <Accordion.Root type="single" {...(props as any)}>
```
knagurski, line 12: Had to do an `as any` since this component has branching types for `single` and `multiple` types. At this point we're already protected by the types when calling `renderComponent`, so we're safe to switch off type checks for this statement.

packages/ui/src/components/alert/Alert.test.tsx
```diff
@@ -16,7 +16,7 @@ describe('Alert', () => {
       </Alert.Container>
     )
 
-    expect(screen.getByText(title)).toBeInTheDocument()
-    expect(screen.getByText(description)).toBeInTheDocument()
+    expect(await screen.findByRole('heading', { name: title })).toBeInTheDocument()
```
knagurski, line 19: Best to always opt for `ByRole` when you can and only fall back to `ByText` when there's nothing else to grab on to.

packages/ui/src/components/accordion/__tests__/accordion.test.tsx
abhinavrastogi-harness: can we disable the lint rule for one line, instead of disabling for the whole file?

abhinavrastogi-harness (approved):

packages/ui/src/components/accordion/__tests__/accordion.test.tsx
knagurski: I was actually going to mirror some other projects we have an disable it for all `.test.ts(x)` files. Those rules are crucial for actual code, but don't make much sense for test files.

PR #1161, opened 2025-02-26
Unified pipeline studio view component

srdjan-harness:


packages/ui/src/views/unified-pipeline-studio/components/palette-drawer/components/step-palette-layout.tsx
```diff
@@ -0,0 +1,15 @@
+const StepsPaletteLayout = {
```
abhinavrastogi-harness, line 1: I doubt if we need so much componentisation for every single div. 
this will not be reused beyond 1 or (maybe) 2 places.

packages/ui/src/views/unified-pipeline-studio/components/palette-drawer/components/step-palette-section.tsx
```diff
@@ -0,0 +1,45 @@
...
+                  <StepsPaletteItemLayout.Title>{step.identifier}</StepsPaletteItemLayout.Title>
+                  {/* <StepsPaletteItem.BadgeWrapper>verified</StepsPaletteItem.BadgeWrapper> */}
+                </StepsPaletteItemLayout.Header>
+                <StepsPaletteItemLayout.Description>{step.description}</StepsPaletteItemLayout.Description>
+              </StepsPaletteItemLayout.Right>
+            </StepsPaletteItemLayout.Root>
+          </StepsPaletteContentLayout.SectionItem>
+        ))
+      ) : (
+        <p className="text-muted-foreground">There is no steps for provided query.</p>
```
abhinavrastogi-harness, line 41: i18n?

packages/ui/src/views/unified-pipeline-studio/components/panel/unified-pipeline-studio-problems-panel.tsx
```diff
@@ -0,0 +1,23 @@
+import { NoProblemsFound, Problems } from '@components/index'
+
+import { useYamlEditorContext } from '@harnessio/yaml-editor'
+
+export function UnifiedPipelineStudioProblemsPanel({ problems }: { problems: any }): JSX.Element {
+  const { updateCursorPosition } = useYamlEditorContext()
+
+  return problems.length === 0 ? (
```
abhinavrastogi-harness, line 8: do we need this wrapper just for this condition?
can we not just return `NoProblemsFound` from `<Problems />` if the passed array is empty?

packages/ui/src/views/unified-pipeline-studio/components/form-inputs/ArrayInput.tsx
shaurya-harness, line 1: should we move these components out of unified-pipeline-studio to a view of their own?
would be reused by any component built on top of forms framework?

packages/ui/src/views/unified-pipeline-studio/components/palette-drawer/components/step-form-layout.tsx
shaurya-harness, line 1: same for this? could be a re-usable layout that can be used by other right-drawers?

shaurya-harness (approved):

PR #1162, opened 2025-02-26
chore: remove deprecated status from Text component

andrewgolovanov:
**Description:**
This pull request removes the `deprecated` status of the `Text` component as there is currently no alternative approach available


knagurski (approved):
👍

PR #1163, opened 2025-02-26
Design tokens editing

cxramina:
Updated version of design tokens.

- Converted all color tokens to be component-specific instead of groups.
- Added gradient tokens for expanded styling options.
- Updated core size values from rem to px, while still keeping rem for typography.


knagurski (approved):
👍🏻

PR #1165, opened 2025-02-27
Secrets view

sans-harness:
https://github.com/user-attachments/assets/5e8104f8-e148-4aae-b2e5-0d1dab81e4f8


apps/design-system/src/subjects/views/secrets/secrets.tsx
```diff
@@ -0,0 +1,50 @@
...
+        )
+      case SecretType.Existing:
+        return <div>Existing Secret Content</div>
+      default:
+        return null
+    }
+  }
+
+  return (
+    <Drawer.Root>
```
abhinavrastogi-harness, line 34: lets add `direction="right"` prop

packages/ui/src/views/secrets/new-secret/new-secret-form.tsx
```diff
@@ -0,0 +1,137 @@
...
+  Spacer,
+  Textarea
+} from '@/components'
+import { SandboxLayout, TranslationStore } from '@/views'
+import { zodResolver } from '@hookform/resolvers/zod'
+import { z } from 'zod'
+
+const newSecretformSchema = z.object({
+  name: z.string().min(1, { message: 'Please provide a name' }),
+  password: z.string().min(1, { message: 'Please provide a password' }),
```
abhinavrastogi-harness, line 20: change both key and message to `value` instead of `password`

packages/ui/src/views/secrets/new-secret/new-secret-form.tsx
```diff
@@ -0,0 +1,137 @@
...
+  description: z.string().optional(),
+  tags: z.string().optional()
+})
+
+export type NewSecretFormFields = z.infer<typeof newSecretformSchema> // Automatically generate a type from the schema
+
+interface NewSecretPageProps {
```
abhinavrastogi-harness, line 27: let's call it `CreateSecret` everywhere.. "new" can be confused with versioning (vs "old")

packages/ui/src/views/secrets/new-secret/new-secret-form.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  size="md"
+                  error={errors.tags?.message?.toString()}
+                  optional
+                />
+              </Fieldset>
+            </Accordion.Content>
+          </Accordion.Item>
+        </Accordion.Root>
+        {/* SUBMIT BUTTONS */}
+        <Fieldset className="mt-auto">
```
abhinavrastogi-harness, line 122: let's move the buttons to the footer, with a background similar to DialogFooter.

packages/ui/src/views/secrets/new-secret/new-secret-form.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  label="Tags"
+                  placeholder="Enter tags"
+                  size="md"
+                  error={errors.tags?.message?.toString()}
+                  optional
+                />
+              </Fieldset>
+            </Accordion.Content>
+          </Accordion.Item>
+        </Accordion.Root>
```
abhinavrastogi-harness, line 120: let's add a way to show errors on the view, if secret creation fails

packages/ui/src/views/secrets/secrets-header.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <StackedList.Root className="overflow-hidden" borderBackground>
+        <StackedList.Item
+          className={cn('cursor-pointer !rounded px-5 py-3', {
+            '!bg-background-4': selectedType === SecretType.New
+          })}
+          isHeader
+          isLast
+          onClick={() => setSelectedType(SecretType.New)}
+          actions={
+            <Checkbox
```
abhinavrastogi-harness, line 27: hmm.. few things here:
1. can this be implemented as a form field to avoid maintaining state?
2. can this be a radio button instead of checkbox?

sans-harness: good point.

abhinavrastogi-harness (approved):

PR #1167, opened 2025-02-27
Replace Navbar with Sidebar

vivek-harness:
Replaces the current Navbar with latest Sidebar from Shadcn.


packages/ui/src/components/navbar-new/sidebar-search-new/sidebar-search-new.tsx
```diff
@@ -0,0 +1,56 @@
...
+    console.warn('⚠️ Search context is null, returning early.')
+    return null
+  }
+
+  const { setIsOpen } = searchContext
+
+  return (
+    <form {...props}>
+      {props.logo}
+      <Sidebar.Group className="py-0 -mt-2 px-2">
```
andrewgolovanov, line 27:
```suggestion
      <Sidebar.Group className="py-0">
```
1. it seems we can get rid of `-mt-2` which breaks the height of the form parent, if we need to get closer to the logo we can just change its current height

2. We also do not need to specify `px-2`, because initially Sidebar.Group has already set the necessary value for it.

image
image

packages/ui/src/components/navbar-new/app-sidebar.tsx
```diff
@@ -0,0 +1,219 @@
...
+                <Link className="flex items-center gap-1.5" to="/">
+                  <Icon name="harness" size={18} className="text-foreground-accent" />
+                  <Icon name="harness-logo-text" width={65} height={15} className="mb-0.5 text-foreground-1" />
+                </Link>
+              }
+            />
+          )}
+        </Sidebar.Header>
+        <Sidebar.Content>
+          <Sidebar.Group className="px-4 pt-5">
```
andrewgolovanov, line 98:
```suggestion
          <Sidebar.Group className="px-2 pt-5">
```
image

packages/ui/src/components/navbar-new/app-sidebar.tsx
```diff
@@ -0,0 +1,219 @@
...
+                      <span className="text-foreground-3 font-medium transition-colors hover:text-primary">More</span>
+                    </div>
+                  </Sidebar.MenuButton>
+                </Sidebar.MenuItem>
+              </Sidebar.Menu>
+            </Sidebar.GroupContent>
+          </Sidebar.Group>
+
+          {!!recentMenuItems.length && (
+            <Sidebar.Group title="Recent" className="border-t px-4 pt-3">
```
andrewgolovanov, line 129:
```suggestion
            <Sidebar.Group title="Recent" className="border-t px-2 pt-3">
```

same

packages/ui/src/components/navbar-new/app-sidebar.tsx
```diff
@@ -0,0 +1,219 @@
...
+                      handleCustomNav={handleCustomNav}
+                      t={t}
+                    />
+                  ))}
+                </Sidebar.Menu>
+              </Sidebar.GroupContent>
+            </Sidebar.Group>
+          )}
+
+          <Sidebar.Group className="border-t px-4 pt-5">
```
andrewgolovanov, line 150:
```suggestion
          <Sidebar.Group className="border-t px-2 pt-5">
```

same

packages/ui/src/components/navbar-new/app-sidebar.tsx
```diff
@@ -0,0 +1,219 @@
...
+            <Sidebar.GroupContent>
+              <Sidebar.Menu>
+                {!!currentUser?.admin && (
+                  <Sidebar.MenuItem>
+                    <Sidebar.MenuButton
+                      asChild
+                      className="cursor-pointer"
```
andrewgolovanov, line 157:
```suggestion

```
I think we can set this class as the default for this component, as it is interactive in nature, so we don't have to pass it every time.

packages/ui/src/components/navbar-new/app-sidebar.tsx
```diff
@@ -0,0 +1,219 @@
...
+                      onClick={() => navigate('/admin/default-settings')}
+                    >
+                      <div>
+                        <Icon
+                          name="account"
+                          size={12}
+                          className="text-icons-4 transition-colors hover:text-primary !w-[12px] !h-[auto]"
```
andrewgolovanov, line 164:
```suggestion
                          className="text-icons-4 transition-colors hover:text-primary"
```

I would prefer to get rid of the `[&>svg]:size-4` class initially in the parent `SidebarMenuButton` component, as this approach does not work in our case, since we pass the svg through the Icon component to which we specify the required sizes, and in the current situation we have to overwrite each time the value set in the parent, so it does not seem to make any sense

packages/ui/src/components/navbar-new/sidebar-item.tsx
```diff
@@ -0,0 +1,131 @@
...
+      ) : null}
+    </>
+  )
+
+  return (
+    <div className="group/menu-item relative">
+      <NavLink className="block" to={item.to || ''} end>
+        {({ isActive }) => (
+          <Sidebar.MenuItem
+            className={cn('hover:bg-background-4 rounded text-foreground-3  transition-colors hover:text-primary', {
```
andrewgolovanov, line 88:
```suggestion
            className={cn('hover:bg-background-4 rounded text-foreground-3 transition-colors hover:text-primary', {
```

extra space

packages/ui/src/components/navbar-new/sidebar-item.tsx
```diff
@@ -0,0 +1,131 @@
...
+              'bg-background-4': isActive
+            })}
+          >
+            <Sidebar.MenuButton asChild isActive={isActive}>
+              <div className="gap-2.5">
+                <Icon
+                  name={iconName!}
+                  size={10}
+                  className={cn('!w-[12px] !h-[auto]', {
```
andrewgolovanov, line 97:
```suggestion
                  className={cn( {
```

similar [story](https://github.com/harness/canary/pull/1167/files#r1975693111)

packages/ui/src/components/navbar-new/sidebar-item.tsx
```diff
@@ -0,0 +1,129 @@
...
+            {t('component:navbar.unpin')}
+          </Text>
+        </DropdownMenu.Item>
+      ) : null}
+    </>
+  )
+
+  return (
+    <Sidebar.MenuItem>
+      <div className="group/menu-item relative">
```
andrewgolovanov, line 84:
```suggestion
```

looks like we can get rid of that div too, since `MenuItem` already contains these classes
image

packages/ui/src/components/navbar-new/sidebar-item.tsx
```diff
@@ -0,0 +1,129 @@
...
+                </div>
+              </Sidebar.MenuButton>
+            </div>
+          )}
+        </NavLink>
+
+        <DropdownMenu.Root>
+          <Sidebar.MenuAction>
+            <DropdownMenu.Trigger asChild>
+              <span className="absolute right-[-0.2rem] text-icons-4 opacity-0 hover:text-icons-2 focus:opacity-100 focus-visible:opacity-100 group-hover/menu-item:opacity-100 data-[state=open]:opacity-100 focus:outline-none focus:ring-0">
```
andrewgolovanov, line 117:
```suggestion

```

why don't we also pass the necessary classes directly through `DropdownMenu.Trigger` then in this case we don't need to specify the `span` for it


```jsx
<DropdownMenu.Trigger className="...">
 ...
 ```

File: packages/ui/src/components/sidebar/sidebar.tsx [hunk 11]
```diff
@@ -455,7 +455,7 @@ const SidebarMenuButton = forwardRef<
       data-sidebar="menu-button"
       data-size={size}
       data-active={isActive}
-      className={cn(sidebarMenuButtonVariants({ variant, size }), className)}
+      className={cn(sidebarMenuButtonVariants({ variant, size }), 'cursor-pointer', className)}
```
andrewgolovanov, line 458:
```suggestion
      className={cn(sidebarMenuButtonVariants({ variant, size }), className)}
```
I think it is more correct to specify this in the base styles in the `sidebarMenuButtonVariants` constant

```jsx 
const sidebarMenuButtonVariants = cva('... cursor-pointer', { ...

File: packages/ui/src/components/navbar-new/sidebar-item.tsx (hunk 10 above)
Comment (vivek-harness) [reply packages/ui/src/components/navbar-new/sidebar-item.tsx]: Passing it directly is breaking the UI. So kept using span for now.

File: packages/ui/src/components/navbar-new/app-sidebar.tsx [hunk 12]
```diff
@@ -0,0 +1,203 @@
...
+          <Sidebar.Group className="border-t px-2 pt-5">
+            <Sidebar.GroupContent>
+              <Sidebar.Menu>
+                {!!currentUser?.admin && (
+                  <Sidebar.MenuItem>
+                    <Sidebar.MenuButton asChild onClick={() => navigate('/admin/default-settings')}>
+                      <div>
+                        <Icon name="account" size={12} className="text-icons-4 transition-colors hover:text-primary" />
+                        <span className="text-foreground-3 font-medium transition-colors hover:text-primary">
+                          User Management
```
shaurya-harness, line 155: non blocking - but lets add translation store here

PR #1171, opened 2025-02-27
chore: Remove dependency on 'react-router-dom' from 'packages/ui'

vardanbansal-harness:
**Migration - Part 1**

Removes references of `react-router-dom` from `packages/ui/src/components`.

Usage with `react-router-dom` `v6`

```
 <RouterProvider>
    <Routes>
      <Route path="/" element={<Home />} />
      <Route path="/about" element={<About />} />
    </Routes>
</RouterProvider>
```

Usage with `react-router-dom` `v5` : Uses `Switch`

```
<RouterProvider>
  <Navigation />
  <Switch>
    <Route path="/" exact component={Home} />
    <Route path="/about" exact component={About} />
  </Switch>
</RouterProvider>
```

### DIFF
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -1,10 +1,11 @@
 import { I18nextProvider } from 'react-i18next'
-import { createBrowserRouter, RouterProvider } from 'react-router-dom'
+import { createBrowserRouter, Link, Outlet, RouterProvider } from 'react-router-dom'
 
 import { QueryClientProvider } from '@tanstack/react-query'
 
 import { CodeServiceAPIClient } from '@harnessio/code-service-client'
 import { ToastProvider, TooltipProvider } from '@harnessio/ui/components'
+import { RouterProvider as RouterProviderV1 } from '@harnessio/ui/context'
 
 import { ExitConfirmProvider } from './framework/context/ExitConfirmContext'
 import { NavigationProvider } from './framework/context/NavigationContext'
@@ -39,7 +40,9 @@ export default function AppV1() {
             <TooltipProvider>
               <ExitConfirmProvider>
                 <NavigationProvider routes={routes}>
-                  <RouterProvider router={router} />
+                  <RouterProviderV1 OutletComponent={Outlet} LinkComponent={Link}>
+                    <RouterProvider router={router} />
+                  </RouterProviderV1>
                 </NavigationProvider>
               </ExitConfirmProvider>
             </TooltipProvider>
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -1,5 +1,5 @@
 import { I18nextProvider } from 'react-i18next'
-import { createBrowserRouter, Link, Outlet, RouterProvider } from 'react-router-dom'
+import { createBrowserRouter, Link, NavLink, Outlet, RouterProvider } from 'react-router-dom'
 
 import { QueryClientProvider } from '@tanstack/react-query'
 
@@ -40,7 +40,7 @@ export default function AppV1() {
             <TooltipProvider>
               <ExitConfirmProvider>
                 <NavigationProvider routes={routes}>
-                  <RouterProviderV1 OutletComponent={Outlet} LinkComponent={Link}>
+                  <RouterProviderV1 Link={Link} NavLink={NavLink} Outlet={Outlet} navigate={router.navigate}>
                     <RouterProvider router={router} />
                   </RouterProviderV1>
                 </NavigationProvider>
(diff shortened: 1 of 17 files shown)


apps/gitness/src/App.tsx
```diff
@@ -1,10 +1,11 @@
 import { I18nextProvider } from 'react-i18next'
-import { createBrowserRouter, RouterProvider } from 'react-router-dom'
+import { createBrowserRouter, Link, NavLink, Outlet, RouterProvider } from 'react-router-dom'
 
 import { QueryClientProvider } from '@tanstack/react-query'
 
 import { CodeServiceAPIClient } from '@harnessio/code-service-client'
 import { ToastProvider, TooltipProvider } from '@harnessio/ui/components'
+import { RouterProvider as RouterProviderV1 } from '@harnessio/ui/context'
```
abhinavrastogi-harness, line 8: why is named with a `V1`?

packages/ui/src/context/RouterContext.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const NavLinkDefault = ({ to, children, className, style, ...props }: NavLinkProps) => {
+  const href = resolveTo(to)
+  const isActive = new URL(href, window.location.origin).pathname === window.location.pathname
+
+  const finalClassName =
+    typeof className === 'function'
+      ? className({ isActive, isPending: false, isTransitioning: false })
+      : cn(className, isActive ? 'font-bold text-blue-600' : 'text-blue-500')
```
abhinavrastogi-harness, line 31: do we really need to include custom styles in the fallback default impl?

abhinavrastogi-harness (approved):

apps/gitness/src/App.tsx
vardanbansal-harness: This is just temporary, eventually this will replace `RouterProvider` from `react-router-dom`, will rename it then.

packages/ui/src/context/RouterContext.tsx
vardanbansal-harness: Not needed, will clean it up.

packages/ui/src/context/RouterContext.tsx
```diff
@@ -0,0 +1,65 @@
+import { ComponentType, createContext, ReactNode, useContext } from 'react'
+import type { LinkProps, NavLinkProps, OutletProps } from 'react-router-dom'
```
knagurski, line 2: I assume the next PR will kill all this off?

packages/ui/src/context/RouterContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+  children,
+  Link = LinkDefault,
+  NavLink = NavLinkDefault,
+  Outlet = OutletDefault,
+  navigate = to => {
+    window.location.href = to
+  }
+}: {
+  children: ReactNode
+} & Partial<RouterContextType>) => {
```
knagurski, line 63: Usually best to use `PropsWithChildren`
```suggestion
}: PropsWithChildren<Partial<RouterContextType>>) => {

File: packages/ui/src/context/RouterContext.tsx [hunk 5]
```diff
@@ -0,0 +1,65 @@
...
+export const useRouterContext = () => useContext(RouterContext)
+
+export const RouterProvider = ({
+  children,
+  Link = LinkDefault,
+  NavLink = NavLinkDefault,
+  Outlet = OutletDefault,
+  navigate = to => {
+    window.location.href = to
+  }
```
knagurski, line 60: This will make `navigate` not dependency array safe since it'll be a new function on every render. Can we extract it to a function/variable like the other defaults? Not an issue in real-world usage when it's overridden, but worth fixing.

packages/ui/src/components/navbar/navbar.tsx
```diff
@@ -1,7 +1,8 @@
 import { useMemo } from 'react'
-import { Link, useLocation, useNavigate } from 'react-router-dom'
+import { useLocation } from 'react-router-dom'
```
knagurski, line 2: Can we get this in the context too?

packages/ui/src/components/navbar/navbar.tsx
```diff
@@ -44,8 +45,9 @@ export const Navbar = ({
   useThemeStore,
   useTranslationStore
 }: NavbarProps) => {
+  const { Link } = useRouterContext()
   const location = useLocation()
-  const navigate = useNavigate()
+  const { navigate } = useRouterContext()
```
knagurski, line 50: Can we do all 3 lines with one call?

PR #1174, opened 2025-02-28
Fixed infinite loading issue for Edit Label (MFE Only)

vivek-harness:
Edit label page goes into an infinite loop when it encounters a SPACE character and other Special Characters in the label. This PR aims fixing that issue, but this still won't work for the following special Characters: #, %, / and \\

Need to ask the Product team if we need to support those characters in labels.

https://github.com/user-attachments/assets/e709b0ff-ecb8-4a40-92a0-82cd2bbb9dcf

### DIFF
--- a/apps/gitness/src/AppMFE.tsx
+++ b/apps/gitness/src/AppMFE.tsx
@@ -20,6 +20,7 @@ import { extractRedirectRouteObjects } from './framework/routing/utils'
 import { useLoadMFEStyles } from './hooks/useLoadMFEStyles'
 import i18n from './i18n/i18n'
 import { mfeRoutes, repoRoutes } from './routes'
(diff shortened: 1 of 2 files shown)


apps/gitness/src/AppMFE.tsx
```diff
@@ -20,6 +29,7 @@ import { extractRedirectRouteObjects } from './framework/routing/utils'
 import { useLoadMFEStyles } from './hooks/useLoadMFEStyles'
 import i18n from './i18n/i18n'
 import { mfeRoutes, repoRoutes } from './routes'
+import { decodeURIComponentIfValid } from './utils/path-utils'
```
vardanbansal-harness, line 32: can we add a `vitest` for this util?

vardanbansal-harness (approved):

PR #1175, opened 2025-03-01
chore: Replace 'react-router-dom' imports with RouterContextProvider in 'packages/ui/views'

vardanbansal-harness:
Follow up to PR https://github.com/harness/canary/pull/1171.

- Removes all imports from the package `react-router-dom` (except types) and replaces it with corresponding imports from [RouterContextProvider](https://github.com/harness/canary/blob/main/packages/ui/src/context/RouterContext.tsx).


packages/ui/src/views/landing-page.tsx
```diff
@@ -1,7 +1,8 @@
 import { FC } from 'react'
-import { Link, LinkProps } from 'react-router-dom'
+import type { LinkProps } from 'react-router-dom'
```
sans-harness, line 2: should we be importing from `react-router-dom` here?

vardanbansal-harness: This package will be removed in a followup PR including the types used from it.

abhinavrastogi-harness (approved):

PR #1178, opened 2025-03-03
Existing secret page

sans-harness:
Uploading Screen Recording 2025-03-03 at 2.43.20 PM.mov…


packages/ui/src/views/secrets/create-secret/create-secret-page.tsx
```diff
@@ -0,0 +1,266 @@
...
+      </FormWrapper>
+
+      {/* Hidden file input */}
+      <Input
+        id="secret-file-input"
+        type="file"
+        ref={fileInputRef}
+        onChange={handleFileChange}
+        style={{ display: 'none' }}
+        accept=".txt,.json,.yaml,.yml,.env,.key,.pem,.crt,.cer,.pub"
```
abhinavrastogi-harness, line 262: why these formats? is this documented somewhere?

packages/ui/src/views/secrets/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './new-secret/new-secret-form'
+export * from './create-secret/create-secret-page'
```
abhinavrastogi-harness, line 1: i don't think this is a "page"... let's just call it `create-secret` since it's already scoped under `views` folder.

abhinavrastogi-harness (approved):

PR #1181, opened 2025-03-04
refactor: tabs navigation

3em:
There was a bug in the tabs: when switching between tabs using the Tab key, you could move forward but get stuck when going backward. The issue is likely related to having links inside buttons in the Tabs component.
Moved tabs to the new TabNav component

https://github.com/user-attachments/assets/43319a57-0f83-4862-87f9-3ca61933bdf1


packages/ui/src/components/tabs.tsx
```diff
@@ -9,7 +9,10 @@ const tabsListVariants = cva('inline-flex items-center text-foreground-4', {
     variant: {
       default: 'h-9 justify-center rounded-lg bg-muted p-1',
       underline: 'h-11 justify-center gap-4',
-      navigation: 'h-[44px] w-full justify-start gap-6 border-b border-borders-5 px-5',
+      /**
+       * @deprecated Use TabNav component instead
+       */
+      navigation: '',
```
knagurski, line 15: Is this variant used anywhere? If so, can we transition those instances and remove this? If not, we can go ahead and kill this off anyway.
```suggestion
```

packages/ui/src/views/repo/pull-request/pull-request-list-page.tsx
```diff
@@ -198,7 +198,8 @@ const PullRequestListPage: FC<PullRequestPageProps> = ({
           <>
             <PRListFilterHandler
               ref={filtersRef}
-              onFilterSelectionChange={onFilterSelectionChange}
+              // TODO: fix type issue
+              {...{ onFilterSelectionChange: onFilterSelectionChange }}
```
knagurski, line 202: I'm not seeing any type issues here 🤔 
```suggestion
              onFilterSelectionChange={onFilterSelectionChange}
```

packages/ui/src/views/layouts/PullRequestLayout.tsx
```diff
@@ -21,105 +36,70 @@ enum PullRequestTabsKeys {
...
+  const navigate = useNavigate()
+  const location = useLocation()
+
+  const getIsActiveTab = useCallback(
+    (tab: string) => (location.pathname.endsWith(tab) ? 'active' : 'inactive'),
+    [location.pathname]
+  )
+  const makeHandleTabChange = useCallback((tab: string) => () => navigate(tab), [navigate])
+
+  const getTabProps = useCallback(
```
knagurski, line 58: Using a `useCallback` here won't make any positive difference to performance. `useCallback` is great for stabilising function references when passing a function to another component/hook etc. In this case it's only being used within the component and called on every render anyway. Perhaps a better idea would be to have a `useMemo` for each unique instance instead to make sure they aren't regenerated on each render.

packages/ui/src/views/layouts/PullRequestLayout.tsx
```diff
@@ -21,105 +36,70 @@ enum PullRequestTabsKeys {
...
   const { pullRequest } = usePullRequestStore()
   const { t } = useTranslationStore()
+  const navigate = useNavigate()
+  const location = useLocation()
+
+  const getIsActiveTab = useCallback(
+    (tab: string) => (location.pathname.endsWith(tab) ? 'active' : 'inactive'),
+    [location.pathname]
+  )
+  const makeHandleTabChange = useCallback((tab: string) => () => navigate(tab), [navigate])
```
knagurski, line 56: Pretty sure you can just kill this function entirely and just use `navigate` directly

packages/ui/src/views/layouts/PullRequestLayout.tsx
```diff
@@ -21,105 +36,70 @@ enum PullRequestTabsKeys {
...
+
+  const getTabProps = useCallback(
+    (tab: PullRequestTabsKeys): TabsTriggerProps => ({
+      value: tab,
+      ...{ 'data-state': getIsActiveTab(tab) },
```
knagurski, line 61: Instead of workarounds like this, would it make sense to alter `TabTrigger` to have an `active` prop that adds this?

packages/ui/src/views/layouts/PullRequestLayout.tsx
```diff
@@ -21,105 +36,70 @@ enum PullRequestTabsKeys {
...
   spaceId,
   repoId,
   updateTitle
 }) => {
-  const { NavLink, Outlet } = useRouterContext()
+  const { Outlet } = useRouterContext()
   const { pullRequest } = usePullRequestStore()
   const { t } = useTranslationStore()
+  const navigate = useNavigate()
+  const location = useLocation()
```
knagurski, line 50: You can get `navigate` and `location` from `useRouterContext()`
```suggestion
  const { location, navigate, Outlet } = useRouterContext()
  const { pullRequest } = usePullRequestStore()
  const { t } = useTranslationStore()
```

packages/ui/src/components/tabs.tsx
ankormoreankor: @knagurski the `Tabs` component with the `"navigation"` variant is used in these components from the `views package`: `SandboxPullRequestCompare`, `SandboxExecutions`, `RepoExecutionLayout`, `SettingsAccountPage`, `SettingsProjectNav`. None of these components seem to be imported anywhere in the repo, but I'm unsure if I can remove or even change them (by changing the Tabs variant, for instance).
The goal of this PR is to make particular and obvious changes.

packages/ui/src/views/layouts/PullRequestLayout.tsx
ankormoreankor: FIxed

packages/ui/src/views/layouts/PullRequestLayout.tsx
ankormoreankor: There is actually a better way to do it 👍

packages/ui/src/views/layouts/PullRequestLayout.tsx
ankormoreankor: Fixed

packages/ui/src/views/layouts/PullRequestLayout.tsx
ankormoreankor: Got rid of useCallback

packages/ui/src/views/repo/pull-request/pull-request-list-page.tsx
ankormoreankor: The issue has gone after rebase👌

praneshg239 (approved):

packages/ui/src/components/tabs.tsx
andrewgolovanov: I've reverted the styles, and added a `TODO` that we revisit this issue in a separate PR for this, that way we'll be sure these changes don't break anything where it's used

andrewgolovanov (approved):

PR #1184, opened 2025-03-04
chore: Add view-preview, portal and gitness build and dev scripts

knagurski:
This PR adds the following root-level npm scripts to help build the 3 apps with their dependencies.

- `gitness:build` - build for production with all dependency packages
- `gitness:dev` - build all dependency packages and start gitness dev mode
- `portal:build` - build for production with all dependency packages
- `portal:dev` - build all dependency packages and start portal dev mode
- `view-preview:build` - build for production with all dependency packages
- `view-preview:dev` - build all dependency packages and start design-system dev mode

### DIFF
--- a/packages/filters/package.json
+++ b/packages/filters/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "run-p build:watch",
     "build": "vite build",
+    "build:ci": "vite build",
     "build:watch": "vite build --watch",
     "prepublishOnly": "pnpm build",
     "pretty": "prettier --check ./src",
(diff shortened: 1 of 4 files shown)


packages/filters/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "run-p build:watch",
     "build": "vite build",
+    "build:ci": "vite build",
```
abhinavrastogi-harness, line 9: what's the purpose of a separate `build:ci` command that runs the same command as `build`?

knagurski: It's really just a standardised placeholder for _"what's the minimum to build this for production"_. There are things we do in build that aren't needed for a production build. For example, we generate types for packages as part of the usual build, but that's just wasted effort for a production build. Currently only the `ui` lib does anything different, but I think it's a good pattern.

PR #1192, opened 2025-03-04
Remove react-router-dom dependencies from sidebar

vivek-harness:
Used the new `useRouterContext()` utility for importing React Router Dom elements instead of using the package itself.

### DIFF
--- a/packages/ui/src/components/navbar-new/app-sidebar.tsx
+++ b/packages/ui/src/components/navbar-new/app-sidebar.tsx
@@ -1,7 +1,6 @@
 import { useState } from 'react'
-import { Link, useNavigate } from 'react-router-dom'
 
-import { IThemeStore } from '@/context'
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/navbar-new/app-sidebar.tsx
abhinavrastogi-harness, line 1: `navbar-new` is a pretty bad name for a component/file. Let's rename this to just `sidebar` after we delete the old `navbar` components

abhinavrastogi-harness (approved):

packages/ui/src/components/navbar-new/app-sidebar.tsx
vivek-harness: I was planning to name it "app-sidebar" since we already have a sidebar component that we are using for rendering sidebar elements like MenuItem, MenuAction, etc.

PR #1194, opened 2025-03-05
Entity Renderer

sans-harness:
https://github.com/user-attachments/assets/1eb03713-fb17-4c0f-884d-d646b7eac19a

- Introduces a generic entity renderer component that can take in a list of items, renderers and can render them together. 
- The current component is tied to the `ExistingSecrets` page and introduces mock data for that page. The example is a POC of how the component works


packages/ui/src/views/secrets/existing-secret/existing-secrets.tsx
```diff
@@ -0,0 +1,154 @@
...
+  onScopeChange: (scope: SecretScope) => void
+  onCancel?: () => void
+}
+
+// Component for selecting existing secrets
+export const ExistingSecrets: React.FC<ExistingSecretsProps> = ({
+  // Data
+  accountSecrets,
+  organizationSecrets,
+
```
sans-harness, line 56: shouldn't be getting them seperately

packages/ui/src/views/secrets/existing-secret/existing-secrets.tsx
```diff
@@ -0,0 +1,154 @@
...
+  selectedEntity,
+  activeScope,
+
+  // Callbacks
+  onSelectEntity,
+  onScopeChange,
+  onCancel
+}) => {
+  // Define available scopes
+  const availableScopes: SecretScope[] = ['account', 'organization']
```
sans-harness, line 67: should include project

packages/ui/src/views/secrets/existing-secret/existing-secrets.tsx
```diff
@@ -0,0 +1,154 @@
...
+      case 'organization':
+        return organizationSecrets
+      default:
+        return []
+    }
+  }
+
+  // Custom entity renderer for secrets
+  const renderEntity = (props: EntityRendererProps<SecretItem>) => {
+    const { entity } = props
```
sans-harness, line 83: take all props, encapsulate most logic inside entity renderer

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+  entities,
+  selectedEntity,
+  parentFolder,
+  childFolder,
+
+  // Callbacks
+  onSelectEntity,
+  onScopeChange,
+
+  showFilter = true,
```
abhinavrastogi-harness, line 64:
```suggestion
  // configs
  showFilters = true,
```

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+      onScopeChange?.(direction)
+    },
+    [onScopeChange]
+  )
+
+  const defaultEntityRenderer = ({ entity, isSelected, onSelect }: EntityRendererProps<T>) => {
+    return (
+      <StackedListItem
+        onClick={() => onSelect?.(entity)}
+        className={cn(isSelected && 'bg-background-4')}
```
abhinavrastogi-harness, line 90: generally better to avoid the `&&` syntax.
```suggestion
        className={cn({ 'bg-background-4': isSelected })}
```

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+        thumbnail={<Icon name="file" size={16} className="text-foreground-5" />}
+        actions={
+          <Button
+            variant="default"
+            size="sm"
+            onClick={e => {
+              e.stopPropagation()
```
abhinavrastogi-harness, line 97: why do we need this?

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+          </Button>
+        }
+      >
+        <StackedListField title={entity.name} />
+      </StackedListItem>
+    )
+  }
+
+  // Default scope selector renderer
+  const defaultScopeSelectorRenderer = ({ parentFolder, onSelect }: ScopeSelectorProps<S>) => {
```
abhinavrastogi-harness, line 111: instead of "scope", we could just call these `parentFolderRenderer` and `childFolderRenderer` to make it independent of Harness scopes concept

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+        <StackedListField title={<span className="capitalize">{String(folder)}</span>} />
+      </StackedListItem>
+    )
+  }
+
+  const renderCombinedList = () => {
+    return (
+      <StackedList>
+        {/* scopes */}
+        {parentFolder && (
```
abhinavrastogi-harness, line 138: generally better to avoid `&&` syntax
```suggestion
        { parentFolder ? (...) : null }
```

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  return (
+    <>
+      {isLoading ? (
+        <SkeletonList />
+      ) : (
+        <>
+          {showFilter && <Input type="text" placeholder="Search" className="mb-4" />}
+          <ScrollArea className="h-[62%]">
```
abhinavrastogi-harness, line 208: might be better to avoid a fixed relative height. and just let it expand to fill the parent

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+      <StackedListItem
+        onClick={() => onSelect?.(folder)}
+        thumbnail={<Icon name="folder" size={16} className="text-foreground-5" />}
+      >
+        <StackedListField title={<span className="capitalize">{String(folder)}</span>} />
+      </StackedListItem>
+    )
+  }
+
+  const renderCombinedList = () => {
```
abhinavrastogi-harness, line 134: could we move this out to a separate pure component?
otherwise it might be better to wrap it in a relevant useMemo

packages/ui/src/views/secrets/secret-reference/secret-reference.tsx
```diff
@@ -0,0 +1,109 @@
...
+          </Button>
+        }
+      >
+        <StackedList.Field title={entity.secret.name} description={entity.secret.description || undefined} />
+      </StackedList.Item>
+    )
+  }
+
+  // Custom scope selector renderer
+  const renderScopeSelector = (props: ScopeSelectorProps<string>) => {
```
abhinavrastogi-harness, line 70: i'd like to avoid overriding the scope selector. the default renderer should cover the standard usecase, since it'll largely be the same for all entities

packages/ui/src/views/secrets/secret-reference/secret-reference.tsx
```diff
@@ -0,0 +1,109 @@
...
+            onClick={e => {
+              e.stopPropagation()
+              onSelect(entity)
+            }}
+          >
+            Select
+          </Button>
+        }
+      >
+        <StackedList.Field title={entity.secret.name} description={entity.secret.description || undefined} />
```
abhinavrastogi-harness, line 64: why do we need `|| undefined`?

packages/ui/src/views/secrets/secret-reference/secret-reference.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <StackedList.Field title={<span className="capitalize">{parentFolder}</span>} />
+      </StackedList.Item>
+    )
+  }
+
+  return (
+    <div className="flex flex-col h-full">
+      <span className="font-medium mb-4">Select an existing Secret:</span>
+      <div className="flex-1">
+        <EntityReference<SecretItem, string, string>
```
abhinavrastogi-harness, line 87: `string, string` is probably not needed as it's the default types

packages/ui/src/views/secrets/types.ts
```diff
@@ -12,6 +14,36 @@ export enum SecretCreationType {
...
+
+export enum SecretScopeEnum {
+  ACCOUNT = 'account',
+  ORGANIZATION = 'organization',
+  PROJECT = 'project'
+}
+
+export enum DirectoryType {
+  UP = 'up',
+  DOWN = 'down'
```
abhinavrastogi-harness, line 29: "parent" and "child" might be better?

abhinavrastogi-harness: also, this should be part of EntityReference types.. not specific to Secrets

packages/ui/src/views/secrets/types.ts
```diff
@@ -12,6 +14,36 @@ export enum SecretCreationType {
...
-  Existing = 'existing'
+  NEW = 'new',
+  EXISTING = 'existing'
+}
+
+export enum SecretScopeEnum {
+  ACCOUNT = 'account',
+  ORGANIZATION = 'organization',
+  PROJECT = 'project'
+}
```
abhinavrastogi-harness, line 25: this is also not specific to secrets.. let's generalise this

packages/ui/src/views/entity-reference/entity-reference-component.tsx
```diff
@@ -0,0 +1,217 @@
...
+// Props for the scope selector item
+export interface ScopeSelectorProps<S = string> {
+  parentFolder: S
+  onSelect: (scope: S) => void
+}
+
+export interface FolderRendererProps<F = string> {
+  folder: F
+  onSelect: (folder: F) => void
+}
```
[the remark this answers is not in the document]
sans-harness: Isn't there a use case where the `parent-folder` and `child-folder` have different fields that are required where this becomes useful ?

packages/ui/src/views/secrets/types.ts
```diff
@@ -12,6 +14,26 @@ export enum SecretCreationType {
...
+  name: string
+  identifier: string
+  orgIdentifier?: string
+  projectIdentifier?: string
+  tags: Record<string, string | undefined>
+  description?: string
+}
+
+// Define our custom scope type
+export type SecretScope = 'account' | 'organization' | 'project'
```
abhinavrastogi-harness, line 32: secrets should not have their own scope

packages/ui/src/views/platform/types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface ChildFolderRendererProps<F = string> {
+  folder: F
+  onSelect: (folder: F) => void
+}
+
+export enum ScopeEnum {
+  ACCOUNT = 'account',
+  ORGANIZATION = 'organization',
+  PROJECT = 'project'
```
abhinavrastogi-harness, line 32: this is specific to harness platform.. and it'll not be applicable to opensource. also not applicable for flex hierarchy. let's re-evaluate this if we can avoid adding scopes in the view itself.

abhinavrastogi-harness (approved):

PR #1201, opened 2025-03-06
fix: Fix type for Router provider context property "location"

vardanbansal-harness:


apps/gitness/src/AppMFE.tsx
```diff
@@ -191,7 +191,7 @@ export default function AppMFE({
                                 Link={Link}
                                 NavLink={NavLink}
                                 Outlet={Outlet}
-                                location={window.location}
+                                location={{ ...window.location, state: {}, key: '' }}
```
cjlee01, line 194: whats the key being used for? @vardanbansal-harness

cjlee01 (approved):

apps/gitness/src/AppMFE.tsx
vardanbansal-harness: It's not used anywhere right now, had to add it to fix the type issue. In React router v6, `Location` interface now has a required `key` and `state` properties as well.

PR #1210, opened 2025-03-07
Forms library - multiple bug fixes, playground

srdjan-harness:
- Added util function that removes hidden inputs values:  unsetHiddenInputsValues
- We can access all values in the input validation schema.
- Prevent reusing default values on recreation (delete and created arr/list item )
- Playground added (inputs and examples)

### DIFF
--- a/packages/forms/src/core/components/RootForm.tsx
+++ b/packages/forms/src/core/components/RootForm.tsx
@@ -76,8 +76,25 @@ export function RootForm<TFieldValues extends FieldValues = FieldValues, TContex
   const submittedRef = useRef(false)
 
   useEffect(() => {
-    methods.reset(defaultValues)
+    methods.reset(defaultValues, {})
   }, [methods.reset, defaultValues])
+
+  // reset defaultValues to prevent default on recreated (deleted then created) array/list items
+  useEffect(() => {
+    setTimeout(() => {
+      methods.reset({} as TFieldValues, {
+        keepErrors: true,
+        keepDirty: true,
+        keepDirtyValues: true,
+        keepValues: true,
+        keepDefaultValues: false, // RESET default values
+        keepIsSubmitted: true,
+        keepTouched: true,
+        keepIsValid: true,
+        keepSubmitCount: true
+      })
+    }, 1)
+  }, [])
 
   const { getValues, handleSubmit } = methods
   const values = getValues()
--- a/packages/forms/src/core/utils/transform-utils.tsx
+++ b/packages/forms/src/core/utils/transform-utils.tsx
@@ -1,4 +1,4 @@
-import { cloneDeep, get, pick, set } from 'lodash-es'
+import { cloneDeep, get, pick, set, unset } from 'lodash-es'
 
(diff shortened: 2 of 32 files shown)


packages/forms/src/core/components/RootForm.tsx
```diff
@@ -76,9 +76,26 @@ export function RootForm<TFieldValues extends FieldValues = FieldValues, TContex
...
 
   useEffect(() => {
-    methods.reset(defaultValues)
+    methods.reset(defaultValues, {})
   }, [methods.reset, defaultValues])
 
+  // reset defaultValues to prevent default on recreated (deleted then created) array/list items
+  useEffect(() => {
+    setTimeout(() => {
+      methods.reset({} as TFieldValues, {
```
abhinavrastogi-harness, line 85: why do we need to cast here? any way to avoid it?

packages/forms/src/core/components/RootForm.tsx
```diff
@@ -76,9 +76,26 @@ export function RootForm<TFieldValues extends FieldValues = FieldValues, TContex
...
+        keepDirty: true,
+        keepDirtyValues: true,
+        keepValues: true,
+        keepDefaultValues: false, // RESET default values
+        keepIsSubmitted: true,
+        keepTouched: true,
+        keepIsValid: true,
+        keepSubmitCount: true
+      })
+    }, 1)
```
abhinavrastogi-harness, line 96: i believe setTimeout(0) would achieve the same thing (basically to push it into the event queue).. 
we could also use `requestIdleCallback` for better scheduling since it's already async.

packages/forms/src/core/utils/transform-utils.tsx
```diff
@@ -84,3 +84,22 @@ export function getTransformers(formDefinition: IFormDefinition): TransformItem[
...
+
+/** Remove values for inputs that are hidden  */
+export function unsetHiddenInputsValues(
+  formDefinition: IFormDefinition,
+  values: Record<string, any>,
+  metadata?: any
+): Record<string, any> {
+  const flattenInputs = flattenInputsRec(formDefinition.inputs)
+
+  const retValues = cloneDeep(values)
```
abhinavrastogi-harness, line 96: this could get expensive for large forms. how frequently do we trigger this?
could we update fields in place?

packages/forms/src/types/types.ts
```diff
@@ -73,7 +73,7 @@ export interface IInputDefinition<T = unknown> {
    * Note: Validation of complex inputs has to respect its formik structure
    */
   validation?: {
-    schema?: Schema<unknown>
+    schema?: (values: any) => Schema<unknown>
```
abhinavrastogi-harness, line 76: looks like a breaking change for any view using the validation schema today.
1. either we should make the corresponding changes in the same PR
2. OR support both static and callback schemas `Schema<unknown> | (values: any) => Schema<unknown>`

packages/forms/src/core/components/RootForm.tsx
srdjan-harness: Updated

PR #1236, opened 2025-03-12
Improve form validation

ankormoreankor:
#### Forms validation improvement

- fixed UI when the error message takes more than 1 line (the error message is no longer positioned absolutely)

image

- Profile settings form: more validation happens on the client side

image

- Profile settings form: validation messages are translated to French, missed translation added

image


packages/ui/src/views/repo/repo-branch-rules/components/repo-branch-rules-fields.tsx
```diff
@@ -352,23 +352,14 @@ export const BranchSettingsRuleListField: FC<{
...
-                    options={
-                      recentStatusChecks?.map(check => ({
-                        id: check,
-                        label: check
-                      })) ?? []
-                    }
-                  />
-                </div>
+                <MultiSelect
+                  className="pl-[26px]"
```
praneshg239, line 356: can we avoid using arbitrary values? like pl-[STATIC_VALUE]

praneshg239 (changes requested):

packages/ui/src/views/profile-settings/profile-settings-general-page.tsx
```diff
@@ -72,22 +104,13 @@ const SettingsAccountGeneralPage: FC<SettingsAccountGeneralPageProps> = ({
...
-    }),
-    [userData]
-  )
-
   const {
     register: registerProfile,
     handleSubmit: handleProfileSubmit,
     reset: resetProfileForm,
     formState: { errors: profileErrors, isValid: isProfileValid, dirtyFields: profileDirtyFields }
   } = useForm<ProfileFields>({
```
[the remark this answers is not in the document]
3em: @praneshg239 Thank you for pointing me towards useZodForm. However, it has not been migrated from Canary to UI yet.

Therefore, we can create a separate task to migrate all forms to useZodForm, but not in this task, as this one is simply for fixing the display of errors.

packages/ui/src/views/repo/repo-branch-rules/components/repo-branch-rules-fields.tsx
3em: @praneshg239 Unfortunately, this is the design with that specific margin, which is why such sizes sometimes appear in the code.

praneshg239: @3em Got it.

We need to update our design to leverage values from the design system. If we continue this practice, then we will add more arbitrary values in the code. It is like overruling the design system. I hope the new design system will fix it 👍 

cc: @cxramina

PR #1244, opened 2025-03-13
Inset theme

ankormoreankor:


apps/design-system/src/pages/view-preview/app-view-wrapper.tsx
```diff
@@ -82,6 +84,18 @@ export const AppViewWrapper: FC<PropsWithChildren<AppViewWrapperProps>> = ({
     setShowSettingsMenu(current => !current)
   }, [])
 
+  const setInset = useCallback(() => {
+    const inset = sessionStorage.getItem('view-preview-is-inset')
+    setIsInset(inset === 'true' ? true : false)
```
3em, line 89: setIsInset(inset === 'true')

apps/design-system/src/pages/view-preview/app-view-wrapper.tsx
```diff
@@ -82,6 +84,18 @@ export const AppViewWrapper: FC<PropsWithChildren<AppViewWrapperProps>> = ({
...
+  }, [])
+
+  useEffect(setInset, [setInset])
+
+  useEffect(() => {
+    window.addEventListener('storageChange', setInset)
+    return () => window.removeEventListener('storageChange', setInset)
+  }, [setInset])
```
3em, line 97: All of this (code with `setInset`) can be written in one `useEffect`.

packages/ui/src/components/app-sidebar/app-sidebar.tsx
```diff
@@ -72,6 +72,8 @@ export const AppSidebar = ({
     setOpenLanguageDialog(false)
   }
 
+  const onInsetChange = (style: ContentStyleType) => setInset(style === ContentStyleType.Inset ? true : false)
```
3em, line 75: `style === ContentStyleType.Inset`

apps/design-system/src/pages/view-preview/app-view-wrapper.tsx
ankormoreankor: FIxed

apps/design-system/src/pages/view-preview/app-view-wrapper.tsx
ankormoreankor: Fixed

packages/ui/src/components/app-sidebar/app-sidebar.tsx
ankormoreankor: FIxed

apps/gitness/src/components-v2/app-shell.tsx
```diff
@@ -226,19 +226,23 @@ export const AppShellMFE = () => {
...
 
 function BreadcrumbsAndOutlet({ className }: { className?: string }) {
+  const { isInset } = useThemeStore()
+
   return (
-    <div className={cn('h-full flex flex-col', className)}>
-      <div className="layer-high sticky top-0 bg-background-1">
-        <Breadcrumbs />
+    <div className={cn('h-full', { 'overflow-hidden h-screen p-2 bg-sidebar-background-1': isInset })}>
+      <div className={cn('h-full flex flex-col', { 'rounded-md overflow-auto bg-background-1': isInset }, className)}>
```
abhinavrastogi-harness, line 240: should the className from props be applied to the outermost div?
with this change, we will not be able to override the inset styles from outside (though is that even a requirement?)

apps/gitness/src/framework/context/ThemeContext.tsx
```diff
@@ -59,6 +61,8 @@ export function ThemeProvider({ children, defaultTheme }: ThemeProviderProps) {
 
     if (!isMFE) {
       root.classList.add(effectiveTheme) // Apply the computed theme class
+
+      if (isInset) root.classList.add('overflow-hidden')
```
abhinavrastogi-harness, line 65: where do we remove this newly added class?
also, preventing overflow on body can cause wierd when the content is genuinely larger than the viewport. does this mean we'll need to rely on fixed-height content on all pages?

apps/gitness/src/pages-v2/repo/repo-commit-details-diff.tsx
```diff
@@ -100,7 +102,12 @@ export const CommitDiffContainer = ({ showSidebar = true }: { showSidebar?: bool
   return (
     <>
       {showSidebar && (
-        <CommitSidebar useTranslationStore={useTranslationStore} navigateToFile={() => {}} filesList={filesList}>
+        <CommitSidebar
+          useTranslationStore={useTranslationStore}
+          navigateToFile={() => {}}
+          filesList={filesList}
+          isInsetTheme={isInset}
```
abhinavrastogi-harness, line 109: instead of passing `isInset` via props, could we utilise `:where` css pseudoclass within the view to selectively apply styles where needed?
requiring a prop to make this work makes it less scalable & maintainable.

apps/gitness/src/pages-v2/repo/repo-commit-details-diff.tsx
```diff
@@ -100,7 +102,12 @@ export const CommitDiffContainer = ({ showSidebar = true }: { showSidebar?: bool
   return (
     <>
       {showSidebar && (
-        <CommitSidebar useTranslationStore={useTranslationStore} navigateToFile={() => {}} filesList={filesList}>
+        <CommitSidebar
+          useTranslationStore={useTranslationStore}
+          navigateToFile={() => {}}
```
abhinavrastogi-harness, line 107: how come this function is a noop?

packages/ui/src/components/theme-selector-v2/theme-dialog.tsx
```diff
@@ -156,14 +151,38 @@ const ThemeDialog: FC<ThemeDialogProps> = ({
             </Select.Root>
           </div>
 
+          <Separator className="bg-borders-4 h-px" />
+
+          {/* Inset Adjustment */}
+          <div className="grid grid-cols-[246px_1fr] gap-x-8">
```
abhinavrastogi-harness, line 157: any way to avoid magic numbers?

packages/ui/src/context/theme/types.ts
```diff
@@ -10,6 +10,11 @@ export enum ColorType {
   ProtanopiaAndDeuteranopia = 'pnd'
 }
 
+export enum ContentStyleType {
+  Inset = 'inset',
+  NoInset = 'noInset'
```
abhinavrastogi-harness, line 15: if these are the only two options, a better convention would be store a boolean `inset: boolean`.
but if we think might have more styles in the future, let's not call it `NoInset`, let's give it a name like `Default` or `Full`.

packages/ui/src/shared-style-variables.css
```diff
@@ -1197,30 +1197,20 @@
...
-    --canary-sidebar-foreground-02: var(--canary-base-chrome-700);
-    --canary-sidebar-foreground-03: var(--canary-base-chrome-800);
-    --canary-sidebar-foreground-04: var(--canary-base-chrome-600);
-    --canary-sidebar-foreground-05: var(--canary-base-chrome-600);
-    --canary-sidebar-foreground-06: var(--canary-base-chrome-800);
-    --canary-sidebar-foreground-accent: var(--canary-base-chrome-1000);
-    --canary-sidebar-icon-01: var(--canary-base-chrome-1000);
-    --canary-sidebar-icon-02: var(--canary-base-chrome-700);
-    --canary-sidebar-icon-03: var(--canary-base-chrome-500);
+    --canary-sidebar-background-01: 240 2% 8%;
```
abhinavrastogi-harness, line 1200: this looks like a regression.. why are we going from using color tokens to hardcoded lch values?
cc @cxramina

packages/ui/src/shared-style-variables.css
```diff
@@ -1288,6 +1278,308 @@
...
+    --diff-add-content-highlight--: hsl(var(--canary-green-100));
+    --diff-del-content-highlight--: hsl(var(--canary-red-100));
+    --diff-add-widget--: hsl(var(--canary-base-chrome-100));
+    --diff-add-widget-color--: hsl(var(--canary-base-chrome-850));
+    --diff-empty-content--: hsl(var(--canary-background-01));
+    --diff-hunk-content-color--: hsl(var(--canary-base-chrome-850));
+
+    /* New colors design variables */
+    /* Text */
+    --canary-foreground-01: 240 2% 8%;
```
abhinavrastogi-harness, line 1406: ideally i'd expect color tokens to be defined at the top of the class, and then use those variables below that.

packages/ui/src/shared-style-variables.css
```diff
@@ -1288,6 +1278,308 @@
...
+    --canary-tag-background-peach-02: 30 100% 93%;
+    /* --red */
+    --canary-tag-foreground-red-01: 3 60% 45%;
+    --canary-tag-border-red-01: 3 93% 88%;
+    --canary-tag-background-red-01: 4 81% 96%;
+    --canary-tag-background-red-02: 2 83% 93%;
+
+    /* Labels */
+    --canary-label-background-black-01: hsla(var(--canary-pure-white) / 0.7);
+    --canary-label-foreground-orange-01: hsla(17, 88%, 40%, 1);
```
abhinavrastogi-harness, line 1556: why are we not using the base tokens for labels?

packages/ui/src/views/repo/repo-sidebar/index.tsx
```diff
@@ -25,14 +27,16 @@ export const RepoSidebar = ({
...
+  isInsetTheme
 }: RepoSidebarProps) => {
   const { branchList } = useRepoBranchesStore()
 
   return (
     // 100vh = screen height - (55px Breadcrumbs Height + 45px SubHeader Height = 100px)
+    // Inset theme compensation - 8px*2
     // Total height of topbar and outlet should be 100vh
-    <div className="sticky top-[100px] h-[calc(100vh-100px)]">
+    <div className={cn('sticky top-[100px] h-[calc(100vh-100px)]', { 'h-[calc(100vh-100px-16px)]': isInsetTheme })}>
```
abhinavrastogi-harness, line 39: we are applying one height value unconditonally, and one based on isInset. This means when `isInset` is true, two height classes will be applied. this should not be the case.
both height classes should be applied conditionally.

packages/ui/src/shared-style-variables.css
cxramina: This is the current implementation of PP, this logic is not related to the design system we are currently working on in Harness. Pranesh and I will delete this file once we have applied all the new tokens to the components.

apps/gitness/src/framework/context/ThemeContext.tsx
3em: > where do we remove this newly added class? 
Line 60

> also, preventing overflow on body can cause wierd when the content is genuinely larger than the viewport. does this mean we'll need to rely on fixed-height content on all pages?
The thing is, for this particular theme, all the content is displayed in a block with a scroll inside. So now we don't have a global scroll for the entire body. That's why the whole layout currently has a wrapper to ensure this specific scroll logic works inside the container. We don't need to do anything with the content height; it works with any height.

apps/gitness/src/pages-v2/repo/repo-commit-details-diff.tsx
3em: We don't need to pass the isInsetTheme prop here, as we have a ThemeProvider in UI that contains all the data. We'll fix it.

apps/gitness/src/pages-v2/repo/repo-commit-details-diff.tsx
3em: Good question, it was like that earlier. We’ll add the missing props.

packages/ui/src/shared-style-variables.css
3em: These colors were introduced as a temporary solution for the light theme, so that Brad had a light theme for the showcase. Now, Nastya is working on a completely different color system, so you can ignore the variables.

packages/ui/src/shared-style-variables.css
3em: These colors were introduced as a temporary solution for the light theme, so that Brad had a light theme for the showcase. Now, Nastya is working on a completely different color system, so you can ignore the variables.

packages/ui/src/views/repo/repo-sidebar/index.tsx
3em: We use the cn function, which works based on tailwind-merge. This means duplicates or similar properties will be removed, and only the last one specified will be applied. This approach is used in all components. Therefore, in this case, there won’t be two height classes; there will always be one, depending on the condition.

apps/gitness/src/pages-v2/repo/repo-commit-details-diff.tsx
ankormoreankor: **File search** in the commit sidebar has not been implemented yet, so it's not even rendered on the page, and the component has a **TODO comment**. 
I think, we better finish this feature in a separate branch.
I can comment above this line, so it won't bother during the review

image
image

packages/ui/src/components/theme-selector-v2/theme-dialog.tsx
ankormoreankor: I think the classname speaks for itself: there are two columns, one is 246px, and another takes the rest of the space. Easy to read, no logic is connected to this magic number. And this approach is approach is used widely in the codebase.
I could move it to a separate classname into `style.css` using `@apply` but it doesn't seem reusable.

apps/gitness/src/components-v2/app-shell.tsx
ankormoreankor: You're right here. Fixed👌

apps/gitness/src/pages-v2/repo/repo-commit-details-diff.tsx
ankormoreankor: Fixed👌

packages/ui/src/context/theme/types.ts
ankormoreankor: Default sounds good👌 FIxed

3em (approved):

PR #1258, opened 2025-03-17
Pipeline studio - multiple updates

srdjan-harness:


apps/design-system/src/subjects/views/unified-pipeline-studio/unified-pipeline-studio.tsx
```diff
@@ -1,17 +1,30 @@
...
 const PipelineStudioViewWrapper = () => {
+  const [yamlRevision, onYamlRevisionChange] = useState<YamlRevision>({ yaml: pipeline1 })
+
   return (
-    <UnifiedPipelineStudio
-      useUnifiedPipelineStudioStore={usePipelineStudioStore}
-      useTranslationStore={useTranslationStore}
-      useTemplateListStore={useTemplateListStore}
-    />
+    <div style={{ width: '80vw' }}>
```
abhinavrastogi-harness, line 16: no, i don't think we can/should hardcode this?

packages/ui/src/views/unified-pipeline-studio/unified-pipeline-studio.tsx
```diff
@@ -44,25 +44,45 @@ export interface UnifiedPipelineStudioProps {
   useTemplateListStore: () => ITemplateListStore
   useTranslationStore: () => TranslationStore
   initialView?: VisualYamlValue
+  // NOTE: new props - wip
+  yamlRevision: YamlRevision
```
abhinavrastogi-harness, line 48: merging this PR in the current state would mean you have two sources of truth: the props and the context. 
should we complete the migration before merging?

srdjan-harness: This will be done in follow up PR.

abhinavrastogi-harness (approved):

PR #1263, opened 2025-03-18
Fix canary components styling inside a Shadow Root

vivek-harness:
This PR aims to fix the styling for Canary Components rendered in NGUI without having to import the Shared Styles separately in NGUI index.scss

Adding :host selector for CSS variables fixed this issue previously, but was breaking the styling for Radix UI elements rendered inside a container. That issue is also taken care of in this PR.

Before Change:
https://github.com/user-attachments/assets/85e82f41-3baf-4559-9b64-e71e62ae924e

After Change:
https://github.com/user-attachments/assets/57f7c580-775a-4300-b588-8d617ab16899

Standalone Test:
https://github.com/user-attachments/assets/b45f8142-72c5-4da9-a78c-0af778748c6c

### DIFF
--- a/apps/gitness/src/AppMFE.tsx
+++ b/apps/gitness/src/AppMFE.tsx
@@ -147,12 +147,12 @@ export default function AppMFE({
   }, [theme])
 
   const shadowRef = useRef<HTMLDivElement>(null)
-  const shadowRootElement = shadowRef.current?.shadowRoot as ShadowRoot | undefined
+  const shadowRoot = shadowRef.current?.shadowRoot as ShadowRoot | undefined
 
   const portalRef = useRef<HTMLDivElement>(null)
   const portalContainer = portalRef.current as Element | undefined
 
-  const isStylesLoaded = useLoadMFEStyles(shadowRootElement)
+  const isStylesLoaded = useLoadMFEStyles(shadowRoot)
 
   // Router Configuration
   const basename = `/ng${renderUrl}`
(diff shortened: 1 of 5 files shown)


apps/gitness/src/AppMFE.tsx
```diff
@@ -148,10 +146,15 @@ export default function AppMFE({
     }
   }, [theme])
 
+  // Styles will be loaded inside shadowRoot
+  const shadowRef = useRef<HTMLDivElement>(null)
+  const shadowRoot = shadowRef.current?.shadowRoot as ShadowRoot | undefined
```
sans-harness, line 151: why do we need to explicitly typecaste this?

apps/gitness/src/AppMFE.tsx
```diff
@@ -148,10 +146,15 @@ export default function AppMFE({
...
+
+  // Radix UI elements will be rendered inside portalContainer
   const portalRef = useRef<HTMLDivElement>(null)
-  const portalContainer = portalRef.current?.shadowRoot as Element | undefined
+  const portalContainer = portalRef.current as Element | undefined
```
sans-harness, line 155: why do we need to explicitly typecaste this?

apps/gitness/src/AppMFE.tsx
vardanbansal-harness: +1
 
Trimming down the actual type `ShadowRoot | null | undefined` to `ShadowRoot | undefined` has no real benefit here.

sans-harness (approved):

vardanbansal-harness (approved):

PR #1265, opened 2025-03-18
feat: add input ref componenet

sans-harness:
https://github.com/user-attachments/assets/9e1271ec-e150-44fb-a23e-7790bb5249b8


packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { cn } from '@utils/cn'
+import { cva, type VariantProps } from 'class-variance-authority'
+
+const inputReferenceVariants = cva(
+  'flex min-h-9 cursor-pointer items-center justify-between rounded px-3 py-2 transition-colors',
+  {
+    variants: {
+      variant: {
+        default: 'bg-input-background text-foreground-1',
+        clean: 'bg-transparent'
```
abhinavrastogi-harness, line 13: do we have a usecase for a "clean" variant?

packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -0,0 +1,167 @@
...
+      },
+      size: {
+        sm: 'h-8',
+        md: 'h-9'
```
abhinavrastogi-harness, line 17: same here.. why two sizes right now?

packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -0,0 +1,167 @@
...
+  variant,
+  size,
+  renderValue,
+  ...props
+}: InputReferenceProps<T>) => {
+  // Determine what to display: rendered value if value exists, otherwise placeholder
+  const displayContent =
+    value !== null && value !== undefined ? (renderValue ? renderValue(value) : value) : placeholder
+
+  const hasValue = value !== null && value !== undefined
```
abhinavrastogi-harness, line 120: this can be reused for `displayContent` above

packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Whether the field is optional
+   */
+  optional?: boolean
+}
+
+/**
+ * InputReference is a component that looks like an input field but acts as a clickable
+ * InputReference element that can trigger actions like opening a drawer, modal, or dropdown.
+ * It supports generic types for values and maintains its own internal state.
```
abhinavrastogi-harness, line 97: what state does it maintain internally?

packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -0,0 +1,167 @@
...
+      onClick()
+    }
+  }
+
+  const state = disabled ? 'disabled' : 'default'
+
+  return (
+    <ControlGroup>
+      {!!label && (
+        // eslint-disable-next-line i18next/no-literal-string
```
abhinavrastogi-harness, line 141: why do we need this eslint ignore comment?

abhinavrastogi-harness (commented):
it seems currently we have haven't handled a11y on this component. we should do the basics like tabindex, aria-role, titles etc at minimum.

packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -0,0 +1,167 @@
...
+    },
+    defaultVariants: {
+      variant: 'default',
+      size: 'md',
+      state: 'default'
+    }
+  }
+)
+
+export interface InputReferenceProps<T> extends VariantProps<typeof inputReferenceVariants> {
```
vardanbansal-harness, line 32: nit: Should `T` be constrained a bit here, to avoid unintended types. This can lead to unexpected behavior if `renderValue` receives a type it doesn't support.

```
export interface InputReferenceProps<T = string> extends ...

Comment (abhinavrastogi-harness) [reply packages/ui/src/views/platform/input-reference-component.tsx:32]: `T = string` only defaults it to a string, and doesn't really constrain what the user can pass.
I believe typically it'd be a DTO like SecretDTO or ConnectorDTO.

File: packages/ui/src/views/platform/input-reference-component.tsx [hunk 7]
```diff
@@ -0,0 +1,172 @@
+import { ReactNode } from 'react'
+
+import { Button, Caption, ControlGroup, Icon, IconProps, Label } from '@/components'
+import { cn } from '@utils/cn'
+import { cva, type VariantProps } from 'class-variance-authority'
+
+const inputReferenceVariants = cva(
+  'flex min-h-9 cursor-pointer items-center justify-between rounded px-3 py-2 transition-colors',
+  {
+    variants: {
```
abhinavrastogi-harness, line 10: We can probably get rid of variant and size variants

abhinavrastogi-harness (approved):

PR #1270, opened 2025-03-19
fix: refactoring of PR conversions Panel and Sidebar components

3em:
- Fixed the issue with updating `prPanelData` in `usePullRequestProviderStore` when calling `dryMerge`.
- Added memoization to functions that were causing unnecessary re-renders of components.
- Fixed dependencies and tested data update in the Panel.
- Updated Panel and Sidebar components to match the design.

Video regarding the issue in the first point:
https://github.com/user-attachments/assets/52b8fcd4-24bd-4b14-936a-33dcd69f1c99

### DIFF
--- a/apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
+++ b/apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
@@ -30,7 +30,11 @@ import {
 } from '@harnessio/code-service-client'
 import { SkeletonList } from '@harnessio/ui/components'
 import { PrincipalType } from '@harnessio/ui/types'
-import { PullRequestConversationPage as PullRequestConversationView } from '@harnessio/ui/views'
+import {
+  LatestCodeOwnerApprovalArrType,
+  PullRequestConversationPage as PullRequestConversationView,
+  TypesPullReq
+} from '@harnessio/ui/views'
 
 import CommitSuggestionsDialog from '../../components-v2/commit-suggestions-dialog'
 import { useAppContext } from '../../framework/context/AppContext'
@@ -56,361 +60,12 @@ import {
 } from './pull-request-utils'
 import { usePullRequestProviderStore } from './stores/pull-request-provider-store'
 
-const onCopyClick = (commentId?: number, isNotCodeComment = false) => {
-  if (!commentId) return
-
-  const url = new URL(window.location.href)
-
-  if (isNotCodeComment) url.pathname = url.pathname.replace('/conversation', '/changes')
-
-  url.searchParams.set('commentId', commentId.toString())
-  copy(url.toString())
-}
-
-export default function PullRequestConversationPage() {
-  const routes = useRoutes()
-  const {
-    pullReqMetadata,
-    refetchPullReq,
-    refetchActivities,
-    setRuleViolationArr,
-    prPanelData,
-    pullReqChecksDecision,
-    updateCommentStatus,
-    dryMerge
-  } = usePullRequestProviderStore(state => ({
-    dryMerge: state.dryMerge,
-    pullReqMetadata: state.pullReqMetadata,
-    refetchPullReq: state.refetchPullReq,
-    refetchActivities: state.refetchActivities,
-    setRuleViolationArr: state.setRuleViolationArr,
-    prPanelData: state.prPanelData,
-    pullReqChecksDecision: state.pullReqChecksDecision,
-    updateCommentStatus: state.updateCommentStatus
-  }))
-
-  const { currentUser: currentUserData } = useAppContext()
-
-  const [checkboxBypass, setCheckboxBypass] = useState(false)
-  const [searchReviewers, setSearchReviewers] = useState('')
-  const [addReviewerError, setAddReviewerError] = useState('')
-  const [removeReviewerError, setRemoveReviewerError] = useState('')
-
-  const [changesLoading, setChangesLoading] = useState(true)
-  const [showDeleteBranchButton, setShowDeleteBranchButton] = useState(false)
-  const [showRestoreBranchButton, setShowRestoreBranchButton] = useState(false)
-  const [errorMsg, setErrorMsg] = useState('')
-  const { spaceId, repoId } = useParams<PathParams>()
-  const {
-    scope: { accountId }
-  } = useMFEContext()
-  const [comment, setComment] = useState<string>('')
-  const [commentId] = useQueryState('commentId')
-  const [isScrolledToComment, setIsScrolledToComment] = useState(false)
-
-  const repoRef = useGetRepoRef()
-  const { pullRequestId } = useParams<PathParams>()
-  const prId = (pullRequestId && Number(pullRequestId)) || -1
-  // const [loadState, setLoadState] = useState('data-loaded')
-
-  const filtersData = usePrFilters()
-
-  const { data: { body: principals } = {} } = useListPrincipalsQuery({
-    // @ts-expect-error : BE issue - not implemnted
-    queryParams: { page: 1, limit: 100, type: 'user', query: searchReviewers, accountIdentifier: accountId }
-  })
-
-  const { data: { body: reviewers } = {}, refetch: refetchReviewers } = useReviewerListPullReqQuery({
-    repo_ref: repoRef,
-    pullreq_number: prId
-  })
-
-  const { data: { body: codeOwners } = {}, refetch: refetchCodeOwners } = useCodeownersPullReqQuery({
-    repo_ref: repoRef,
-    pullreq_number: prId
-  })
-
-  const { data: { body: activityData } = {} } = useListPullReqActivitiesQuery({
-    repo_ref: repoRef,
-    pullreq_number: prId,
-    queryParams: {}
-  })
-
-  /**
-   * get all label-related data
-   */
-  const { searchLabel, changeSearchLabel, labels, labelsValues, handleAddLabel, handleRemoveLabel, appliedLabels } =
-    usePrConversationLabels({
-      repoRef,
-      prId,
-      refetchData: refetchActivities
-    })
-
-  const { mutateAsync: restoreBranch } = useRestorePullReqSourceBranchMutation({})
-  const onRestoreBranch = () => {
-    restoreBranch({
-      repo_ref: repoRef,
-      pullreq_number: prId,
-      body: { bypass_rules: false }
-    })
-      .then(res => {
-        if (res.body.name) {
-          setErrorMsg('')
-          setShowRestoreBranchButton(false)
-          setShowDeleteBranchButton(true)
-          refetchActivities()
-        }
-      })
-      .catch(err => {
-        setErrorMsg(err.message)
-      })
-  }
-  const {
-    data: { body: sourceBranch } = {},
-    error: branchError,
-    refetch: refetchBranch
-  } = useGetBranchQuery({
-    repo_ref: repoRef,
-    branch_name: pullReqMetadata?.source_branch || '',
-    queryParams: { include_checks: true, include_rules: true }
-  })
-  const { mutateAsync: deleteBranch } = useDeletePullReqSourceBranchMutation({
-    repo_ref: repoRef,
-    pullreq_number: prId,
-    queryParams: { dry_run_rules: true }
-  })
-  const { mutateAsync: createBranch } = useCreateBranchMutation({})
-  const { mutateAsync: updateTitle } = useUpdatePullReqMutation({
-    repo_ref: repoRef,
-    pullreq_number: Number(pullRequestId)
-  })
-
-  const handleUpdateDescription = useCallback(
-    (title: string, description: string) => {
-      updateTitle({ body: { title, description } }).catch(err => {
-        setErrorMsg(err.message)
-      })
-      refetchPullReq()
-    },
-    [updateTitle, refetchPullReq]
-  )
-
-  const onDeleteBranch = () => {
-    deleteBranch({
-      repo_ref: repoRef,
-      pullreq_number: prId,
-      queryParams: { bypass_rules: true, dry_run_rules: false }
-    })
-      .then(res => {
-        refetchBranch()
-        if (res?.body?.rule_violations) {
-          const { checkIfBypassAllowed } = extractInfoFromRuleViolationArr(res.body?.rule_violations)
-          if (checkIfBypassAllowed) {
-            setShowDeleteBranchButton(true)
-          } else {
-            setShowDeleteBranchButton(false)
-            setShowRestoreBranchButton(true)
-          }
-        }
-        setShowRestoreBranchButton(true)
-        setErrorMsg('')
-        refetchActivities()
-      })
-      .catch(err => {
-        setErrorMsg(err.message)
-      })
-  }
-
-  useEffect(() => {
-    if (sourceBranch && (pullReqMetadata?.merged || pullReqMetadata?.closed)) {
-      setShowDeleteBranchButton(true)
-    }
-  }, [sourceBranch, pullReqMetadata?.merged, pullReqMetadata?.closed])
-
-  useEffect(() => {
-    if (!branchError) return
-
-    if (pullReqMetadata?.merged || pullReqMetadata?.closed) {
-      setShowRestoreBranchButton(true)
-      return
-    }
-
-    setShowDeleteBranchButton(false)
-    createBranch({
-      repo_ref: repoRef,
-      body: {
-        name: pullReqMetadata?.source_branch || '',
-        target: pullReqMetadata?.source_sha,
-        bypass_rules: true,
-        dry_run_rules: true
-      }
-    }).then(res => {
-      if (res?.body?.rule_violations) {
-        const { checkIfBypassAllowed } = extractInfoFromRuleViolationArr(res.body?.rule_violations)
-
-        setShowRestoreBranchButton(checkIfBypassAllowed)
-        return
-      }
-
-      setShowRestoreBranchButton(true)
-    })
-  }, [branchError])
-
-  const [activities, setActivities] = useState<TypesPullReqActivity[] | undefined>(activityData)
-  const approvedEvaluations = reviewers?.filter(evaluation => evaluation.review_decision === 'approved')
-  const latestApprovalArr = approvedEvaluations?.filter(
-    approved => !checkIfOutdatedSha(approved.sha, pullReqMetadata?.source_sha as string)
-  )
-  const changeReqEvaluations = reviewers?.filter(evaluation => evaluation.review_decision === 'changereq')
-  const changeReqReviewer =
-    changeReqEvaluations && !isEmpty(changeReqEvaluations)
-      ? capitalizeFirstLetter(
-          changeReqEvaluations[0].reviewer?.display_name || changeReqEvaluations[0].reviewer?.uid || ''
-        )
-      : 'Reviewer'
-  const codeOwnerChangeReqEntries = findChangeReqDecisions(
-    codeOwners?.evaluation_entries,
-    CodeOwnerReqDecision.CHANGEREQ
-  )
-  const codeOwnerPendingEntries = findWaitingDecisions(codeOwners?.evaluation_entries)
-  const codeOwnerApprovalEntries = findChangeReqDecisions(codeOwners?.evaluation_entries, CodeOwnerReqDecision.APPROVED)
-  const latestCodeOwnerApprovalArr = codeOwnerApprovalEntries
-    ?.map(entry => {
-      // Filter the owner_evaluations for 'changereq' decisions
-      const entryEvaluation = entry?.owner_evaluations.filter(
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (evaluation: any) => !checkIfOutdatedSha(evaluation?.review_sha, pullReqMetadata?.source_sha as string)
-      )
-      // If there are any 'changereq' decisions, return the entry along with them
-      if (entryEvaluation && entryEvaluation?.length > 0) {
-        return { entryEvaluation }
-      }
-    }) // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    .filter((entry: any) => entry !== null && entry !== undefined) // Filter out the null entries
-
-  useEffect(() => {
-    refetchCodeOwners()
-  }, [pullReqMetadata, pullReqMetadata?.title, pullReqMetadata?.state, pullReqMetadata?.source_sha, refetchCodeOwners])
-
-  useEffect(() => {
-    setActivities(activityData)
-  }, [activityData])
-
-  useEffect(() => {
-    if (!commentId || isScrolledToComment || prPanelData.PRStateLoading || activityData?.length === 0) return
-    // Slight timeout so the UI has time to expand/hydrate
-    const timeoutId = setTimeout(() => {
-      const elem = document.getElementById(`comment-${commentId}`)
-      if (!elem) return
-      elem.scrollIntoView({ behavior: 'smooth', block: 'center' })
-      setIsScrolledToComment(true)
-    }, 500)
-
-    return () => {
-      clearTimeout(timeoutId)
-    }
-  }, [commentId, isScrolledToComment, prPanelData.PRStateLoading, activityData])
-
-  const changesInfo = extractInfoForCodeOwnerContent({
-    approvedEvaluations,
-    reqNoChangeReq: prPanelData?.atLeastOneReviewerRule,
-    reqCodeOwnerApproval: prPanelData?.reqCodeOwnerApproval,
-    minApproval: prPanelData?.minApproval,
-    reqCodeOwnerLatestApproval: prPanelData?.reqCodeOwnerLatestApproval,
-    minReqLatestApproval: prPanelData?.minReqLatestApproval,
-    codeOwnerChangeReqEntries,
-    codeOwnerPendingEntries,
-    latestCodeOwnerApprovalArr,
-    latestApprovalArr,
-    codeOwnerApprovalEntries,
-    changeReqReviewer,
-    changeReqEvaluations
-  })
-
-  useEffect(() => {
-    if (
-      !prPanelData?.PRStateLoading &&
-      changesInfo?.title !== '' &&
-      changesInfo?.statusMessage !== '' &&
-      changesInfo?.statusIcon !== ''
-    ) {
-      setChangesLoading(false)
-    } else if (pullReqMetadata?.merged) {
-      setChangesLoading(false)
-    }
-  }, [
-    changesInfo?.title,
-    changesInfo?.statusMessage,
-    changesInfo?.statusIcon,
-    prPanelData?.PRStateLoading,
-    pullReqMetadata?.merged
-  ])
-
-  const handleAddReviewer = (id?: number) => {
-    reviewerAddPullReq({ repo_ref: repoRef, pullreq_number: prId, body: { reviewer_id: id } })
-      .then(() => refetchReviewers())
-      .catch(error => setAddReviewerError(error.message))
-  }
-  const handleDeleteReviewer = (id: number) => {
-    reviewerDeletePullReq({ repo_ref: repoRef, pullreq_number: prId, pullreq_reviewer_id: id })
-      .then(() => refetchReviewers())
-      .catch(error => setRemoveReviewerError(error.message))
-  }
-
-  const onPRStateChanged = useCallback(() => {
-    refetchCodeOwners()
-    refetchPullReq()
-    refetchActivities()
-  }, [refetchCodeOwners, repoRef, handleAddReviewer, handleDeleteReviewer]) // eslint-disable-line react-hooks/exhaustive-deps
-
-  const handleMerge = (method: string) => {
-    const payload: OpenapiMergePullReq = {
-      method: method as EnumMergeMethod,
-      source_sha: pullReqMetadata?.source_sha,
-      bypass_rules: checkboxBypass,
-      dry_run: false
-      // message: data.commitMessage
-    }
-    mergePullReqOp({ body: payload, repo_ref: repoRef, pullreq_number: prId }).then(() => {
-      onPRStateChanged()
-      setRuleViolationArr(undefined)
-    })
-    //todo: add catch t o show errors
-    // .catch(exception => showError(getErrorMessage(exception)))
-  }
-
-  const handlePrState = (state: string) => {
-    const payload: OpenapiStatePullReqRequest = {
-      ...(state === 'draft' && { is_draft: true }),
-      state: state === 'draft' ? 'open' : (state as EnumPullReqState)
-    }
-    statePullReq({ body: payload, repo_ref: repoRef, pullreq_number: prId }).then(() => {
-      onPRStateChanged()
-      setRuleViolationArr(undefined)
-    })
-  }
-
-  const handleRebaseBranch = () => {
-    const payload: RebaseBranchRequestBody = {
-      base_branch: pullReqMetadata?.target_branch,
-      bypass_rules: true,
-      dry_run_rules: false,
-      head_branch: pullReqMetadata?.source_branch,
-      head_commit_sha: pullReqMetadata?.source_sha
-    }
-
-    rebaseBranch({ body: payload, repo_ref: repoRef }).then(
-      () => {
-        onPRStateChanged()
-        setRuleViolationArr(undefined)
-      },
-      error => {
-        setRebaseErrorMessage(error.message)
-      }
-    )
-  }
-
-  const mockPullRequestActions = [
+const getMockPullRequestActions = (
+  handlePrState: (data: string) => void,
+  handleMerge: (data: EnumMergeMethod) => void,
+  pullReqMetadata?: TypesPullReq
+) => {
+  return [
     ...(pullReqMetadata?.closed
       ? [
           {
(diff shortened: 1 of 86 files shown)


packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
```diff
@@ -28,78 +30,79 @@ const ReviewersHeader = ({
...
-  const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setSearchQuery(event.target.value)
-  }
+
+  const { search, handleSearchChange, handleResetSearch } = useDebounceSearch({
+    handleChangeSearchValue: setSearchQuery,
+    searchValue: searchQuery
+  })
+
+  const handleCloseValuesView = useRef(debounce(() => handleResetSearch(), 300)).current
```
knagurski, line 39: Should be safe to remove the inline anonymous function

```suggestion
  const handleCloseValuesView = useRef(debounce(handleResetSearch, 300)).current
```

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
```diff
@@ -28,78 +30,79 @@ const ReviewersHeader = ({
...
-          <div
-            className="relative flex items-center justify-between border-b border-borders-1 px-3 py-2.5"
-            role="presentation"
-            onKeyDown={e => e.stopPropagation()}
+    <div className="mb-0.5 flex items-center justify-between">
+      <h5 className="text-14 font-medium text-foreground-1">{t('views:pullRequests.reviewers', 'Reviewers')}</h5>
+
+      <DropdownMenu.Root onOpenChange={isOpen => !isOpen && handleCloseValuesView()}>
+        <DropdownMenu.Trigger asChild>
+          <Button
```
knagurski, line 47: I haven't checked the output here, but does `DropdownMenu.Trigger` (with `asChild`) pass down something that screen readers can see? If not, can we add a hidden bit of text as a sibling of the icon (e.g. `<div className="sr-only">Something to describe this button</div>`) or an `aria-label` to the `Button`?

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
```diff
@@ -28,78 +30,79 @@ const ReviewersHeader = ({
...
-            )}
-            <div className="max-h-[360px] overflow-y-auto">
-              {usersList?.length === 1 && usersList[0].uid === currentUserId ? (
-                <div className="px-5 py-4 text-center">
-                  <span className="text-14 leading-tight text-foreground-2">{t('views:pullRequests.noUsers')}</span>
-                </div>
-              ) : (
-                usersList?.map(({ display_name, id, uid }) => {
+          {!usersList?.length && (
+            <span className="block px-5 py-4 text-center leading-tight text-foreground-2">
```
knagurski, line 68: Instead of turning an inline element into a block, can we just use a block element?
```suggestion
            <div className="px-5 py-4 text-center leading-tight text-foreground-2">
```

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
```diff
@@ -28,78 +30,79 @@ const ReviewersHeader = ({
...
+              {t('views:pullRequests.noUsers', 'No users found.')}
+            </span>
+          )}
+
+          <ScrollArea viewportClassName="max-h-80">
+            {usersList?.length === 1 && usersList[0].uid === currentUserId ? (
+              <span className="block px-5 py-4 text-center leading-tight text-foreground-2">
```
knagurski, line 75: Same here
```suggestion
              <div className="px-5 py-4 text-center leading-tight text-foreground-2">
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/panel-accordion-show-button.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const PanelAccordionShowButton: FC<PanelAccordionShowButtonProps> = ({
+  isShowButton,
+  value,
+  accordionValues
+}) => {
+  if (!isShowButton) return <></>
+
+  return (
+    <span className="self-start px-2 text-14 text-foreground-2 transition-colors duration-200 group-hover:text-foreground-1">
+      Show {accordionValues.includes(value) ? 'less' : 'more'}
```
knagurski, line 18: Can we pull out the text here?

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -9,10 +11,65 @@ import {
...
+    default:
+      return <Icon name="success" className="text-foreground-success" />
+  }
+}
+
+interface HeaderItemProps {
+  header: string
+}
+
+const HeaderItem: FC<HeaderItemProps> = ({ header }: HeaderItemProps) => {
```
knagurski, line 36: `FC` gives you the type already
```suggestion
const HeaderItem: FC<HeaderItemProps> = ({ header }) => {
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -9,10 +11,65 @@ import {
...
+  return <span className="text-12 text-foreground-1">{header}</span>
+}
+
+interface AvatarItemProps {
+  evaluations: TypesOwnerEvaluation[] | undefined
+}
+
+const AvatarItem: FC<AvatarItemProps> = ({ evaluations }: AvatarItemProps) => {
```
knagurski, line 44: Same here
```suggestion
const AvatarItem: FC<AvatarItemProps> = ({ evaluations }) => {
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -9,10 +11,65 @@ import {
...
+              if (idx < 2) {
+                return (
+                  <Avatar.Root key={owner?.id || idx}>
+                    <Avatar.Fallback>{getInitials(owner?.display_name || '')}</Avatar.Fallback>
+                  </Avatar.Root>
+                )
+              }
+              if (idx === 2 && evaluations.length && evaluations.length > 2) {
+                // TODO: do popover with all the names
+                return <span key={owner?.id} className="text-12">{`+${evaluations.length - 2}`}</span>
```
knagurski, line 61: Template string seems unnecessary here.
```suggestion
                return <span key={owner?.id} className="text-12">+{evaluations.length - 2}</span>
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
+        if (
+          reqCodeOwnerLatestApproval &&
+          minReqLatestApproval &&
+          latestCodeOwnerApprovalArr &&
+          latestCodeOwnerApprovalArr?.length < minReqLatestApproval
         )
+          return {
+            icon: <Icon name="pending-clock" className="text-icons-alert" />,
+            text: 'Latest changes are pending approval from required reviewers'
+          }
```
knagurski, line 176: Can we add brackets on this. For readability, only omit brackets when the return is on the same line as a the condition
```suggestion
        ) {
          return {
            icon: <Icon name="pending-clock" className="text-icons-alert" />,
            text: 'Latest changes are pending approval from required reviewers'
          }
        }
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
-      default:
-        return <Icon name="success" className="text-foreground-success" />
-    }
-  }
+  const renderCodeOwnerStatus = useMemo(() => {
+    const getData = () => {
+      if (!!codeOwnerPendingEntries?.length && reqCodeOwnerLatestApproval)
+        return {
+          icon: <Icon name="circle" className="text-warning" />,
+          text: 'Waiting on code owner reviews of latest changes'
```
knagurski, line 139: Can we extract these to the strings files?

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
-        return <Icon name="pending-clock" className="text-icons-alert" />
-      case 'warning':
-        return <Icon name="triangle-warning" className="text-tertiary-background" />
-      case 'error':
-        return <Icon name="triangle-warning" className="text-destructive" />
-      default:
-        return <Icon name="success" className="text-foreground-success" />
-    }
-  }
+  const renderCodeOwnerStatus = useMemo(() => {
```
knagurski, line 134: `renderCodeOwnerStatus` sounds like a function, but this is a value. Maybe it would be better to from the `render` part.
```suggestion
  const codeOwnerStatus = useMemo(() => {
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
+    const getData = () => {
+      if (!!codeOwnerPendingEntries?.length && reqCodeOwnerLatestApproval)
+        return {
+          icon: <Icon name="circle" className="text-warning" />,
+          text: 'Waiting on code owner reviews of latest changes'
+        }
```
knagurski, line 140: Include brackets when the return isn't on the same line as the condition
```suggestion
      if (!!codeOwnerPendingEntries?.length && reqCodeOwnerLatestApproval) {
        return {
          icon: <Icon name="circle" className="text-warning" />,
          text: 'Waiting on code owner reviews of latest changes'
        }
      }
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
-          <Icon name="circle" className="text-warning" />
-          <span className="text-14 text-foreground-1">Waiting on code owner reviews of latest changes</span>
-        </div>
-      )
-    }
+      if (!!codeOwnerPendingEntries?.length && reqCodeOwnerApproval)
+        return {
+          icon: <Icon name="circle" className="text-warning" />,
+          text: 'Changes are pending approval from code owners'
+        }
```
knagurski, line 146:
```suggestion
      if (!!codeOwnerPendingEntries?.length && reqCodeOwnerApproval) {
        return {
          icon: <Icon name="circle" className="text-warning" />,
          text: 'Changes are pending approval from code owners'
        }
      }
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
-          <Icon name="circle" className="text-warning" />
-          <span className="text-14 text-foreground-1">Changes are pending approval from code owners</span>
-        </div>
-      )
-    }
+      if (!!codeOwnerApprovalEntries?.length && !!codeOwnerPendingEntries?.length)
+        return {
+          icon: <Icon name="circle" className="text-tertiary-background" />,
+          text: 'Some changes were approved by code owners'
+        }
```
knagurski, line 152:
```suggestion
      if (!!codeOwnerApprovalEntries?.length && !!codeOwnerPendingEntries?.length) {
        return {
          icon: <Icon name="circle" className="text-tertiary-background" />,
          text: 'Some changes were approved by code owners'
        }
      }
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
-            <Icon name="pending-clock" className="text-icons-alert" />
-            <span className="text-14 text-foreground-1">
-              Latest changes are pending approval from required reviewers
-            </span>
-          </div>
+      if (!!latestCodeOwnerApprovalArr?.length && reqCodeOwnerLatestApproval)
+        return {
+          icon: <Icon name="success" className="text-foreground-success" />,
+          text: 'Latest changes were approved by code owners'
+        }
```
knagurski, line 158:
```suggestion
      if (!!latestCodeOwnerApprovalArr?.length && reqCodeOwnerLatestApproval) {
        return {
          icon: <Icon name="success" className="text-foreground-success" />,
          text: 'Latest changes were approved by code owners'
        }
      }
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
```diff
@@ -103,105 +125,77 @@ const PullRequestChangesSection = ({
...
+
+      if (!!codeOwnerApprovalEntries?.length && reqCodeOwnerApproval)
+        return {
+          icon: <Icon name="success" className="text-foreground-success" />,
+          text: 'Changes were approved by code owners'
+        }
```
knagurski, line 164:
```suggestion
      if (!!codeOwnerApprovalEntries?.length && reqCodeOwnerApproval) {
        return {
          icon: <Icon name="success" className="text-foreground-success" />,
          text: 'Changes were approved by code owners'
        }
      }
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-comment-section.tsx
```diff
@@ -7,32 +7,33 @@ interface PullRequestMergeSectionProps {
...
-        {commentsInfo.status === 'failed' && (
-          <Text
-            onClick={() => {
-              handleAction?.()
-            }}
-            className="pr-2"
-            size={1}
-          >
+        {commentsInfo.status === 'failed' && !!handleAction && (
+          <Text className="pr-2" onClick={() => handleAction?.()} size={1}>
```
knagurski, line 36: Probably can drop the optional chaining here since its existence was proven on the line before
```suggestion
          <Text className="pr-2" onClick={() => handleAction()} size={1}>
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -1,25 +1,57 @@
...
 
-import { Accordion, CopyButton, Icon, Layout, StackedList } from '@/components'
+import { Accordion, Button, CopyButton, Icon, Layout, StackedList } from '@/components'
 import { cn } from '@utils/cn'
+import { PanelAccordionShowButton } from '@views/repo/pull-request/details/components/conversation/sections/panel-accordion-show-button'
 import { isEmpty } from 'lodash-es'
 
 import { LineDescription, LineTitle } from './pull-request-line-title'
 
+const stepInfo = (item: { step: string; description: string; code?: string; comment?: string }, index: number) => {
```
knagurski, line 10: This really is a component
```suggestion
interface StepInfoProps {
  item: {
    step: string
    description: string
    code?: string
    comment?: string
  }
  index: number
}

const StepInfo: FC<StepInfoProps> = ({ item, index }) => {
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -49,145 +81,99 @@ const PullRequestMergeSection = ({
...
+          className="flex gap-y-1"
+          title={
+            <LineTitle
+              textClassName={isConflicted ? 'text-foreground-danger' : ''}
+              text={
+                unchecked
+                  ? 'Merge check in progress...'
+                  : !mergeable
+                    ? 'Conflicts found in this branch'
+                    : `This branch has no conflicts with ${pullReqMetadata?.target_branch} branch`
```
knagurski, line 109: Can we extract things to the string files.

packages/ui/src/views/repo/pull-request/details/pull-request-details-types.ts
```diff
@@ -490,6 +490,12 @@ export interface PullRequestRoutingProps {
   toPRCheck: ({ pipelineId, executionId }: { pipelineId: string; executionId: string }) => string
 }
 
+export type LatestCodeOwnerApprovalArrType =
```
knagurski, line 493: Do we need this to be exported?

packages/ui/src/views/repo/pull-request/details/pull-request-details-types.ts
```diff
@@ -521,12 +527,7 @@ export interface PullRequestChangesSectionProps {
       }
     | undefined
```
knagurski, line 528: Looks like there's a bunch of stuff around that's both optional and allowed to be `undefined`. If it's optional, then it gets the `undefined` anyway, so we can simplify and kill off those `| undefined`

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-panel.tsx
```diff
@@ -135,14 +109,14 @@ const HeaderTitle = ({ ...props }: HeaderProps) => {
...
-  mergeable,
+  isMergeable,
   ruleViolation,
   isDraft,
   checksInfo,
   checkboxBypass,
   canBypass
 }: {
-  mergeable: boolean
+  isMergeable: boolean
```
knagurski, line 119: Probably best to make this optional like the other booleans in this.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-panel.tsx
```diff
@@ -187,6 +159,32 @@ const getButtonState = ({
...
+export interface PullRequestPanelProps
+  extends Omit<
+      PullRequestChangesSectionProps,
+      'reqNoChangeReq' | 'reqCodeOwnerApproval' | 'minApproval' | 'reqCodeOwnerLatestApproval' | 'minReqLatestApproval'
+    >,
+    Partial<PullRequestRoutingProps> {
+  handleRebaseBranch: () => void
+  handlePrState: (state: string) => void
+  pullReqMetadata?: TypesPullReq
+  checks?: TypesPullReqCheck[] | null
```
knagurski, line 171: So this can be `undefined`, `null` or an array of `TypesPullReqCheck`?

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-panel.tsx
```diff
@@ -219,18 +217,31 @@ const PullRequestPanel = ({
...
+    return {
+      isMergeable: pullReqMetadata?.merge_check_status === MergeCheckStatus.MERGEABLE,
+      isClosed,
+      isOpen,
+      isDraft,
+      isUnchecked: pullReqMetadata?.merge_check_status === MergeCheckStatus.UNCHECKED && !isClosed,
+      isRebasable: pullReqMetadata?.merge_target_sha !== pullReqMetadata?.merge_base_sha && !pullReqMetadata?.merged,
+      isShowMoreTooltip: isOpen && !isDraft
+    }
+  }, [pullReqMetadata])
```
knagurski, line 244: Since nothing in here is complex and it deals exclusively with booleans, it would probably be as fast to not memoize it and cut down on the memo overhead.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-panel.tsx
```diff
@@ -335,73 +293,107 @@ const PullRequestPanel = ({
...
-                    </Button>
-                  )}
+                {actions && !pullReqMetadata?.closed ? (
+                  <ButtonWithOptions
+                    id="pr-type"
+                    theme={buttonState.theme as ButtonWithOptionsTheme}
+                    disabled={buttonState.disabled}
+                    variant={buttonState?.variant as ButtonWithOptionsVariant}
+                    selectedValue={mergeButtonValue}
+                    handleOptionChange={value => setMergeButtonValue(value)}
```
knagurski, line 334: You can probably simplify here
```suggestion
                    handleOptionChange={setMergeButtonValue}
```

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
3em: Yes, Radix passes the same attributes as it sets itself when using `DropdownMenu.Trigger` without `asChild`. It sets `aria-haspopup="menu"` and `aria-expanded="false"`

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/panel-accordion-show-button.tsx
3em: Can you please clarify what you mean?
The current text depends on whether the given value (which is the ID of the accordion item) is in the array of open items.
Translations via i18n were planned to be added in a separate PR.

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-changes-section.tsx
3em: Yes, that's the plan, but in a separate PR to avoid bloating the current one.

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
3em: Yes, that's the plan, but in a separate PR to avoid bloating the current one.

packages/ui/src/views/repo/pull-request/details/pull-request-details-types.ts
3em: Yes, this type is used in `app/gitness` for typing data in `reduce`.

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-panel.tsx
3em: If we believe the interface that Lee wrote, then it's exactly like that
image

packages/ui/src/views/repo/pull-request/details/pull-request-details-types.ts
3em: I tried to remove it, but unfortunately, the `app/gitness` `pnpm typecheck` fails – apparently, there is still some data mapper there that can return `undefined` as a value inside the array.

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
knagurski: Had a look and this will not add anything that a screen reader can read out to describe the button. The usage in the Radix demos use `aria-label` to describe things. The `aria-haspopup` and `aria-expanded` properties just tell the screen reader how to describe the interaction, not what the interaction is for.
Screenshot 2025-03-20 at 11 27 26

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/panel-accordion-show-button.tsx
knagurski: I was just referring to the translations stuff. No problem if it'll be addressed in another PR 👍🏻

packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-panel.tsx
knagurski: Annoying that we're not generating the types from what the backend publishes instead of hand-coding them. ¯\\_(ツ)_/¯

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -49,145 +91,103 @@ const PullRequestMergeSection = ({
...
+              </div>
+            )}
+            <span className="text-14 text-foreground-2">
+              Conflicting files <span className="text-foreground-4">{`(${conflictingFiles?.length || 0})`}</span>
+            </span>
+
+            {!isEmpty(conflictingFiles) && (
+              <div className="mt-1">
+                {conflictingFiles?.map((file, idx) => (
+                  <div className="flex items-center gap-x-2 py-1.5" key={`${file}-${idx}`}>
```
knagurski, line 180: I would've thought `file` would be unique
```suggestion
                {conflictingFiles?.map(file => (
                  <div className="flex items-center gap-x-2 py-1.5" key={file}>
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -49,145 +91,103 @@ const PullRequestMergeSection = ({
...
-            </div>
-          </Accordion.Content>
-        )}
-      </Accordion.Item>
-    </Accordion.Root>
+                </ol>
+              </div>
+            )}
+            <span className="text-14 text-foreground-2">
+              Conflicting files <span className="text-foreground-4">{`(${conflictingFiles?.length || 0})`}</span>
```
knagurski, line 174: No need for a template string here I think
```suggestion
              Conflicting files <span className="text-foreground-4">({conflictingFiles?.length || 0})</span>
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -49,145 +91,103 @@ const PullRequestMergeSection = ({
...
+          }
+          description={
+            <>
+              {unchecked && <LineDescription text={'Checking for ability to merge automatically...'} />}
+              {isConflicted && (
+                <Layout.Vertical className="ml-6">
+                  <p className="text-14 font-normal text-foreground-4">
+                    Use the&nbsp;
+                    <Button variant="link_accent" size="text" onClick={handleCommandLineClick} asChild>
+                      <span>command line</span>
```
knagurski, line 141: Curious why we'd need the empty `span`?
```suggestion
                      command line
```

With the `asChild`, it would assign a click handler to a `span` which would be terrible for accessibility.

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -49,145 +91,103 @@ const PullRequestMergeSection = ({
...
 
-  const onValueChange = (value: string | string[]) => {
-    if (typeof value === 'string') {
-      setValue(value)
-    }
+  const handleCommandLineClick = (e: MouseEvent<HTMLButtonElement>) => {
+    e.stopPropagation()
+
+    setAccordionValues(prevState => [...prevState, ACCORDION_VALUE])
+    setShowCommandLineInfo(!showCommandLineInfo)
```
knagurski, line 99: For toggling state, it's usually a good idea to use the callback version
```suggestion
    setShowCommandLineInfo(current => !current)
```

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -1,25 +1,67 @@
...
+    description: string
+    code?: string
+    comment?: string
+  }
+  index: number
+}
+
+const StepInfo: FC<StepInfoProps> = ({ item, index }) => {
+  return (
+    <li key={index}>
```
knagurski, line 22: Looking at it, I don't think you need a `key` at all here since the loop is done elsewhere.
```suggestion
}

const StepInfo: FC<StepInfoProps> = ({ item }) => {
  return (
    <li>
```

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
```diff
@@ -28,78 +30,79 @@ const ReviewersHeader = ({
...
-            )}
-            <div className="max-h-[360px] overflow-y-auto">
-              {usersList?.length === 1 && usersList[0].uid === currentUserId ? (
-                <div className="px-5 py-4 text-center">
-                  <span className="text-14 leading-tight text-foreground-2">{t('views:pullRequests.noUsers')}</span>
-                </div>
-              ) : (
-                usersList?.map(({ display_name, id, uid }) => {
+            ) : (
+              <>
```
knagurski, line 79: I'm not sure you need this fragment

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
```diff
@@ -28,78 +30,79 @@ const ReviewersHeader = ({
...
+          <Button
+            className="text-icons-1 hover:text-icons-2 data-[state=open]:text-icons-2"
+            size="xs_icon"
+            variant="custom"
           >
+            <Icon name="vertical-ellipsis" size={12} />
+          </Button>
+        </DropdownMenu.Trigger>
+        <DropdownMenu.Content className="w-80" align="end" sideOffset={-6} alignOffset={10}>
+          <div className="px-2 py-1.5" role="presentation" onKeyDown={e => e.stopPropagation()}>
```
knagurski, line 56: It's annoying that React doesn't have things like Vue and Svelte's event modifiers to handle things like `stopPropagation` etc

cjlee01:
image

should it also show time like that 14:12?

cjlee01:
when adding reviewers or deleting reviewers it should trigger refetch in overview similiar to how labels worked?

https://github.com/user-attachments/assets/0c50ac92-a185-4f6d-9c23-fa9eb508c1c6

apps/gitness/src/pages/pull-request/utils.ts
```diff
@@ -187,6 +187,28 @@ export function extractSpecificViolations(violationsData: any, rule: string) {
...
+  requiresCommentApproval: boolean
+  resolvedCommentArrParams?: number[]
+}) => {
+  const resolvedComments = requiresCommentApproval && !resolvedCommentArrParams
+
+  if (resolvedComments) {
+    return { header: 'All comments are resolved', content: undefined, status: 'success' }
+  }
+
+  const unresolvedCount = resolvedCommentArrParams?.length || 0 // Ensure a default value
```
cjlee01, line 203: shouldn't be length since its a number array should be should something like resolvedCommentArrparams[0]

since object is like 
[2] for resolvedCommentArrParams so right now it'l only ever return 1 if true

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
3em: Because we are using the Button with the asChild prop, and we need to pass an Element so that all the button's properties are applied to it.

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
3em: yep

3em:
> image while looking over pr, should this reply box not be full width? @3em

@cjlee01  Yes, this will be fixed in one of the upcoming PRs. This PR only addressed work on two components: Panel and Sidebar, and even then, only partially on the Sidebar. There are many issues on this page that will still be fixed.

3em:
> when adding reviewers or deleting reviewers it should trigger refetch in overview similiar to how labels worked? @3em
> 
>  Screen.Recording.2025-03-20.at.3.03.09.PM.mov

@cjlee01  Yes, I also noticed this issue, but as I mentioned earlier, unfortunately, it’s impossible to fix everything in one PR, otherwise there would be too many changes and it would be hard to review. So, this will be fixed in the upcoming PRs.

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
knagurski: For accessibility reasons, we should _never_ attach click handlers to things that aren't buttons or anchors. Using `asChild` here makes a `span` clickable, with no indication to assistive technologies that it can be clicked. It won't be read out to screen readers and cannot be focused by keyboard users. Below is what the screen reader would see for this kind of element; it is seen as a "generic" element.  
Screenshot 2025-03-24 at 10 56 23

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
```diff
@@ -95,23 +173,23 @@ export default function PullRequestConversationPage() {
...
-  const [comment, setComment] = useState<string>('')
-  const [commentId] = useQueryState('commentId')
-  const [isScrolledToComment, setIsScrolledToComment] = useState(false)
-
   const repoRef = useGetRepoRef()
   const { pullRequestId } = useParams<PathParams>()
-  const prId = (pullRequestId && Number(pullRequestId)) || -1
-  // const [loadState, setLoadState] = useState('data-loaded')
+
+  const prId = useMemo(() => (pullRequestId && Number(pullRequestId)) || -1, [pullRequestId])
```
knagurski, line 192: Probably no need to memoize this. The memo storage and look up will probably take longer than the action itself.

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
```diff
@@ -195,31 +278,35 @@ export default function PullRequestConversationPage() {
...
         if (res?.body?.rule_violations) {
           const { checkIfBypassAllowed } = extractInfoFromRuleViolationArr(res.body?.rule_violations)
-          if (checkIfBypassAllowed) {
-            setShowDeleteBranchButton(true)
-          } else {
-            setShowDeleteBranchButton(false)
-            setShowRestoreBranchButton(true)
-          }
+
+          checkIfBypassAllowed ? setShowDeleteBranchButton(true) : onSuccessDeleteCommon()
```
knagurski, line 298: I'm not a fan of using ternaries as a substitute for `if` when invoking functions.
```suggestion
          if (checkIfBypassAllowed) {
            setShowDeleteBranchButton(true)
          } else {
            onSuccessDeleteCommon()
          }
```

Not a blocker, I just don't find it as immediately readable. Whenever I see then, I look for the assignment.

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
```diff
@@ -346,137 +492,64 @@ export default function PullRequestConversationPage() {
...
-
-    rebaseBranch({ body: payload, repo_ref: repoRef }).then(
-      () => {
-        onPRStateChanged()
+  }, [refetchCodeOwners, refetchPullReq, refetchActivities])
+
+  const handleMerge = useCallback(
+    (method: string) => {
+      const payload: OpenapiMergePullReq = {
+        method: method as EnumMergeMethod,
```
knagurski, line 523: You can get rid of `as EnumMergeMethod` by just setting the arg as that type (with a corresponding change in `getMockPullRequestActions`.

Also, `EnumMergeMethod` really annoys me since it is not an enum, but just a type.
```suggestion
    (method: EnumMergeMethod) => {
      const payload: OpenapiMergePullReq = {
        method,
```

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
```diff
@@ -346,137 +492,64 @@ export default function PullRequestConversationPage() {
...
+        // message: data.commitMessage
+      }
+      mergePullReqOp({ body: payload, repo_ref: repoRef, pullreq_number: prId }).then(() => {
+        handleRefetchData()
         setRuleViolationArr(undefined)
-      },
-      error => {
-        setRebaseErrorMessage(error.message)
+      })
+      //todo: add catch t o show errors
```
knagurski, line 533: lol, little typo

```suggestion
      //todo: add catch to show errors
```

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
```diff
@@ -346,137 +492,64 @@ export default function PullRequestConversationPage() {
...
+      // .catch(exception => showError(getErrorMessage(exception)))
+    },
+    [pullReqMetadata?.source_sha, checkboxBypass, repoRef, prId, handleRefetchData, setRuleViolationArr]
+  )
+
+  const handlePrState = useCallback(
+    (state: string) => {
+      const payload: OpenapiStatePullReqRequest = {
+        ...(state === 'draft' && { is_draft: true }),
+        state: state === 'draft' ? 'open' : (state as EnumPullReqState)
```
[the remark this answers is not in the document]
knagurski: Also, I'm not really a fan of conditionally spreading inline objects into another object. Might be couple more lines, but it's _way_ more obvious what's going on...
```suggestion
    (state: string) => {
      const payload: OpenapiStatePullReqRequest = { state }
      
      if (state === 'draft') {
        payload.is_draft = true
        payload.state = 'open'
```

knagurski (changes requested):
Accessibility issues are a blocker.

packages/ui/src/views/repo/pull-request/components/reviewers/pull-request-reviewers-header.tsx
3em: I added aria-label

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
3em: Due to the fact that the current text is already rendered in a button, I can't use a button again, so I'm using a span. However, for screen readers, I've added `role`, `tabIndex`, `aria-label`, and `onKeyDown`

apps/gitness/src/pages-v2/pull-request/pull-request-conversation.tsx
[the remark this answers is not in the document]
3em: I can't make `state` of type `EnumPullReqState` because a `string` can be passed. Below is the check `state === 'draft'`, and `'draft'` is not a value from `EnumPullReqState`.

3em:
> Accessibility issues are a blocker.

Fixed

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
knagurski: Any reason to not just use without `asChild` and render a normal button?

packages/ui/src/views/repo/pull-request/details/components/conversation/sections/pull-request-merge-section.tsx
```diff
@@ -49,145 +88,115 @@ const PullRequestMergeSection = ({
...
+            />
+          }
+          description={
+            <>
+              {unchecked && <LineDescription text={'Checking for ability to merge automatically...'} />}
+              {isConflicted && (
+                <Layout.Vertical className="ml-6">
+                  <p className="text-14 font-normal text-foreground-4">
+                    Use the&nbsp;
+                    <Button variant="link_accent" size="text" onClick={handleCommandLineClick} asChild>
```
knagurski, line 137: Any reason we couldn't just kill off the child and drop the `asChild` entirely? Instead of trying to get a span to act like a button, can we just use the button?

Screenshot 2025-03-25 at 15 22 03

knagurski (approved):
Bummer about the button in a button. We should perhaps revisit the UX on having an interactive element within an interactive element, but as far as coding around that, this solution looks good to me, just a shame we have to do it this way.

PR #1271, opened 2025-03-19
fix: Exclude `core-design-system` pkg from pnpm build and publish:all scripts

vardanbansal-harness:
During `pnpm publish:all`, we're getting error `ERR_PNPM_GIT_UNCLEAN  Unclean working tree. Commit or stash changes first.`

This PR excludes the package `core-design-system` from `pnpm build` and `pnpm publish:all` to prevent building the package.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,9 +18,6 @@ jobs:
           pnpm install
           pnpm build
 
-      - name: Reset uncommitted changes
-        run: git checkout -- .
-
       - name: Authenticate to npm
         run: |
           echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_PUBLISH_TOKEN }}" > ~/.npmrc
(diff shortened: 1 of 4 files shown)


vardanbansal-harness:
Fails without the fix: https://github.com/harness/canary/actions/runs/13955040713/job/39063884367

Works with the fix: https://github.com/harness/canary/actions/runs/13955207400/job/39064436226

.github/workflows/publish.yml
```diff
@@ -18,6 +18,9 @@ jobs:
           pnpm install
           pnpm build
 
+      - name: Reset uncommitted changes
+        run: git checkout -- .
```
abhinavrastogi-harness, line 22: why is this needed? there should not be any uncommitted changes resulting from just running install and build.

vardanbansal-harness: Removed it.

PR #1272, opened 2025-03-19
feat: Project scoped branch rules page 

sans-harness:
https://github.com/user-attachments/assets/9ad3d4ec-2cfa-4b92-bdca-ed00ee51100c

### DIFF
--- a/apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
+++ b/apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
@@ -33,7 +33,7 @@ export const ProjectRulesCreateOrUpdateContainer = () => {
   const { ruleId: ruleIdentifier } = useParams()
   const { setPresetRuleData, setPrincipals, setRecentStatusChecks } = useProjectRulesStore()
   const [principalsSearchQuery, setPrincipalsSearchQuery] = useState('')
-  const { dispatch } = useBranchRulesStore()
+  const { dispatch, resetRules } = useBranchRulesStore()
   const [isSubmitSuccess, setIsSubmitSuccess] = useState<boolean>()
   const {
     scope: { accountId }
@@ -51,8 +51,9 @@ export const ProjectRulesCreateOrUpdateContainer = () => {
       setPresetRuleData(null)
       setPrincipals(null)
       setRecentStatusChecks(null)
+      resetRules()
     }
-  }, [setPresetRuleData, setPrincipals, setRecentStatusChecks])
+  }, [resetRules, setPresetRuleData, setPrincipals, setRecentStatusChecks])
 
   const { data: { body: rulesData } = {} } = useSpaceRuleGetQuery(
     { space_ref: spaceURL ?? '', rule_identifier: ruleIdentifier ?? '' },
--- a/apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
+++ b/apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
@@ -8,32 +8,25 @@ import {
   useSpaceRuleGetQuery,
   useSpaceRuleUpdateMutation
 } from '@harnessio/code-service-client'
-import { SkeletonForm } from '@harnessio/ui/components'
 import { PrincipalType } from '@harnessio/ui/types'
 import {
   BranchRulesActionType,
   getBranchRules,
   MergeStrategy,
-  NotFoundPage,
   RepoBranchSettingsFormFields,
   RepoBranchSettingsRulesPage
 } from '@harnessio/ui/views'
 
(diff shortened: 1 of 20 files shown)


apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
```diff
@@ -0,0 +1,208 @@
...
+  RepoBranchSettingsFormFields,
+  RepoBranchSettingsRulesPage
+} from '@harnessio/ui/views'
+
+import { useGetSpaceURLParam } from '../../framework/hooks/useGetSpaceParam'
+import { useMFEContext } from '../../framework/hooks/useMFEContext'
+import { useTranslationStore } from '../../i18n/stores/i18n-store'
+import { transformDataFromApi, transformFormOutput } from '../../utils/repo-branch-rules-utils'
+import { useBranchRulesStore } from '../repo/stores/repo-branch-rules-store'
+import { useProjectRulesStore } from './stores/project-rules-store'
```
abhinavrastogi-harness, line 25: i see we added a `projectRulesStore`. why do we need to use `branchRulesStore` here from repoRules?

apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
```diff
@@ -0,0 +1,208 @@
...
+    {
+      onSuccess: () => {
+        setIsSubmitSuccess(true)
+        navigate('..')
+      }
+    }
+  )
+
+  const { data: { body: principals } = {}, error: principalsError } = useListPrincipalsQuery({
+    // @ts-expect-error : BE issue - not implemnted
```
abhinavrastogi-harness, line 79: did we file a BE ticket for this? let's link it here

apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const {
+    mutate: updateRule,
+    error: updateRuleError,
+    isLoading: updatingRule
+  } = useSpaceRuleUpdateMutation(
+    { space_ref: spaceURL, rule_identifier: ruleIdentifier! },
+    {
+      onSuccess: () => {
+        setIsSubmitSuccess(true)
+        navigate('..')
```
abhinavrastogi-harness, line 92: works with router v5?

apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
```diff
@@ -0,0 +1,208 @@
...
+      // Update existing rule
+      updateRule({
+        body: formattedData
+      })
+    } else {
+      // Add new rule
+      addRule({
+        body: formattedData
+      })
+    }
```
abhinavrastogi-harness, line 114: this can be potentially simplified for all use add/edit usecases:

```suggestion
    const addOrUpdate = ruleIdentifier ? updateRule : addRule
    addOrUpdate({
      body: formattedData
    })
```

apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
abhinavrastogi-harness: wait, I think I'm mixing a few things. maybe the naming is not ideal.
I see we have `repoRules`, `branchRules` and `projectRules`. How do they relate to each other?

apps/gitness/src/pages-v2/project/project-rules-list-container.tsx
abhinavrastogi-harness, line 1: hmm.. this look like largely a copy of repo rules list, with only the API hook changed.
would it be more efficient to just add an `if` condition on the same container to switch b/w different hooks?

apps/gitness/src/pages-v2/project/stores/project-rules-store.tsx
abhinavrastogi-harness, line 1: this doesn't seem to be congruent with `repo-branch-rules-store.tsx`. why so much difference b/w the two.
also, could we share the schema and just have two instances of it for project rules and repo rules?

apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
sans-harness: Yep, tested with MFE

apps/gitness/src/pages-v2/project/project-rules-create-container.tsx
sans-harness: Since the two functions have different path_params, TS is not happy with this.

abhinavrastogi-harness (approved):

abhinavrastogi-harness (approved):

PR #1289, opened 2025-03-24
Style fixes - Secrets Page

sans-harness:
Screenshot 2025-03-24 at 12 24 49 PM

Error and scroll fixes on secrets drawer.

### DIFF
--- a/packages/ui/src/views/platform/entity-reference-component.tsx
+++ b/packages/ui/src/views/platform/entity-reference-component.tsx
@@ -1,6 +1,6 @@
 import { useCallback } from 'react'
 
-import { Alert, Button, Icon, Input, ScrollArea, SkeletonList, StackedList } from '@/components'
+import { Button, Icon, Input, ScrollArea, SkeletonList, StackedList } from '@/components'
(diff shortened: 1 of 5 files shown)


packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -113,25 +119,24 @@ export function EntityReference<T extends BaseEntityProps, S = string, F = strin
       {isLoading ? (
         <SkeletonList />
       ) : (
-        <>
+        <div className="h-[60vh]">
```
abhinavrastogi-harness, line 122: i'd prefer to avoid hardcoding fixed/relative heights.
shared components should default to 100% of container height, and allow passing `className` as a prop for parent to override the height if needed

abhinavrastogi-harness (approved):

packages/ui/src/views/platform/entity-reference-component.tsx
sans-harness: Made parent container `h-full` but the scroll area needs a specific height, moved `h-[60vh]` to `ScrollArea`

PR #1292, opened 2025-03-24
fix: Fix scroll issue

vardanbansal-harness:
Fixes the scroll issue inside `Sheet` component as per known issue: https://github.com/radix-ui/primitives/issues/3353#issuecomment-2664085003

https://github.com/user-attachments/assets/c400861d-b00b-4b17-9d15-3f0a1d949b0c

#### Change overview:

Wrap Content with Overlay instead of placing them as siblings:

```
<DialogPrimitive.Overlay>
   <DialogPrimitive.Content>....</DialogPrimitive.Content>
</DialogPrimitive.Overlay>

### DIFF
--- a/packages/ui/src/components/sheet.tsx
+++ b/packages/ui/src/components/sheet.tsx
@@ -87,28 +87,29 @@ const SheetContent = React.forwardRef<React.ElementRef<typeof SheetPrimitive.Con
     const { portalContainer } = usePortal()
     return (
       <SheetPortal container={portalContainer}>
-        <SheetOverlay modal={modal} className={overlayClassName} handleClose={handleClose || props.onClick} />
-        <SheetPrimitive.Content ref={ref} className={cn(sheetVariants({ side }), className)} {...props}>
-          {children}
-          {!hideCloseButton && (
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/ui/package.json [hunk 1]
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@harnessio/ui",
   "description": "Harness Canary UI component library",
-  "version": "0.0.4",
+  "version": "0.0.5",
```
abhinavrastogi-harness, line 4: let's not bump version in the same PR as code changes.. 
we can combine multiple fixes into one new release

abhinavrastogi-harness (commented):
can you check if the Drawer component has the same issue and might need the same fix?

packages/ui/package.json
vardanbansal-harness: Yes makes sense, reverted.

vardanbansal-harness:
> can you check if the Drawer component has the same issue and might need the same fix?

I replaced `Sheet` with `Drawer` just to test and scroll works fine by default with `Drawer`.

sans-harness (approved):

PR #1297, opened 2025-03-25
Update sidebar design

ankormoreankor:
So so draft


packages/ui/src/components/app-sidebar/app-sidebar.tsx
```diff
@@ -155,26 +158,53 @@ export const AppSidebar = ({
               <Sidebar.Menu>
                 {!!currentUser?.admin && (
                   <Sidebar.MenuItem>
-                    <Sidebar.MenuButton asChild onClick={() => navigate('/admin/default-settings')}>
-                      <Sidebar.MenuItemText
-                        text={t('component:navbar.user-management', 'User Management')}
-                        icon={<Icon name="account" size={14} />}
-                      />
-                    </Sidebar.MenuButton>
+                    <button className="w-full" onClick={() => navigate('/admin/default-settings')}>
```
3em, line 161: It seems odd that we're wrapping a button component inside another button. Can we use `Sidebar.MenuButton` as a button directly?

packages/ui/src/components/app-sidebar/app-sidebar.tsx
```diff
@@ -155,26 +158,53 @@ export const AppSidebar = ({
...
+                          : t('component:navbar.sidebarToggle.collapse', 'Collapse')
+                      }
+                      text={t('component:navbar.sidebarToggle.collapse', 'Collapse')}
+                      icon={<Icon name={collapsed ? 'sidebar-right' : 'sidebar-left'} size={14} />}
+                    />
+                  </Sidebar.MenuButton>
+                </button>
+              </Sidebar.MenuItem>
+            </Sidebar.Menu>
+          </Sidebar.Group>
```
3em, line 205: Before, there was a check for useIsMFE to render the collapse button. What was it used for?

packages/ui/src/components/app-sidebar/sidebar-item.tsx
```diff
@@ -40,46 +41,37 @@ export const SidebarItem = ({
     handleRemoveRecentMenuItem(item)
   }
 
+  const dropdownItemClassNames =
+    'text-sidebar-foreground-6 data-[highlighted]:bg-sidebar-background-2 data-[highlighted]:text-sidebar-foreground-1'
```
3em, line 45: Move this constant to the top, outside of the component.

packages/ui/src/components/app-sidebar/sidebar-search-legacy.tsx
```diff
@@ -25,18 +27,37 @@ function SidebarSearchLegacy({ logo, t }: ProjectProps) {
...
-        onSearch={openSearchDialog}
-        handleChange={openSearchDialog}
-        theme="sidebar"
-      />
+      <div className="my-5 flex max-w-full items-center overflow-hidden pl-2">{logo}</div>
+      <div className="relative w-full">
+        <Button
+          variant="ghost"
+          tabIndex={collapsed ? 0 : -1}
+          aria-label="Open search dialog"
```
3em, line 35: add translation

packages/ui/src/components/sidebar/sidebar.tsx
```diff
@@ -619,11 +638,7 @@ const SidebarMenuSkeleton = forwardRef<
       <Skeleton
         className="h-4 max-w-[--skeleton-width] flex-1"
         data-sidebar="menu-skeleton-text"
-        style={
-          {
-            '--skeleton-width': width
-          } as CSSProperties
-        }
+        style={{ '--skeleton-width': width } as CSSProperties}
```
3em, line 641: we need `as CSSProperties` here?

packages/ui/src/components/app-sidebar/app-sidebar.tsx
ankormoreankor: This check is no longer needed since `AppSidebar` is used only in `AppShell`.
Breadcrumbs with a collapse button are also used in `AppShellMFE`, so that check was needed to hide it in MFE

packages/ui/src/components/sidebar/sidebar.tsx
ankormoreankor: Yes, you need type assertion when using css custom properties

packages/ui/src/components/app-sidebar/app-sidebar.tsx
ankormoreankor: This wrapper was added to make menu items accessible from a keyboard, cos Sidebar.MenuButton with asChild prop breaks it. After your comment, I found a better solution to fix it

3em: Yes, you did the same thing as I was talking about. Use `Sidebar.MenuButton` without `asChild`.

3em (approved):

sans-harness:
@ankormoreankor this PR completely breaks the MFE since we haven't wrapper the `AppMFE` in `sidebar-provider`. Also, changes to layouts  - `SandboxLayout`, `PRLayout` etc are breaking layouts across pages in the MFE. As a result I am  reverting some of the changes in this PR. Changes to layouts are usually dangerous and need more oversight before merging. 

cc: @abhinavrastogi-harness @vardanbansal-harness

PR #1303, opened 2025-03-25
feat: [pipe-26013]: add api error to rightdrawer

cjlee01:
add api error to connector rightdrawer as a prop

### DIFF
--- a/packages/ui/src/views/connectors/connector-entity-form.tsx
+++ b/packages/ui/src/views/connectors/connector-entity-form.tsx
@@ -1,6 +1,7 @@
 import { useMemo } from 'react'
 
 import { TranslationStore } from '@/views'
+import { Alert } from '@components/alert'
 import { Button } from '@components/button'
 import { EntityFormLayout } from '@views/unified-pipeline-studio/components/entity-form/entity-form-layout'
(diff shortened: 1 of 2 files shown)


packages/ui/src/views/connectors/connector-entity-form.tsx
```diff
@@ -96,6 +98,11 @@ export const ConnectorEntityForm = (props: ConnectorEntityFormProps): JSX.Elemen
             </EntityFormSectionLayout.Header>
             <EntityFormSectionLayout.Form>
               <RenderForm className="space-y-4" factory={inputComponentFactory} inputs={formDefinition} />
+              {apiError && (
+                <Alert.Container variant="destructive" className="my-8">
+                  <Alert.Description>{apiError?.toString()}</Alert.Description>
```
vardanbansal-harness, line 103: `apiError` is already checked in the conditional `({apiError && ...})`, so `apiError?.toString()` can be simplified to `apiError.toString()`.

vardanbansal-harness (approved):

packages/ui/src/views/connectors/connector-entity-form.tsx
cjlee01: fixed

PR #1311, opened 2025-03-27
feat: Added `Logo` component bases on `simple-icons` pkg

vardanbansal-harness:


packages/ui/src/components/logo/logo.tsx
```diff
@@ -0,0 +1,20 @@
...
+  name: keyof typeof LogoNameMap
+  size?: number
+}
+
+const Logo: React.FC<LogoProps> = ({ name, size = 24 }) => {
+  const icon = LogoNameMap[name]
+
+  return (
+    <svg width={size} height={size} viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path d={icon.path} fill={`#${icon.hex}`} />
```
abhinavrastogi-harness, line 15: we can add currentColor here to make it respect the theme

vardanbansal-harness: https://github.com/user-attachments/assets/0343759c-e8f1-4c4b-bdd5-621d330c1629

Yup, that seems to be working, done.

packages/ui/src/components/logo/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './logo'
+export * from './logo-name-map'
```
knagurski, line 2: Do we need `LogoNameMap` in the publicly exposed API of the library?
```suggestion
```

I'd consider removing this `index` file and just changing the import in `src/components/index.ts` to instead be `export * from './logo/logo'`, though it's fine to keep the `index.ts` file as well if you prefer the optics.

packages/ui/src/components/logo/logo-name-map.ts
```diff
@@ -0,0 +1,9 @@
+import { siBitbucket, siGithub, siGitlab, siJira, siKubernetes, type SimpleIcon } from 'simple-icons'
+
+export const LogoNameMap = {
+  github: siGithub,
+  gitlab: siGitlab,
+  bitbucket: siBitbucket,
+  jira: siJira,
+  kubernetes: siKubernetes
+} satisfies Record<string, SimpleIcon>
```
knagurski, line 9: Not a problem, but just curious why you opted for `satisfies` over typing the variable? Don't feel the need to change it, I'm just wondering that's all.
```typescript tsx
export const LogoNameMap: Record<string, SimpleIcon> = {
  github: siGithub,
  gitlab: siGitlab,
  bitbucket: siBitbucket,
  jira: siJira,
  kubernetes: siKubernetes
}
```

packages/ui/src/components/logo/logo.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
```
knagurski, line 1: This triggers an ESLint warning since we don't have synthetic default exports turned on and React doesn't have a default export.

Instead, can we just import `FC`?
```suggestion
import { FC } from 'react'
```

packages/ui/src/components/logo/logo.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { LogoNameMap } from './logo-name-map'
+
+interface LogoProps {
+  name: keyof typeof LogoNameMap
+  size?: number
+  original?: boolean // Uses brand color when true
+}
+
+const Logo: React.FC<LogoProps> = ({ name, size = 24, original = false }) => {
+  const icon = LogoNameMap[name]
```
knagurski, line 12: What happens if `name` isn't in the map? Might be a good idea to check for that and render a blank SVG or return a fragment?

packages/ui/src/components/logo/logo.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  return (
+    <svg width={size} height={size} viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path d={icon.path} fill={original ? `#${icon.hex}` : 'currentColor'} />
```
knagurski, line 16: Interesting that they were able to get all logos to work on a single-path/single-colour basis and only with fills and not strokes 🤔

packages/ui/src/components/logo/logo-name-map.ts
vardanbansal-harness: I agree using the type itself is better here. I've cleaned it up.

packages/ui/src/components/logo/index.ts
vardanbansal-harness: Done as suggested.

packages/ui/src/components/logo/logo.tsx
vardanbansal-harness: Done, returning a `<></>` if name is not in map.

apps/design-system/src/subjects/components/componentPages.ts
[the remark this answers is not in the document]
vardanbansal-harness: Done, moved the docs to `apps/portal` instead.

apps/portal/src/content/docs/foundations/logos.mdx
```diff
@@ -0,0 +1,87 @@
...
+title: Logos
+description: Logo component
+---
+
+The `Logo` component renders brand icons from the `simple-icons` package.
+
+import { DocsPage } from "@/components/docs-page";
+import { Logo, LogoNameMap } from "@harnessio/ui/components";
+
+<DocsPage.ComponentExample
```
abhinavrastogi-harness, line 11: i think this format of documentation works better for components.
for the foundations, let's render a grid of all available logos. we don't need to show code samples for it.

abhinavrastogi-harness (approved):

apps/portal/src/content/docs/foundations/logos.mdx
vardanbansal-harness: Sounds good, will take it up in a follow up PR.

apps/portal/src/content/docs/foundations/logos.mdx
knagurski, line 1: This should be in the components area, not the foundations

apps/portal/src/content/docs/foundations/logos.mdx
```diff
@@ -0,0 +1,87 @@
+---
+title: Logos
```
knagurski, line 2: The title should match the component name

apps/portal/src/content/docs/foundations/logos.mdx
knagurski, line 1: Also, the name of the file should match the component name

apps/portal/src/content/docs/foundations/logos.mdx
```diff
@@ -0,0 +1,87 @@
...
+description: Logo component
+---
+
+The `Logo` component renders brand icons from the `simple-icons` package.
+
+import { DocsPage } from "@/components/docs-page";
+import { Logo, LogoNameMap } from "@harnessio/ui/components";
```
knagurski, line 9: No need to import these since they're not used in this page

apps/portal/src/content/docs/foundations/logos.mdx
```diff
@@ -0,0 +1,87 @@
...
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    </div>
+  </>`}
+/>
+
+<DocsPage.ComponentExample
```
knagurski, line 40: Only one component example block is needed. If there are multiple examples, break them up in a similar way as other components

apps/portal/src/content/docs/foundations/logos.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+### Props
+
+| Name                                                          | Type                                                                          | Default | Description                                                                 |
+| ------------------------------------------------------------- | ----------------------------------------------------------------------------- | ------- | --------------------------------------------------------------------------- |
+| <p style={{ whiteSpace: 'nowrap' }}><code>original</code></p> | `boolean`                                                                     | `false` | If `true`, the logo will use its brand-specific color. Defaults to `false`. |
+| `name`                                                        | <p style={{ whiteSpace: 'nowrap' }}><code>keyof typeof LogoNameMap</code></p> | N/A     | The name of the brand icon to render (must match a key in `LogoNameMap`).   |
+| `size`                                                        | `number`                                                                      | `24`    | The size (in pixels) of the logo.                                           |
```
knagurski, line 87: Use the same `PropsTable` component as other components

knagurski (commented):
We'll need to move and alter the docs to bring it inline with the rest of the components, but that can follow. This doesn't qualify as a foundation piece and we don't yet have designs for what foundation pieces would look like anyway, but I'd say we'll treat them more as topic explorations rather than component docs; `Logo` is a component so should be with the other components.

PR #1337, opened 2025-03-28
fix: fix summary line

cjlee01:
fix summary line
-Can we make the line under "Summary" to be 2px instead of 1px and use a dark black color?
image


sans-harness (approved):
Can we add a picture?

cjlee01:
> Can we add a picture?

sure @sans-harness

PR #1366, opened 2025-04-01
feat: [pipe-26410]: connectors edit flow

shaurya-harness:


apps/design-system/src/utils/connectors/aws-kms-connector.tsx
```diff
@@ -1,8 +1,18 @@
 import { IFormDefinition } from '@harnessio/forms'
-import { CredTypeValues, DelegateTypes, IInputConfigWithConfigInterface, InputConfigType } from '@harnessio/ui/views'
+import { IInputConfigWithConfigInterface, InputConfigType } from '@harnessio/ui/views'
 
 export const AWS_KMS_CONNECTOR_CATEGORY = 'Secrets Manager'
 
+export enum CredTypeValues {
```
vardanbansal-harness, line 6: is this enum relevant only to AWS KMS connector type?

packages/ui/src/views/connectors/connector-entity-form.tsx
```diff
@@ -100,7 +125,8 @@ export const ConnectorEntityForm = (props: ConnectorEntityFormProps): JSX.Elemen
           <EntityFormSectionLayout.Root>
             <EntityFormSectionLayout.Header>
               <EntityFormSectionLayout.Title className="!my-0">
-                Connect to {formEntity.data.name}
+                {intent === ConnectorEntityIntent.CREATE ? 'Connect to ' : 'Edit '}
```
vardanbansal-harness, line 128: Why extra space in `'Connect to '`?

packages/ui/src/views/connectors/types.ts
```diff
@@ -142,3 +133,8 @@ export enum ConnectorSelectionType {
   NEW = 'new',
   EXISTING = 'existing'
 }
+
+export enum ConnectorEntityIntent {
```
vardanbansal-harness, line 137: Isn't this something Secrets would also use? In general, any entity might be created or edited. If yes, then lets name it something less specific to connectors.

cjlee01: it would be changed to EntityIntent

packages/ui/src/views/connectors/connector-entity-form.tsx
cjlee01: removed

vardanbansal-harness (approved):

apps/design-system/src/utils/connectors/aws-kms-connector.tsx
cjlee01: change name to relate only to aws

PR #1371, opened 2025-04-01
feat: Adding Secret list page

vardanbansal-harness:
Preview: https://deploy-preview-1371--harness-xd-review.netlify.app/view-preview/secrets-list-page


packages/ui/src/views/secrets/secrets-list/secrets-list-page.tsx
```diff
@@ -0,0 +1,86 @@
...
+}) => {
+  const { t } = useTranslationStore()
+  const { navigate } = useRouterContext()
+
+  const { search: searchInput, handleSearchChange: handleInputChange } = useDebounceSearch({
+    handleChangeSearchValue: (val: string) => setSearchQuery(val.length ? val : undefined),
+    searchValue: searchQuery || ''
+  })
+
+  if (isError) {
```
abhinavrastogi-harness, line 33: since loading state is handled in the list view.. maybe the error state should also be in the same place

abhinavrastogi-harness (approved):

packages/ui/src/views/secrets/secrets-list/secrets-list-page.tsx
vardanbansal-harness: Sure, will clean up this as discussed in a follow up PR.

PR #1374, opened 2025-04-02
Integrate `Logo` component in Connectors List

vardanbansal-harness:
Preview: https://deploy-preview-1374--harness-xd-review.netlify.app/view-preview/connectors-list-page

- Makes `original` prop `true` by default, to render logos with original brand colors.
- Clean up and improved documentation for `<Logo />` component.


packages/ui/src/views/connectors/connectors-list/connectors-list.tsx
```diff
@@ -76,7 +86,13 @@ export function ConnectorsList({
             >
               <Table.Cell className="max-w-80 content-center truncate">
                 <div className="flex items-center gap-2.5">
-                  <Icon name="connectors" size={24} />
+                  <div className="min-w-[40px]">
+                    {connector.type && LogoNameMap[connector.type.toLowerCase()] ? (
+                      <Logo name={connector.type.toLowerCase()} size={32} />
```
abhinavrastogi-harness, line 91: i don't think connector type will always map to a logo name.
we should maintain more control over this by maintaining a map of type to logo.

vardanbansal-harness: Agreed, I have cleaned up this a bit to dynamically generate mapping of logo names to the actual logo imported from `simple-icons` package.

packages/ui/src/views/connectors/utils.tsx
```diff
@@ -0,0 +1,28 @@
...
+    case 'Jira':
+      return 'jira'
+    case 'K8sCluster':
+      return 'kubernetes'
+    default:
+      return undefined
+  }
+}
+
+export const getConnectorLogo = (connectorType: ConnectorConfigType): SimpleIcon | React.ReactNode => {
```
vardanbansal-harness, line 24: Moved this out to a separate util for reusability, this could be used on Create/Edit connector drawer as well.

packages/ui/src/views/connectors/utils.tsx
```diff
@@ -0,0 +1,28 @@
...
+    case 'Github':
+      return 'github'
+    case 'Gitlab':
+      return 'gitlab'
+    case 'Bitbucket':
+      return 'bitbucket'
+    case 'Jira':
+      return 'jira'
+    case 'K8sCluster':
+      return 'kubernetes'
```
vardanbansal-harness, line 18: These logo names `github`, `gitlab`, etc. should ideally come from an enum or a union type but I couldn't figure out how to dynamically generate it based on only the icons we're importing from `simple-icons`. But I will give it a try in a follow up PR.

packages/ui/src/components/logo/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * and the values are `SimpleIcon` objects containing path data and brand color.
+ *
+ * @example
+ * ```ts
+ * const logoMap = getLogoNameMap();
+ * const githubIcon = logoMap.get("github");
+ * console.log(githubIcon?.path); // SVG path data
+ * ```
+ */
+export const getLogoNameMap = (): Map<string, SimpleIcon> =>
```
abhinavrastogi-harness, line 18: ooh.. i think there was some misunderstanding.
this logic should NOT be in the logo component or in the `ui` package. the earlier simple implementation was correct.
instead, the usage of it should be via a map in the view layer.

vardanbansal-harness: Done, removed this logic.

packages/ui/src/views/connectors/connectors-list/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { LogoName } from '@components/logo'
+
+import { ConnectorConfigType } from '../types'
+
+export const ConnectorTypeToLogoNameMap: Map<ConnectorConfigType, LogoName> = new Map([
```
abhinavrastogi-harness, line 5: `Record<ConnectorConfigType, LogoName>` might be a cleaner way to manage this

abhinavrastogi-harness (approved):

packages/ui/src/views/connectors/connectors-list/utils.ts
vardanbansal-harness: Will cleanup in a follow up PR.

PR #1384, opened 2025-04-03
chore: Add MFE build verification to PR checks

vardanbansal-harness:


.github/workflows/build-and-deploy-storybook.yml
```diff
@@ -52,3 +52,17 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: pnpm/action-setup@v4
+      - run: pnpm deps
+      - run: pnpm build
+      - run: |
+          cd apps/gitness
+          pnpm install
+          pnpm run build:webpack
```
knagurski, line 68: Instead of cd'ing to the directory, you can just filter from the top
```suggestion
      - run: pnpm -r -F ./apps/gitness build:webpack
```

knagurski: Or, since this is the only one that does use `build:webpack`, you can simplify more
```suggestion
      - run: pnpm -r build:webpack
```

.github/workflows/build-and-deploy-storybook.yml
```diff
@@ -52,3 +52,17 @@ jobs:
...
       - run: pnpm test
+
+  Build_Microfrontend:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: pnpm/action-setup@v4
+      - run: pnpm deps
```
knagurski, line 63: We should kill off `deps` at some point since it's entirely pointless. `pnpm install` or `pnpm i` from the root installs all packages/apps anyway, no need to recursively do it

knagurski (approved):
Nice and simple 👍🏻

PR #1408, opened 2025-04-07
Add eslint rule to warn old design system classnames

praneshg239:
A `warn` rule has been added to warn the usage of old design system variables.

image


.eslintrc.js
```diff
@@ -0,0 +1,104 @@
...
+    'tailwindcss/no-contradicting-classname': 'error',
+    'import/no-unresolved': 'off',
+    'react/prop-types': 'off',
+    'react/react-in-jsx-scope': 'off',
+
+    /**
+     * Soon it will be converted to "error"
+     * once all the old variants are removed
+     */
+    'no-restricted-syntax': ['warn', ...getRules()],
```
praneshg239, line 76: It will be converted to `error` once all the colors mentioned in `tailwind-design-system.ts` are migrated to new design system.

packages/ui/tailwind-design-system.ts
```diff
@@ -129,11 +129,6 @@ export default {
           DEFAULT: 'var(--cn-text-1)'
         },
         background: {
-          // TODO: remove DEFAULT, cause use old color var
-          DEFAULT: 'hsl(var(--canary-background))',
```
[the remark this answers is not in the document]
praneshg239: Every color entity defined in `tailwind-design-system.ts` (`primary`, `background`, `foreground`, `borders` etc,.) has to be checked one by one of all its variants like `1` , `2`, `DEFAULT`, `success` etc,. and have to be removed manually. Once a variant is removed, it should be added to the eslint class list to prevent from using. 

@vivek-harness is working on it currently.

vivek-harness (approved):

.eslintrc.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * Rules to adopt new design system
+ *
+ * class list will be updated once the
+ * relevant class is completed removed from
+ * tailwind-design-system.ts
+ *
+ */
+const getRules = () => {
```
abhinavrastogi-harness, line 9: can we move this to a separate file? we might have more custom rules in the future.
having separate plugins for each purpose makes more sense to me.

apps/gitness/package.json
```diff
@@ -24,7 +24,6 @@
     "@harnessio/code-service-client": "3.5.0",
     "@harnessio/forms": "workspace:*",
     "@harnessio/pipeline-graph": "workspace:*",
-    "@harnessio/core-design-system": "workspace:*",
```
praneshg239, line 27: Removing `@harnessio/core-design-system` from `gitness`.

PR #1409, opened 2025-04-07
Color changes for the new design system

vivek-harness:
This PR replaces the colors from the current design system with the new design system.


knagurski:
Has all this been tested? I see changes to a package that doesn't use either the UI lib or the core-design-stystem package directly.

srdjan-harness (approved):

PR #1413, opened 2025-04-07
fix: Add App Router Provider to avoid re-routing issue

vardanbansal-harness:
- Moves `RouterContextProvider` inside `RouterProvider` to allow calling `useNavigate` hooks and pass down the correct implementation of `navigate` function to be used via `useRouterContext`.
- This fixes the unexpected re-routing issue seen while visiting Pull Requests list page.

**Without fix:**

https://github.com/user-attachments/assets/74d1266a-37dc-4af3-99eb-740b651a0129

**With fix:**

https://github.com/user-attachments/assets/b959600e-814f-4305-849a-463eaf1765f4

### DIFF
--- a/apps/gitness/src/App.tsx
+++ b/apps/gitness/src/App.tsx
@@ -1,19 +1,10 @@
 import { I18nextProvider } from 'react-i18next'
-import {
-  createBrowserRouter,
-  Link,
-  NavLink,
-  Outlet,
-  RouterProvider,
-  useMatches,
-  useSearchParams
-} from 'react-router-dom'
+import { createBrowserRouter, RouterProvider } from 'react-router-dom'
 
 import { QueryClientProvider } from '@tanstack/react-query'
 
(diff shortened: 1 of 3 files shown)


apps/gitness/src/App.tsx
```diff
@@ -48,17 +39,7 @@ export default function App() {
             <Tooltip.Provider>
               <ExitConfirmProvider>
                 <NavigationProvider routes={routes}>
-                  <RouterContextProvider
-                    Link={Link}
-                    NavLink={NavLink}
-                    Outlet={Outlet}
-                    location={{ ...window.location, state: {}, key: '' }}
-                    navigate={router.navigate}
```
vardanbansal-harness, line 56: This `router.navigate` was seemingly causing unexpected re-routing. It has been replaced with `navigate` function from `useNavigate`.

abhinavrastogi-harness (approved):
there's too many layers of routers, contexts and providers. let's discuss on how we can simplify this. not a blocker though.

PR #1416, opened 2025-04-07
Add filters to connectors page

sans-harness:
https://github.com/user-attachments/assets/05b06f87-096b-47ed-8253-6d9049188b97


abhinavrastogi-harness:
generally I'm not happy with the filters spec/design right now. But let's follow up on that separately with design/product.

abhinavrastogi-harness (approved):

PR #1417, opened 2025-04-08
fix: added router issue for both mfe and design-system

harish-viswa:
Fixed the same issue mentioned here #1413 for mfe and design-system
Updated `window.location` to `useLocation`, 
since mfe would break if window.location is used, as it will contain path for both parent app and child mfe

https://github.com/user-attachments/assets/85c78e20-bda7-406b-b46f-b66348fc6b44

https://github.com/user-attachments/assets/e767f97d-0d77-4163-b80a-acda6891eb3c

### DIFF
--- a/apps/gitness/src/routes.tsx
+++ b/apps/gitness/src/routes.tsx
@@ -1133,10 +1133,12 @@ export const mfeRoutes = (mfeProjectId = '', mfeRouteRenderer: JSX.Element | nul
   {
     path: '/',
     element: (
-      <AppProvider>
-        {mfeRouteRenderer}
-        <AppShellMFE />
-      </AppProvider>
+      <AppRouterProvider>
+        <AppProvider>
+          {mfeRouteRenderer}
+          <AppShellMFE />
+        </AppProvider>
+      </AppRouterProvider>
     ),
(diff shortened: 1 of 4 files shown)


apps/gitness/src/routes.tsx
```diff
@@ -1133,10 +1133,12 @@ export const mfeRoutes = (mfeProjectId = '', mfeRouteRenderer: JSX.Element | nul
   {
     path: '/',
     element: (
-      <AppProvider>
-        {mfeRouteRenderer}
-        <AppShellMFE />
-      </AppProvider>
+      <AppRouterProvider>
```
knagurski, line 1136: I thought this had to be done on the Harness Core UI side of things since the values need to come from the version of React Router over there 🤔

harish-viswa: Hey @knagurski, the router values will be passed only for the view level components imported from UI pkg.
Since mfe has it's own router context, passing router values from parent will cause links and navigation to break, as it will be created based on the parent app context

I tried to see what would happen, but you can see at the bottom how the link has been created for child app
image

vardanbansal-harness (approved):

PR #1427, opened 2025-04-09
feat: [pipe-26514]: fix styling and test connection dialog

cjlee01:
fix styling and test connection dialog


packages/ui/locales/en/views.json
```diff
@@ -280,7 +283,11 @@
     "activityHistory": "Activity history",
     "createNew": "Create new connector",
     "updated": "Last updated",
-    "lastConnected": "Last connected"
+    "lastConnected": "Last connected",
+    "viewConnectorDetails": "View connector details",
+    "connector": "Connector:",
+    "status": "Status:",
```
vardanbansal-harness, line 289: Avoid `:` in translations, otherwise this cannot be used where we just need to have `Connector` or `Status`.

packages/ui/src/views/execution/execution-status.tsx
```diff
@@ -4,15 +4,15 @@ import { ExecutionState } from '@views/repo/pull-request'
...
+  const { status, duration, minimal, inConnector = false } = props
   switch (status.toLowerCase()) {
     case ExecutionState.WAITING_ON_DEPENDENCIES:
     case ExecutionState.BLOCKED:
     case ExecutionState.PENDING:
       return minimal ? (
         <div className="flex items-center gap-1">
           <div className="size-2 rounded-full bg-cn-background-softgray" />
-          <span className="text-cn-foreground-disabled">Pending</span>
+          <span className={inConnector ? 'text-cn-foreground-1' : 'text-cn-foreground-disabled'}>Pending</span>
```
vardanbansal-harness, line 15: I feel having an entity specific logic (entity being a connector) in a common component like this would scale, without having complex `if-else`s.

cjlee01: so i shud make a seperate status component for this ? @vardanbansal-harness

vardanbansal-harness: Discussing offline

packages/ui/locales/en/views.json
cjlee01: fixed

vardanbansal-harness (approved):

PR #1449, opened 2025-04-10
Brand icons component

hiderr:


knagurski:
I think we can drop this PR. We already have the `Logo` component which does this.

PR #1470, opened 2025-04-14
add timestamp popup design review

ankormoreankor:
##### Timestamp popup design review

- adjusted the timestamp popup to match the new design
- added an arrow-pointer to the `HoverCard` component

Before/after:
image

### DIFF
--- a/packages/ui/src/components/hover-card.tsx
+++ b/packages/ui/src/components/hover-card.tsx
@@ -7,6 +7,19 @@ const HoverCardRoot = HoverCardPrimitive.Root
 
 const HoverCardTrigger = HoverCardPrimitive.Trigger
 
+const HoverCardArrow = React.forwardRef<
+  React.ElementRef<typeof HoverCardPrimitive.Arrow>,
+  React.HTMLAttributes<HTMLDivElement>
+>(({ className, ...props }, ref) => (
+  <HoverCardPrimitive.Arrow ref={ref} asChild>
+    <div
+      className={cn('bg-cn-background-2 size-1.5 border-t border-l rotate-[225deg] -m-[55%] rounded-[1px]', className)}
+      {...props}
+    />
+  </HoverCardPrimitive.Arrow>
+))
+HoverCardArrow.displayName = HoverCardPrimitive.Arrow.displayName
+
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/hover-card.tsx
```diff
@@ -27,7 +40,8 @@ HoverCardContent.displayName = HoverCardPrimitive.Content.displayName
 const HoverCard = {
   Root: HoverCardRoot,
   Trigger: HoverCardTrigger,
-  Content: HoverCardContent
+  Content: HoverCardContent,
+  Arrow: HoverCardArrow
```
knagurski, line 44: I wonder if we have a use-case where we _don't_ want the arrow? Unless there is that I don't know about, for consistency, I'd bake the arrow into the `Content` component and not export it here at all.

ankormoreankor: Currently, the component is used once, and we don't actually know if we need the arrow as a separate component, so it makes sense to do what you suggested 👍

packages/ui/src/components/time-ago-hover-card.tsx
```diff
@@ -54,18 +54,27 @@ export const TimeAgoHoverCard: FC<TimeAgoHoverCardProps> = ({ formattedDate, tim
...
+          const date = zone === 'UTC' ? formattedDates.utcDate : formattedDates.localDate
+          const time = zone === 'UTC' ? formattedDates.utcTime : formattedDates.localTime
+
+          return (
+            <>
+              <Badge variant="soft" size="sm">
+                {zone === 'UTC' ? 'UTC' : getTimeZoneAbbreviation()}
+              </Badge>
+              <time dateTime={date}>{date}</time>
+              <time dateTime={time} className="text-cn-foreground-3 ml-auto">
```
knagurski, line 72: Nice use of the `time` tag. I don't think the `date` and `time` we're passing in will be in the correct format (see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#valid_datetime_values), but we can follow up on that in another PR.

knagurski (approved):
Looks good. Very handy to share that component beyond gitness 👍🏻

PR #1472, opened 2025-04-14
changes in branch selector

hiderr:
# Branch Creation Encapsulation PR

This PR introduces a new component that encapsulates all data-fetching logic for the CreateBranchDialog UI component, following the common approach described in the [wiki](https://github.com/harness/canary/wiki/State-Management-Patterns) 

Key changes include:

1. Added a new reusable `CreateBranchDialog` component in `apps/gitness/src/components-v2/create-branch-dialog.tsx` that provides a consistent interface for branch creation throughout the application.

2. Improved the branch selector component by adding a `preSelectedTab` prop that allows pre-selecting either branches or tags tab when opening the selector. We need it to preselect Tag tab, when user selects Create Branch menu from a specific tag on Tags page

3. Integrated the toast notification system for branch creation with a new `useToastNotification` hook in `packages/ui/src/components/toast/use-toast.ts` to provide consistent success messages.

4. Refactored the branch creation flow in multiple places to use the new centralized dialog component, reducing code duplication.

5. Added and improved translations for branch-related operations in both English and French locales.

6. Updated the UI styling to maintain consistency with the application's design system.

### DIFF
--- a/apps/design-system/src/subjects/views/repo-branches/index.tsx
+++ b/apps/design-system/src/subjects/views/repo-branches/index.tsx
@@ -14,7 +14,7 @@ export function RepoBranchesView() {
     <RepoBranchListView
       isLoading={false}
       isCreatingBranch={false}
-      onSubmit={noop}
+      onSubmit={async () => {}}
       useRepoBranchesStore={useRepoBranchesStore}
       useTranslationStore={useTranslationStore}
       isCreateBranchDialogOpen={isCreateBranchDialogOpen}
--- a/apps/gitness/src/components-v2/create-branch-dialog.tsx
+++ b/apps/gitness/src/components-v2/create-branch-dialog.tsx
@@ -1,0 +1,92 @@
+import { useCallback, useEffect, useState } from 'react'
+
+import { useCreateBranchMutation } from '@harnessio/code-service-client'
+import {
+  BranchSelectorListItem,
+  BranchSelectorTab,
+  CreateBranchDialog as CreateBranchDialogComp,
+  CreateBranchFormFields
+} from '@harnessio/ui/views'
+
+import { useGetRepoRef } from '../framework/hooks/useGetRepoPath'
+import { useTranslationStore } from '../i18n/stores/i18n-store'
+import { BranchSelectorContainer } from './branch-selector-container'
+
+interface CreateBranchDialogProps {
+  open: boolean
+  onClose: () => void
+  onSuccess?: () => void
+  preselectedBranchOrTag?: BranchSelectorListItem | null
+  preselectedTab?: BranchSelectorTab
+}
+
+export const CreateBranchDialog = ({
+  open,
+  onClose,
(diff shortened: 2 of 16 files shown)


apps/design-system/src/subjects/views/repo-branches/index.tsx
```diff
@@ -14,7 +14,7 @@ export function RepoBranchesView() {
     <RepoBranchListView
       isLoading={false}
       isCreatingBranch={false}
-      onSubmit={noop}
+      onSubmit={async () => {}}
```
sans-harness, line 17: why do we need this instead of noop ?

apps/gitness/src/components-v2/create-branch-dialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const selectBranchOrTag = useCallback((branchTagName: BranchSelectorListItem) => {
+    setSelectedBranchOrTag(branchTagName)
+  }, [])
+
+  const { showToast } = useToastNotification({
+    title: t('views:repos.branchCreated'),
+    description: t('views:repos.branchCreatedDescription', { name: createdBranchName })
+  })
+
```
sans-harness, line 46: Unsure if we should be using toasts for branch creation ? 
cc: @abhinavrastogi-harness

apps/gitness/src/components-v2/create-branch-dialog.tsx
sans-harness, line 1: Is there a case for making this an `encapsulated-component` ? Are we using this component with the same business logic across pages ?

apps/design-system/src/subjects/views/repo-branches/index.tsx
hiderr: because of type change. this function now returns promise

apps/gitness/src/components-v2/create-branch-dialog.tsx
hiderr: The designers suggested adding toast, since the current branch creation flow will also be used on the tags list, where there is no visual indication that the branch has been created — the user stays on the tags page. It’s important for the user to understand that the action was completed successfully.

apps/gitness/src/components-v2/create-branch-dialog.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const repo_ref = useGetRepoRef()
+
+  const [selectedBranchOrTag, setSelectedBranchOrTag] = useState<BranchSelectorListItem | null>(null)
+
+  const selectBranchOrTag = useCallback((branchTagName: BranchSelectorListItem) => {
+    setSelectedBranchOrTag(branchTagName)
+  }, [])
+
+  const {
+    mutateAsync: createBranch,
```
sans-harness, line 39: NIT: why not just use mutate here?

sans-harness (approved):

PR #1473, opened 2025-04-14
fix: fix suggestion type

cjlee01:
fix suggestion type on pr changes

### DIFF
--- a/packages/ui/src/views/repo/pull-request/components/pull-request-diff-viewer.tsx
+++ b/packages/ui/src/views/repo/pull-request/components/pull-request-diff-viewer.tsx
@@ -570,6 +570,7 @@ const PullRequestDiffViewer = ({
                                   />
                                 ) : (
                                   <PRCommentView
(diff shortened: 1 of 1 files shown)


packages/ui/src/views/repo/pull-request/components/pull-request-diff-viewer.tsx
```diff
@@ -570,6 +570,7 @@ const PullRequestDiffViewer = ({
                                   />
                                 ) : (
                                   <PRCommentView
+                                    parentItem={parent?.payload as CommentItem<TypesPullReqActivity>}
```
sans-harness, line 573: Do we need to explicitly typecast this?

sans-harness (approved):

packages/ui/src/views/repo/pull-request/components/pull-request-diff-viewer.tsx
cjlee01: yeah

PR #1476, opened 2025-04-15
make inset layout default app layout

ankormoreankor:
This PR makes the inset theme used by default and removes an option to choose it.

### DIFF
--- a/apps/design-system/src/App.tsx
+++ b/apps/design-system/src/App.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect, useState } from 'react'
+import { FC } from 'react'
 import { createBrowserRouter, Navigate, RouterProvider } from 'react-router-dom'
 
 import ComponentPage from '@/pages/component-page'
@@ -22,26 +22,10 @@ const router = createBrowserRouter([
 ])
 
 const App: FC = () => {
-  const [isInset, setIsInset] = useState<boolean>(false)
   const themeStore = useThemeStore()
 
-  /**
-   * Set inset on mount and listen for changes in ViewSettings
-   */
-  useEffect(() => {
-    const setInset = () => {
-      const inset = sessionStorage.getItem('view-preview-is-inset')
-      setIsInset(inset === 'true')
-    }
-
-    setInset()
-    window.addEventListener('storageChange', setInset)
-
-    return () => window.removeEventListener('storageChange', setInset)
-  }, [])
-
   return (
-    <ThemeProvider {...{ ...themeStore, isInset }}>
+    <ThemeProvider {...{ ...themeStore }}>
       <RouterProvider router={router} />
     </ThemeProvider>
   )
--- a/apps/design-system/src/App.tsx
+++ b/apps/design-system/src/App.tsx
@@ -25,7 +25,7 @@ const App: FC = () => {
   const themeStore = useThemeStore()
 
   return (
-    <ThemeProvider {...{ ...themeStore }}>
+    <ThemeProvider {...themeStore}>
       <RouterProvider router={router} />
     </ThemeProvider>
   )
--- a/apps/gitness/src/framework/context/ThemeContext.tsx
+++ b/apps/gitness/src/framework/context/ThemeContext.tsx
@@ -73,5 +73,9 @@ export function ThemeProvider({ children, defaultTheme }: ThemeProviderProps) {
     }
   }, [theme, setTheme, systemMode])
 
-  return <UIThemeProvider {...{ theme, setTheme, isLightTheme }}>{children}</UIThemeProvider>
+  return (
+    <UIThemeProvider theme={theme} setTheme={setTheme} isLightTheme={isLightTheme}>
+      {children}
+    </UIThemeProvider>
+  )
 }
--- a/apps/gitness/src/framework/context/ThemeContext.tsx
+++ b/apps/gitness/src/framework/context/ThemeContext.tsx
@@ -12,9 +12,7 @@ export const useThemeStore = create<IThemeStore>()(
   persist(
     set => ({
       theme: undefined,
-      isInset: false,
       setTheme: (newTheme: FullTheme) => set({ theme: newTheme }),
-      setInset: (isInset: boolean) => set({ isInset }),
       isLightTheme: false
     }),
     {
@@ -28,7 +26,7 @@ interface ThemeProviderProps {
   defaultTheme: FullTheme
 }
 export function ThemeProvider({ children, defaultTheme }: ThemeProviderProps) {
-  const { theme, setTheme, setInset, isLightTheme, isInset } = useThemeStore()
+  const { theme, setTheme, isLightTheme } = useThemeStore()
   const isMFE = useIsMFE()
 
   const mediaQuery = window.matchMedia('(prefers-color-scheme: dark)')
@@ -61,10 +59,6 @@ export function ThemeProvider({ children, defaultTheme }: ThemeProviderProps) {
 
     if (!isMFE) {
       root.classList.add(effectiveTheme) // Apply the computed theme class
-
-      if (isInset) {
-        root.classList.add('inset-layout')
-      }
     } else {
       root.classList.add(mode)
     }
@@ -77,7 +71,7 @@ export function ThemeProvider({ children, defaultTheme }: ThemeProviderProps) {
     if (!colorSchemeMeta.parentNode) {
       document.head.appendChild(colorSchemeMeta)
     }
-  }, [theme, setTheme, systemMode, isInset])
+  }, [theme, setTheme, systemMode])
 
-  return <UIThemeProvider {...{ theme, setTheme, setInset, isLightTheme, isInset }}>{children}</UIThemeProvider>
+  return <UIThemeProvider {...{ theme, setTheme, isLightTheme }}>{children}</UIThemeProvider>
 }
--- a/packages/ui/src/components/app-sidebar/sidebar-item.tsx
+++ b/packages/ui/src/components/app-sidebar/sidebar-item.tsx
@@ -1,5 +1,5 @@
 import { DropdownMenu, Icon, IconProps, Sidebar, Text, useSidebar } from '@/components'
-import { useRouterContext, useTheme } from '@/context'
+import { useRouterContext } from '@/context'
 import { cn } from '@/utils'
 import { TFunction } from 'i18next'
 
@@ -34,7 +34,6 @@ export const SidebarItem = ({
 }: NavbarItemProps) => {
   const { NavLink } = useRouterContext()
   const { collapsed } = useSidebar()
-  const { isInset } = useTheme()
 
   const iconName = item.iconName && (item.iconName.replace('-gradient', '') as IconProps['name'])
 
@@ -94,15 +93,12 @@ export const SidebarItem = ({
       {!collapsed && (
         <DropdownMenu.Root>
           <DropdownMenu.Trigger asChild>
-            <Sidebar.MenuAction
-              className={cn('text-sidebar-icon-3 hover:text-sidebar-icon-1', isInset ? 'right-[3px]' : 'right-0')}
-              showOnHover
-            >
+            <Sidebar.MenuAction className={cn('text-sidebar-icon-3 hover:text-sidebar-icon-1 right-[3px]')} showOnHover>
               <Icon name="menu-dots" size={12} />
             </Sidebar.MenuAction>
           </DropdownMenu.Trigger>
           <DropdownMenu.Content
-            className="w-[128px] border-sidebar-border-3 bg-sidebar-background-4"
+            className="border-sidebar-border-3 bg-sidebar-background-4 w-[128px]"
             align="end"
             sideOffset={3}
             alignOffset={4}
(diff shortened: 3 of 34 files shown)


apps/design-system/src/App.tsx
```diff
@@ -22,26 +22,10 @@ const router = createBrowserRouter([
...
-
-    setInset()
-    window.addEventListener('storageChange', setInset)
-
-    return () => window.removeEventListener('storageChange', setInset)
-  }, [])
-
   return (
-    <ThemeProvider {...{ ...themeStore, isInset }}>
+    <ThemeProvider {...{ ...themeStore }}>
```
knagurski, line 28:
```suggestion
    <ThemeProvider {...themeStore}>
```

apps/gitness/src/framework/context/ThemeContext.tsx
```diff
@@ -77,7 +71,7 @@ export function ThemeProvider({ children, defaultTheme }: ThemeProviderProps) {
     if (!colorSchemeMeta.parentNode) {
       document.head.appendChild(colorSchemeMeta)
     }
-  }, [theme, setTheme, systemMode, isInset])
+  }, [theme, setTheme, systemMode])
 
-  return <UIThemeProvider {...{ theme, setTheme, setInset, isLightTheme, isInset }}>{children}</UIThemeProvider>
+  return <UIThemeProvider {...{ theme, setTheme, isLightTheme }}>{children}</UIThemeProvider>
```
knagurski, line 76: I wonder why this way done as a spread instead of just using the props as normal 🤔  
```suggestion
  return <UIThemeProvider theme={theme} setTheme={setTheme} isLightTheme={isLightTheme}>{children}</UIThemeProvider>
```

packages/ui/src/components/app-sidebar/sidebar-item.tsx
```diff
@@ -94,15 +93,12 @@ export const SidebarItem = ({
       {!collapsed && (
         <DropdownMenu.Root>
           <DropdownMenu.Trigger asChild>
-            <Sidebar.MenuAction
-              className={cn('text-sidebar-icon-3 hover:text-sidebar-icon-1', isInset ? 'right-[3px]' : 'right-0')}
-              showOnHover
-            >
+            <Sidebar.MenuAction className={cn('text-sidebar-icon-3 hover:text-sidebar-icon-1 right-[3px]')} showOnHover>
```
knagurski, line 96: Looks like we don't need `cn` here anymore.
```suggestion
            <Sidebar.MenuAction className="text-sidebar-icon-3 hover:text-sidebar-icon-1 right-[3px]" showOnHover>
```

packages/ui/src/components/topbar.tsx
```diff
@@ -18,7 +18,7 @@ const Topbar = {
     return (
       <div
         className={cx(
```
knagurski, line 20: Looks like this should be using `cn` from the utilities, not `cx` directly from CVA.
```suggestion
        className={cn(
```

apps/gitness/src/framework/context/ThemeContext.tsx
ankormoreankor: In this particular case, the names of the props and values are the same, so the "spread way" just reduces duplications

packages/ui/src/components/topbar.tsx
```diff
@@ -18,7 +18,7 @@ const Topbar = {
...
-          `grid w-full ${gridCols} font-regular h-[var(--cn-breadcrumbs-height)] topbar-bottom-border items-center gap-6 px-5 text-sm`,
+          `grid w-full ${gridCols} font-regular h-[var(--cn-breadcrumbs-height)] items-center gap-6 px-5 text-sm`,
```
knagurski, line 21: I think we can kill off this template string
```suggestion
          'grid w-full font-regular h-[var(--cn-breadcrumbs-height)] items-center gap-6 px-5 text-sm',
          hasCenter ? 'grid-cols-[auto_1fr_auto]' : 'grid-cols-[1fr_auto]',
```

packages/ui/src/shared-style-variables.css
knagurski, line 1: Do we need these changes? Are they related to making the inset the standard behaviour? I thought we were moving away from the `--canary` prefixed tokens? 🤔

packages/ui/src/views/layouts/PullRequestLayout.tsx
```diff
@@ -66,8 +65,7 @@ export const PullRequestLayout: FC<PullRequestLayoutProps> = ({
         <Tabs.Root variant="tabnav" className="mb-7" defaultValue={PullRequestTabsKeys.CONVERSATION}>
           <Tabs.List
             className={cn(
```
knagurski, line 67: Looks like `cn` isn't needed here anymore

packages/ui/src/views/layouts/main-content-layout.tsx
```diff
@@ -25,40 +24,35 @@ type MainContentLayoutProps = PropsWithChildren<{
...
   const isMobile = sidebarData?.isMobile
 
   return (
     <div
       className={cn(
-        'min-h-screen bg-cn-background-1',
+        'flex flex-col min-h-[calc(100vh-var(--cn-inset-layout-indent)*2)] my-[var(--cn-inset-layout-indent)] mr-[var(--cn-inset-layout-indent)] border rounded-md min-w-fit bg-cn-background-1',
         {
           'ml-[var(--cn-inset-layout-indent)]': isMobile,
           'min-h-[calc(100vh-var(--cn-breadcrumbs-height))]': withBreadcrumbs,
```
knagurski, line 37: Looks like we can kill off this line now since it's overridden in the next set of classes with the same condition.
```suggestion
```

packages/ui/src/shared-style-variables.css
ankormoreankor: Yes, we do, but there are a lot of component tokens that are still relevant

packages/ui/src/views/repo/repo-tags/repo-tags-list-page.tsx
```diff
@@ -104,7 +104,7 @@ export const RepoTagsListView: FC<RepoTagsListViewProps> = ({
         <Spacer size={2} />
         {(isLoading || !!tagsList.length || isDirtyList) && (
           <>
-            <span className="text-24 font-medium text-cn-foreground-1">{t('views:repos.tags', 'Tags')}</span>
+            <span className="text-24 text-cn-foreground-1 font-medium">{t('views:repos.tags', 'Tags')}</span>
```
knagurski, line 107: This looks like some kind of heading, but a `span` was used. Can we change this to something more appropriate?

apps/gitness/src/framework/context/ThemeContext.tsx
knagurski: That's true, but spreads in props are _usually_ used to take a larger thing that's been passed, not as a shorthand, so it's less obvious at a glance. I prefer more code that's super obvious, rather than a few chars less that makes you think at all.

packages/ui/src/views/layouts/main-content-layout.tsx
ankormoreankor: Make sense

apps/design-system/src/App.tsx
ankormoreankor: Fixed

apps/gitness/src/framework/context/ThemeContext.tsx
ankormoreankor: Fixed

packages/ui/src/components/app-sidebar/sidebar-item.tsx
ankormoreankor: Fixed

packages/ui/src/components/topbar.tsx
ankormoreankor: Fixed

packages/ui/src/components/topbar.tsx
ankormoreankor: FIxed

packages/ui/src/styles.css
```diff
@@ -518,35 +518,7 @@ mark {
...
-
-  .inset-layout & {
-    @apply bg-sidebar-background-1;
-  }
-}
-
-.main-content-height {
-  @apply h-[calc(100%-var(--cn-breadcrumbs-height))];
+  /* where 2px is the Inset layout borders */
+  @apply h-[calc(100vh-var(--cn-page-nav-height)-var(--cn-inset-layout-indent)-2px)];
```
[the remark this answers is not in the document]
ankormoreankor: Fixed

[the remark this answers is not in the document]
ankormoreankor: Added a new variable

packages/ui/src/views/repo/repo-branch/components/branch-list.tsx
```diff
@@ -131,19 +132,19 @@ export const BranchesList: FC<BranchListPageProps> = ({
                   <div className="flex items-center gap-2">
                     <Avatar.Root size="4.5">
                       {!!branch?.user?.avatarUrl && <Avatar.Image src={branch?.user?.avatarUrl} />}
-                      <Avatar.Fallback className="text-center text-10">
+                      <Avatar.Fallback className="text-10 text-center">
                         {getInitials(branch?.user?.name ?? '')}
                       </Avatar.Fallback>
                     </Avatar.Root>
-                    <span className="truncate text-cn-foreground-1">{branch?.timestamp}</span>
+                    <span className="text-cn-foreground-1 truncate">{branch?.timestamp}</span>
```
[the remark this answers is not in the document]
ankormoreankor: Replaced with `<time />`

packages/ui/src/views/repo/repo-tags/repo-tags-list-page.tsx
ankormoreankor: Replaced with `<h1 />`

packages/ui/src/views/layouts/SandboxLayout.tsx
```diff
@@ -79,13 +83,12 @@ interface ContentProps {
   children: ReactNode
   maxWidth?: 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl' | '5xl' | '6xl' | '7xl'
   className?: string
-  paddingClassName?: string
 }
 
-function Content({ children, maxWidth, className, paddingClassName = 'px-5 pt-7 pb-11' }: ContentProps) {
-  const widthClass = maxWidth ? `max-w-${maxWidth} mx-auto` : ''
+function Content({ children, maxWidth, className }: ContentProps) {
+  const widthClass = !!maxWidth && `max-w-${maxWidth} mx-auto`
```
knagurski, line 89: `max-w-${maxWidth}` would be problematic. The tailwind utilities don't understand concatenation since it's at runtime and class-pruning is at build time. These classes will only exist if they are referenced in full somewhere the "safe" option would be to use the `maxWidth` as a look up to a map that brings back the full classname.

e.g.

```typescript
type maxWidthOptions = 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl' | '5xl' | '6xl' | '7xl'

const maxWidthMap: Record<maxWithOptions, string> = {
  sm: 'max-w-sm',
  md: 'max-w-md',
  // ...
  '7xl': 'max-w-7xl'
}

interface ContentProps {
  children: ReactNode
  maxWidth?: maxWidthOptions
  className?: string
}

// ...

`${maxWidthMap[maxWidth]} mx-auto`

File: packages/ui/src/views/layouts/SandboxLayout.tsx [hunk 12]
```diff
@@ -79,13 +83,12 @@ interface ContentProps {
...
 
-  return <div className={cn(paddingClassName, widthClass, className)}>{children}</div>
+  return <div className={cn('px-5 pt-7 pb-11 flex flex-col grow w-full', widthClass, className)}>{children}</div>
```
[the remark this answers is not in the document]
ankormoreankor: Fixed

packages/ui/src/views/layouts/SandboxLayout.tsx
ankormoreankor: This way it works well - I tested in the app

image

ankormoreankor: We also use concatenation in the `LabelMarker` component, for instance

ankormoreankor: By the way, your approach will also fail because classnames from `maxWidthMap` won't be added to the CSS bundle since they aren't used within the `className` prop. To fix this, you still need to add them to Tailwind's `safelist`

knagurski: Tailwind's use of CSS Prune used to be very open and just scanned for the text in any known file, regardless of how it was being used, weird that they'd make a change to specifically target `className` since that's a React-specific thing (all the other frameworks use just `class`) and the tailwind team has been famously framework agnostic 🤔 ¯\\_(ツ)_/¯

knagurski (approved):
I don't see any obvious issues 👍🏻

packages/ui/src/views/layouts/SandboxLayout.tsx
ankormoreankor: I could probably mislead you here, so my statement needs to be double-checked😓
But I had this issue in a different project that is written in Go/Templ: once you take something from the `class` prop, Tailwind loses it. And `safelist` works well for cases like this

PR #1491, opened 2025-04-16
fix: gitness mfe build should build for production

abhinavrastogi-harness:


praneshg239 (approved):

sans-harness (approved):

apps/gitness/package.json
```diff
@@ -17,8 +17,8 @@
     "test:changed": "vitest --run --changed=origin/main",
     "extract": "i18next --config './i18n.config.ts'",
     "typecheck": "tsc -b",
-    "build:webpack": "rm -rf dist && webpack --config webpack.config.cjs",
-    "start:webpack": "webpack serve --config webpack.config.cjs"
+    "build:webpack": "rm -rf dist && webpack --config webpack.config.cjs --mode=production",
```
knagurski, line 20: _Technically_ you don't need this since `production` is the default if nothing is specified, but good to be explicit.

knagurski (approved):
Looks ok to me. We should probably double back and set up base/dev/prod configs, but we can keep it simple for now and expand when we need to. I would say that since the target of this repo is always a webpack build, the standard `build` command should be the prod webpack, not vite... one for another PR

PR #1503, opened 2025-04-18
feat: basic design matching for connector details

abhinavrastogi-harness:
| Before | After|
| --- | --- |
| Screenshot 2025-04-18 at 2 28 46 PM | Screenshot 2025-04-18 at 2 38 38 PM |


packages/ui/src/views/connectors/types.ts
```diff
@@ -18,7 +18,7 @@ export type ConnectorEntity<T = any> = {
   name?: string
   spec: T
   description?: string
-  tags?: string[]
+  tags?: Record<string, string>
```
sans-harness, line 21: We should check and fix prop on NGUI side as well

sans-harness (approved):

PR #1507, opened 2025-04-21
feat: added labels filter under pr-list-page

harish-viswa:
- Created a custom type under filters group
- Added labels-field as part of pr-list-page

image

### DIFF
--- a/apps/gitness/src/pages-v2/pull-request/pull-request-list.tsx
+++ b/apps/gitness/src/pages-v2/pull-request/pull-request-list.tsx
@@ -31,6 +31,7 @@ export default function PullRequestListPage() {
   const [populateLabelStore, setPopulateLabelStore] = useState(false)
   const [searchParams] = useSearchParams()
   const defaultAuthorId = searchParams.get('created_by')
+  const labelBy = searchParams.get('label_by')
   const mfeContext = useMFEContext()
   usePopulateLabelStore({ queryPage, query: labelsQuery, enabled: populateLabelStore, inherited: true })
 
@@ -83,10 +84,10 @@ export default function PullRequestListPage() {
   }, [page, queryPage, setPage])
 
   useEffect(() => {
-    if (searchParams.get('label_by')) {
+    if (labelBy) {
       setPopulateLabelStore(true)
     }
-  }, [searchParams])
+  }, [labelBy])
 
   return (
     <SandboxPullRequestListPage
@@ -115,7 +116,7 @@ export default function PullRequestListPage() {
             if ((key === 'created_gt' || key === 'created_lt') && value instanceof Date) {
               acc[key] = value.getTime().toString()
             }
-            if (key === 'created_by' && 'value' in value) {
+            if (key === 'created_by' && typeof value === 'object' && 'value' in value) {
               acc[key] = value.value
             }
             if (key === 'label_by') {
--- a/apps/gitness/src/pages-v2/pull-request/pull-request-list.tsx
+++ b/apps/gitness/src/pages-v2/pull-request/pull-request-list.tsx
@@ -28,10 +28,11 @@ export default function PullRequestListPage() {
   const [queryPage, setQueryPage] = useQueryState('page', parseAsInteger.withDefault(1))
   const [filterValues, setFilterValues] = useState<ListPullReqQueryQueryParams>({})
   const [principalsSearchQuery, setPrincipalsSearchQuery] = useState<string>()
+  const [populateLabelStore, setPopulateLabelStore] = useState(false)
   const [searchParams] = useSearchParams()
   const defaultAuthorId = searchParams.get('created_by')
   const mfeContext = useMFEContext()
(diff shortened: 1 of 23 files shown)


packages/ui/src/views/repo/pull-request/pull-request-list-page.tsx
```diff
@@ -39,22 +53,100 @@ const PullRequestListPage: FC<PullRequestPageProps> = ({
...
+                color={labelDetails.color}
+                label={labelDetails.key}
+                value={valueDetails?.value || ''}
+              />
+            )}
+
+            {remainingLabelValues > 0 && (
+              <Badge
+                size="sm"
+                className="text-xs"
```
praneshg239, line 132: Shall we remove `text-xs` from `Badge`? We can use the `sm` `counter` Badge as it is. Could you try removing the `primary` theme too?

```tsx
<Badge
  size="sm"
  variant="counter"
```

packages/ui/src/components/checkbox.tsx
```diff
@@ -17,11 +17,20 @@ const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, CheckboxP
     <div className={cn('flex gap-x-2.5', className)}>
       <CheckboxPrimitive.Root
         ref={ref}
-        className="peer flex size-4 shrink-0 items-center justify-center rounded-sm border border-icons-1 hover:border-icons-3 disabled:cursor-not-allowed disabled:border-icons-4 data-[state=checked]:border-icons-2 data-[state=checked]:bg-cn-background-accent data-[state=checked]:text-cn-foreground-primary"
+        className={cn(
+          'peer flex size-4 shrink-0 items-center justify-center rounded-sm border border-icons-1 disabled:cursor-not-allowed disabled:border-icons-4 data-[state=checked]:border-icons-2 data-[state=checked]:bg-cn-background-accent data-[state=checked]:text-cn-foreground-primary',
+          {
+            'bg-cn-background-accent text-cn-foreground-primary border-icons-2': props.checked === 'indeterminate'
```
praneshg239, line 23: Do we need `border-icons-2` on checked / indeterminate?

packages/ui/src/components/checkbox.tsx
```diff
@@ -17,11 +17,20 @@ const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, CheckboxP
...
+          }
+        )}
         {...props}
       >
         <CheckboxPrimitive.Indicator className={cn('flex items-center justify-center text-current mt-[1px]')}>
-          <Icon name="checkbox" width={10} height={7} className="text-icons-5" />
+          {props.checked === 'indeterminate' ? (
+            <Icon name="minus" size={10} className="text-icons-5" />
+          ) : (
+            <Icon name="checkbox" width={10} height={7} className="text-icons-5" />
```
praneshg239, line 32: can we replace `width={10} height={7}` with nearest `size` prop?

apps/gitness/src/pages-v2/pull-request/pull-request-list.tsx
```diff
@@ -75,6 +82,12 @@ export default function PullRequestListPage() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [page, queryPage, setPage])
 
+  useEffect(() => {
+    if (searchParams.get('label_by')) {
+      setPopulateLabelStore(true)
+    }
+  }, [searchParams])
```
vardanbansal-harness, line 89: This `useEffect` may cause unnecessary re-renders even if the actual parameters haven’t changed, because the searchParams object itself might be re-instantiated every render, making it a new reference each time.

Prefer below:

```
const labelBy = searchParams.get('label_by');

useEffect(() => {
  if (labelBy) {
    setPopulateLabelStore(true);
  }
}, [labelBy]);
```

apps/gitness/src/pages-v2/pull-request/pull-request-list.tsx
```diff
@@ -86,22 +99,41 @@ export default function PullRequestListPage() {
...
-          )
+          Object.entries(filterData).reduce<
+            Record<string, ListPullReqQueryQueryParams[keyof ListPullReqQueryQueryParams]>
+          >((acc, [key, value]) => {
+            if ((key === 'created_gt' || key === 'created_lt') && value instanceof Date) {
+              acc[key] = value.getTime().toString()
+            }
+            if (key === 'created_by' && 'value' in value) {
+              acc[key] = value.value
+            }
```
vardanbansal-harness, line 120: You're checking `value` key in `value`, but TypeScript doesn't know what value actually is. Add a type guard or cast.

```
if (key === 'created_by' && typeof value === 'object' && value && 'value' in value) {
...
```

harish-viswa: @vardanbansal-harness value is properly typed here and it is non-nullable
But it might be better to add typeof check
image

vardanbansal-harness (approved):

PR #1513, opened 2025-04-22
Switch component migration to the new design system

vivek-harness:
This PR modifies the Switch component based on the [new design system changes](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=622-85847&p=f&t=buWz95FoNgYqDKfY-0) and updates its usage across the app to reflect the migration changes.

https://github.com/user-attachments/assets/1f2a4356-6fb3-4350-beb7-0572b23e08f1

### DIFF
--- a/packages/ui/src/components/switch.tsx
+++ b/packages/ui/src/components/switch.tsx
@@ -22,6 +22,7 @@ const Switch = forwardRef<
           <label htmlFor={props.id || switchId} className="switch-label">
             {(props.required ? `${label} *` : label) || ''}
           </label>
+          {/* TODO: Design system: update to Text component once available */}
           <p className="switch-description">{description || ''}</p>
         </div>
       )}
--- a/packages/ui/src/components/switch.tsx
+++ b/packages/ui/src/components/switch.tsx
@@ -5,27 +5,22 @@ import * as SwitchPrimitives from '@radix-ui/react-switch'
 
 const Switch = forwardRef<
   ElementRef<typeof SwitchPrimitives.Root>,
-  ComponentPropsWithoutRef<typeof SwitchPrimitives.Root>
->(({ className, ...props }, ref) => (
-  <SwitchPrimitives.Root
-    className={cn(
-      `  
-      data-[state=checked]:bg-icons-success data-[state=unchecked]:bg-icons-6 
-      peer inline-flex h-[18px] w-8 shrink-0 cursor-pointer items-center !rounded-full 
-      border-2 border-transparent transition-colors      
-      disabled:cursor-not-allowed disabled:opacity-50
-      `,
-      className
+  ComponentPropsWithoutRef<typeof SwitchPrimitives.Root> & {
+    label?: string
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/switch.tsx
```diff
@@ -5,27 +5,22 @@ import * as SwitchPrimitives from '@radix-ui/react-switch'
...
+  }
+>(({ className, label, description, ...props }, ref) => (
+  <div className="switch-wrapper">
+    <SwitchPrimitives.Root className={cn('switch-root', className)} {...props} ref={ref}>
+      <SwitchPrimitives.Thumb className="switch-thumb" />
+    </SwitchPrimitives.Root>
+    {(label || description) && (
+      <div className="switch-label-wrapper">
+        <label className="switch-label">{(props.required ? `${label} *` : label) || ''}</label>
+        <label className="switch-description">{description || ''}</label>
```
praneshg239, line 20: This can be normal `Text` component.

packages/ui/tailwind-utils-config/components/switch.ts
```diff
@@ -0,0 +1,89 @@
...
+  '.switch-label-wrapper': {
+    display: 'flex',
+    flexDirection: 'column',
+    gap: 'var(--cn-spacing-1)'
+  },
+
+  '.switch-root': {
+    display: 'inline-flex',
+    position: 'relative',
+    width: 'var(--cn-size-9)',
```
praneshg239, line 17: @cxramina are we not adding tokens related to `Switch` component? Like `--cn-switch-width` etc,.

packages/ui/tailwind-utils-config/components/switch.ts
```diff
@@ -0,0 +1,89 @@
...
+    borderRadius: 'var(--cn-rounded-full)',
+
+    transition: 'transform',
+    'transition-duration': '100ms',
+
+    backgroundColor: `var(--cn-comp-selection-unselected-item)`,
+    borderColor: `var(--cn-comp-selection-unselected-border)`,
+
+    '&:where([disabled])': {
+      backgroundColor: `var(--cn-state-disabled-text)`,
```
praneshg239, line 56: why we have `text` color as background color?
cc: @cxramina

praneshg239 (changes requested):

packages/ui/tailwind-utils-config/components/switch.ts
cxramina: Screen Shot 2025-04-23 at 1 00 33 PM
it's text here, because it's a basically an item like "icon" which we use on bg element, it should be recognizable

packages/ui/tailwind-utils-config/components/switch.ts
cxramina: I haven't added them

If you think so, that we need to add them for consistency, I can fix that

@praneshg239

PR #1514, opened 2025-04-22
feat: Add "routes" to MFE context for proper navigation from Side nav items

vardanbansal-harness:
- Fix the links so that navigation works properly.
- Fix font weight for side nav items.

https://github.com/user-attachments/assets/2f3bf841-72ca-4e12-8edf-6ef80ad92cf3


abhinavrastogi-harness (approved):

apps/gitness/src/AppMFE.tsx
```diff
@@ -91,6 +91,11 @@ interface AppMFEProps {
   customPromises: Partial<{
     getCurrentUser: Promise<Unknown>
   }>
+  routes: Partial<{
+    toAccountSettings: Unknown
+    toOrgSettings: Unknown
+    toProjectSettings: Unknown
```
sans-harness, line 97: Do these need to be unknown?

sans-harness (approved):

apps/gitness/src/AppMFE.tsx
vardanbansal-harness: No, actually, these do have the known type. Will fix it.

PR #1516, opened 2025-04-22
feat: create secret now comes from the form framework

sans-harness:
https://github.com/user-attachments/assets/b9503e24-a626-4ab5-b5a5-2aa4a396001a


apps/design-system/src/utils/secrets/utils.tsx
```diff
@@ -0,0 +1,25 @@
+import { AnySecretDefinition } from '@harnessio/ui/views'
+
+import { secretsFormDefinition } from './secrets-form-schema'
+
+export function getHarnessSecretDefinition(): AnySecretDefinition {
```
abhinavrastogi-harness, line 5: why do we need this util?
I see this being used in connectors as well.. could we reuse it?
or can it be built into the form framework? cc @srdjan-harness

packages/ui/src/views/secrets/secret-entity-form.tsx
```diff
@@ -0,0 +1,107 @@
...
+    inputComponentFactory,
+    intent
+  } = props
+  const { t: _t } = useTranslationStore()
+
+  const onSubmit = (data: onSubmitSecretProps) => {
+    onFormSubmit?.(data)
+  }
+
+  const formDefinition = useMemo(() => {
```
abhinavrastogi-harness, line 39: why do we need this complex mapping function? formdef should work directly from config to the component.

packages/ui/src/views/secrets/secret-entity-form.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <EntityFormLayout.Footer className="border-none">
+              <div className="bg-cn-background-2 absolute inset-x-0 bottom-0 flex justify-between gap-x-3 p-4 shadow-md">
+                <Button variant="secondary" onClick={() => onBack?.()}>
+                  Back
+                </Button>
+                <Button onClick={() => rootForm.submitForm()}>Submit</Button>
+              </div>
+            </EntityFormLayout.Footer>
+          ) : (
+            <div className="border-cn-borders-3 mt-5 flex flex-row justify-end border-t pt-5">
```
abhinavrastogi-harness, line 99: how come edit footer styles are different from create footer on line 91?
can we move the condition to inside the footer?

apps/design-system/src/utils/secrets/utils.tsx
sans-harness: Can we look into de-duping this as a followup?

vardanbansal-harness: Pls pick this as a follow up.

vardanbansal-harness (approved):

PR #1546, opened 2025-04-25
refactor: update connectors to manage views and drawer state

andrewgolovanov:


packages/ui/src/views/unified-pipeline-studio/components/palette-drawer/uinfied-pipeline-step-palette-drawer.tsx
```diff
@@ -1,24 +1,43 @@
...
+    Footer: Drawer.Footer
+  },
+  false: {
+    Header: EntityFormLayout.Header,
+    Title: EntityFormLayout.Title,
+    Inner: 'div',
+    Footer: EntityFormLayout.Footer
+  }
+}
+
```
srdjan-harness, line 32: This parent could be useful if we render outside drawer.  For now may be not needed. All forms are inside right drawer. We gonna discuss this today. If need we can make one utils file with all drawer comps except root.

3em: I wasn’t sure if the `UnifiedPipelineStudioStepPalette` form component would only ever be used inside a `Drawer`, so I included this functionality via `componentsMap`.
Some of the other forms (from the new feature screens) might be rendered either in a `Drawer` or directly on the page.
That said, I can remove this condition if it turns out to be unnecessary

shaurya-harness (approved):

PR #1554, opened 2025-04-29
fix: [CODE-3571]: webhook routing

sans-harness:


vardanbansal-harness (approved):

packages/ui/src/views/repo/webhooks/webhook-list/components/repo-webhook-list.tsx
```diff
@@ -27,13 +28,16 @@ export function RepoWebhookList({
...
+  toRepoWebhookDetails,
+  toRepoWebhookCreate
 }: RepoWebhookListProps) {
   const { t } = useTranslationStore()
   const { navigate } = useRouterContext()
 
   const handleNavigate = () => {
-    navigate('create')
+    if (toRepoWebhookCreate) {
+      navigate(toRepoWebhookCreate())
```
abhinavrastogi-harness, line 39: can we use link on the webhook rows instead of `onclick + navigate`?

abhinavrastogi-harness (approved):

packages/ui/src/views/repo/webhooks/webhook-list/components/repo-webhook-list.tsx
sans-harness: Wrapping a <Table.Row> component with a <Link> element breaks the table’s structure — it causes all the content within that row to render as a single cell.

PR #1567, opened 2025-04-30
Add cn prefix to badge and button classes

praneshg239:
## Changes

Add `-cn` prefix to the component classes.

### Button

image

---

### SplitButton

image

---

### StatusBadge

image

---

### CounterBadge

image

### DIFF
--- a/packages/ui/tailwind-design-system.ts
+++ b/packages/ui/tailwind-design-system.ts
@@ -93,16 +93,16 @@ export default {
         },
         toast: {
           background: {
-              danger: 'hsla(var(--canary-toast-background-danger))'
+            danger: 'hsla(var(--canary-toast-background-danger))'
           },
           foreground: {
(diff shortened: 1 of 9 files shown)


packages/ui/tailwind-design-system.ts
```diff
@@ -376,8 +376,6 @@ export default {
         13: 'var(--cn-font-size-13)'
       },
       backgroundImage: {
-        'ai-button':
```
praneshg239, line 379: It is not being used now. Hence removed.

vivek-harness (approved):

packages/ui/tailwind-utils-config/components/badge.ts
knagurski, line 1: Not one for now, but we should consider breaking this out into `counter-badge.ts` and `status-badge.ts` since they're now unrelated components that might gravitate away from each other in look and function.

knagurski (approved):
👍🏻

PR #1574, opened 2025-05-02
feat: added support for genenric sort

harish-viswa:
Added support for sort with sort selection and multi-sort
TODO - Once single-sort design is ready, create new component under Sort

```
<Sort.Root {..props}> <br>
    <Sort.Select>       // list sort to select from
    <Sort.MultiSort>. //  Add additional sorts .. Update sort order
</Sort.Root>
```

This screenshot is just an example usage of sort and pr-list-page does not contain any sort 
image

### DIFF
--- a/packages/ui/locales/fr/component.json
+++ b/packages/ui/locales/fr/component.json
@@ -121,8 +121,9 @@
   },
   "sort": {
     "defaultLabel": "Trier",
-    "inputPlaceholder": "Trier par...",
+    "resetSort": "Reset sort",
     "buttonLabel": "Réinitialiser le tri",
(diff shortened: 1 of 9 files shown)


packages/ui/locales/fr/component.json
```diff
@@ -121,8 +121,9 @@
   },
   "sort": {
     "defaultLabel": "Trier",
-    "inputPlaceholder": "Trier par...",
+    "resetSort": "Reset sort",
```
vardanbansal-harness, line 124: nit: Add proper `fr` translation

vardanbansal-harness (approved):

PR #1587, opened 2025-05-06
Make checkbox optional by default if the prop is not passed

vivek-harness:
Changed the API to make the checkbox optional by default when the prop is not passed.

Previously, it made it required when the optional prop was not passed. This made it confusing since the API was in-consistent compared to the default html checkbox API where we need to set it `required` for making the checkbox required, or else it is optional by default.

As per the current API, we are not allowing the `required` prop, instead the user needs to use the `optional` prop.

This also fixes all the checkboxes where it made it required by default since the optional prop was not passed.

### DIFF
--- a/packages/ui/src/components/checkbox.tsx
+++ b/packages/ui/src/components/checkbox.tsx
@@ -17,6 +17,7 @@ interface CheckboxProps extends ComponentPropsWithoutRef<typeof CheckboxPrimitiv
 const Checkbox = forwardRef<ElementRef<typeof CheckboxPrimitive.Root>, Omit<CheckboxProps, 'required'>>(
   ({ className, label, caption, optional, ...props }, ref) => {
     const checkboxId = props.id || `checkbox-${Math.random().toString(36).slice(2, 11)}`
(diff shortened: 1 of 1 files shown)


knagurski:
I'm not sure I agree with this from a developer experience point of view. For boolean props, including should always mean `true` and omitting should always mean `false`. So if we are going to make things optional by default, then to override that, we need to have a `required` prop. Similarly, if we want things to be required by default, then to override that, we need to have an `optional` prop. Neither should be required and neither should require a value. I would consider needing to pass `optional={false}` to be an anti-pattern.

vivek-harness:
> I'm not sure I agree with this from a developer experience point of view. For boolean props, including should always mean `true` and omitting should always mean `false`. So if we are going to make things optional by default, then to override that, we need to have a `required` prop. Similarly, if we want things to be required by default, then to override that, we need to have an `optional` prop. Neither should be required and neither should require a value. I would consider needing to pass `optional={false}` to be an anti-pattern.

So then, can we make `optional` prop mandatory? Because, right now we are making it required accidentally by skipping the optional prop.

PR #1589, opened 2025-05-06
Textarea design migration

ankormoreankor:
##### Textarea design migration

- migrated component to the [new design](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=1474-32236&t=E2RFKoNJNn9iJ3qM-0)
- added conditional counter
- added `cn-` prefix to label class names
- added `disabled` state to `Label` and `FormCaption` components
- fixed label text overflow
- added new utils - `useMergeRefs` and `anyTypeOf`, organized the utils properly

image


packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,100 @@
...
+          ref={ref}
+          className={cn('cn-scrollbar', textareaVariants({ theme }), { 'resize-y': resizable }, className)}
+          disabled={disabled}
+          onChange={handleChange}
+          {...props}
+        />
+
+        {(disabled || caption) && <FormCaption disabled={disabled}>{caption}</FormCaption>}
+
+        {!disabled && error && <FormCaption theme="danger">{error}</FormCaption>}
```
3em, line 93: We previously discussed that we either display the caption or an error

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,100 @@
...
+    // Set initial count
+    useEffect(() => {
+      if (anyTypeOf(props.defaultValue, ['number', 'string'])) {
+        setCharactersCount(String(props.defaultValue))
+      }
+
+      if (anyTypeOf(props.value, ['number', 'string'])) {
+        setCharactersCount(String(props.value))
+      }
+    }, [props.defaultValue, props.value, setCharactersCount])
```
3em, line 58: We should only do the character count if `maxCharacters` is provided — otherwise, there’s no point in updating the state

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <textarea
+          id={id}
+          ref={ref}
+          className={cn('cn-scrollbar', textareaVariants({ theme }), { 'resize-y': resizable }, className)}
+          disabled={disabled}
+          onChange={handleChange}
+          {...props}
+        />
+
+        {(disabled || caption) && <FormCaption disabled={disabled}>{caption}</FormCaption>}
```
3em, line 91: It's a weird condition that we're still showing the Caption block when `disabled` is `true` — even if no `caption` is actually passed.

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: My idea is to prevent showing an error in the disabled field

ankormoreankor: Especially if it doesn't react on the theme changing

3em: You can still end up showing both the error and the error prop right now — but our logic should be: if there’s an error, it gets displayed and the caption is hidden.

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,115 @@
...
+      ...props
+    },
+    ref
+  ) => {
+    const [counter, setCounter] = useState(0)
+
+    const setCharactersCount = useCallback((value: string) => setCounter(value.length), [])
+
+    const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+      setCharactersCount(e.target.value)
```
3em, line 58: If `maxCharacters` isn’t provided, we don’t need to update the state here; 
And it’s better to move this check into `setCharactersCount`

ankormoreankor: FIxed

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: Fixed

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: Fixed

packages/ui/src/utils/utils.ts
```diff
@@ -150,3 +150,47 @@ export function generateAlphaNumericHash(length: number) {
...
+  const actualType = typeof value
+  return types.includes(actualType as JavaScriptType)
+}
+
+type PossibleRef<T> = MutableRefObject<T> | RefCallback<T> | ForwardedRef<T>
+
+/**
+ * A React hook that merges multiple refs into a single ref callback.
+ */
+export function useMergeRefs<T>(refs: Array<PossibleRef<T> | null | undefined>): RefCallback<T> {
```
3em, line 175: let's move it to separate file, same with `anyTypeOf`

ankormoreankor: Organized utils properly

packages/ui/src/utils/mergeUtils.ts
abhinavrastogi-harness, line 1: is this change needed for textarea? if not, lets remove it.

packages/ui/src/utils/typeUtils.tsx
abhinavrastogi-harness, line 1: is this change needed for textarea? if not, lets remove it.

packages/ui/src/utils/utils.ts
abhinavrastogi-harness, line 1: is this change needed for textarea? if not, lets remove it.

packages/ui/src/views/connectors/connectors-list/connectors-list.tsx
```diff
@@ -12,7 +12,7 @@ import {
   Text,
   Tooltip
 } from '@/components'
-import { timeAgo } from '@utils/utils'
+import { timeAgo } from '@/utils'
```
abhinavrastogi-harness, line 15: none of these import changes are required in this PR. lets remove it.

packages/ui/tailwind-utils-config/components/label.ts
```diff
@@ -1,24 +1,42 @@
 export default {
-  '.label': {
+  '.cn-label': {
```
abhinavrastogi-harness, line 2: label css changes will impact other inputs as well.
it this required for textarea specifically?
what about testing all other inputs?

packages/ui/tailwind-utils-config/components/textarea.ts
```diff
@@ -0,0 +1,104 @@
...
+    minHeight: '34px', // about 1 line of text
+    borderWidth: 'var(--cn-input-border)',
+    borderStyle: 'solid',
+    borderColor: 'var(--cn-border-2)',
+    backgroundColor: 'var(--cn-bg-2)',
+    '@apply font-body-normal': '',
+    color: 'var(--cn-text-1)',
+    whiteSpace: 'pre-wrap',
+    wordBreak: 'break-word',
+    resize: 'none',
```
abhinavrastogi-harness, line 43: why disallow resize forcefully? i think this goes against the principles of accessibility

packages/ui/src/utils/mergeUtils.ts
ankormoreankor: It's needed

packages/ui/src/utils/typeUtils.tsx
ankormoreankor: It's needed

packages/ui/src/utils/utils.ts
ankormoreankor: It's needed

packages/ui/tailwind-utils-config/components/label.ts
ankormoreankor: It won't impact other inputs
This change is needed for more stable use Label component in MFE

packages/ui/tailwind-utils-config/components/textarea.ts
ankormoreankor: We have a prop called `resizeable` in the [component API](https://gist.github.com/knagurski/7ce33d8c24f925e1d2f06c1729e1f18b), so this prop enables the resize if passed

packages/ui/src/views/connectors/connectors-list/connectors-list.tsx
ankormoreankor: Explained in [this thread](https://pixelpointworkspace.slack.com/archives/C084FDWFA21/p1746785725188729?thread_ts=1746699952.109979&cid=C084FDWFA21)

packages/ui/src/components/form-input/components/form-textarea.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { Textarea, TextareaProps } from '@components/form-primitives'
+import { useMergeRefs } from '@utils/mergeUtils'
+
+interface FormTextareaPropsType extends TextareaProps {
+  name: string
+}
+
+const FormTextarea = forwardRef<HTMLTextAreaElement, FormTextareaPropsType>((props, ref) => {
+  const inputRef = useRef<HTMLTextAreaElement | null>(null)
```
praneshg239, line 12: Do we need `inputRef` here? It is present in my implementation of `form-text-input.tsx` as well. It can be removed.

packages/ui/src/components/form-input/components/form-textarea.tsx
```diff
@@ -0,0 +1,37 @@
...
+      control={formContext.control}
+      render={({ field, fieldState }) => {
+        const error = fieldState.error?.message ?? props.error
+        return <Textarea {...{ ...props, ...field, error }} ref={mergedRef} theme={error ? 'danger' : 'default'} />
+      }}
+    />
+  )
+})
+
+FormTextarea.displayName = 'FormInput.Text'
```
praneshg239, line 35:
```suggestion
FormTextarea.displayName = 'FormInput.Textarea'
```

packages/ui/src/components/inputs/text-input.tsx
```diff
@@ -25,19 +25,23 @@ const TextInput = forwardRef<HTMLInputElement, TextInputProps>((props, ref) => {
...
+          {error}
+        </FormCaption>
       ) : warning ? (
-        <FormCaption theme="warning">{warning}</FormCaption>
+        <FormCaption disabled={restProps.disabled} theme="warning">
+          {warning}
+        </FormCaption>
       ) : caption ? (
-        <FormCaption>{caption}</FormCaption>
+        <FormCaption disabled={restProps.disabled}>{caption}</FormCaption>
```
praneshg239, line 44: can we de-structure `disabled` at the top  and pass it simply as `disabled` rather repeating as `restProps.disabled` in multiple places?

packages/ui/src/components/form-input/components/form-textarea.tsx
ankormoreankor: So you suggest remove it from both - TextInput and Textarea?

packages/ui/src/views/profile-settings/components/profile-settings-keys-create-dialog.tsx
```diff
@@ -65,26 +59,21 @@ export const ProfileSettingsKeysCreateDialog: FC<ProfileSettingsKeysCreateDialog
...
+          <FormInput.Text
+            id="identifier"
+            value={identifier}
+            {...register('identifier')}
+            placeholder={t('views:profileSettings.enterNamePlaceholder', 'Enter the name')}
+            label={t('views:profileSettings.newSshKey', 'New SSH key')}
+            autoFocus
+          />
+          <FormInput.Textarea
+            className="text-cn-foreground-1"
```
praneshg239, line 71: We can remove this. `color` should come from textarea design system class.

packages/ui/src/views/repo/pull-request/components/pull-request-header.tsx
```diff
@@ -69,10 +69,10 @@ export const PullRequestHeader: React.FC<PullRequestTitleProps> = ({
   return (
     <>
```
praneshg239, line 70: Is refactoring the only change made in this file?

praneshg239: Got it. Change in import path.

```js
import { timeAgo } from '@/utils'
```

ankormoreankor: Yeah, I'll run the linter to fix the rest of the changes👍

packages/ui/tailwind-utils-config/components/textarea.ts
```diff
@@ -0,0 +1,108 @@
+import { CSSRuleObject } from 'tailwindcss/types/config'
+
+// themes can be extended in the future
+const themes = ['danger'] as const
+
+function createInputThemeStyles() {
+  const styles: CSSRuleObject = {}
+
+  themes.forEach(theme => {
+    styles[`&:where(.cn-textarea-${theme}):not(textarea[disabled])`] = {
```
praneshg239, line 10: Little enhancement: 

```suggestion
    styles[`&:where(.cn-textarea-${theme}):not(:disabled)`] = {
```

praneshg239: In `input` we had like `:has(input[disabled])`, because we were applying styles for the container. As we are applying styles to `textarea` directly, we can use `:disabled`.

packages/ui/tailwind-utils-config/components/index.ts
```diff
@@ -11,5 +11,6 @@ export { default as linkStyles } from './link'
 export { default as avatarStyles } from './avatar'
 export { default as cardStyles } from './card'
 export { default as inputStyles } from './input'
+export { default as textareaStyles } from './textarea'
```
praneshg239, line 14: Scope for enhancement: 
Should we update every component as named export? So that we can avoid `default as` in all exports.

cc: @abhinavrastogi-harness @knagurski

packages/ui/src/views/repo/webhooks/webhook-create/components/create-webhooks-form-fields.tsx
```diff
@@ -26,54 +26,45 @@ export const WebhookToggleField: FC<WebhookFormFieldProps> = ({ register, watch,
 )
 
 export const WebhookNameField: FC<WebhookFormFieldProps & { disabled?: boolean }> = ({ register, disabled, t }) => (
-  <ControlGroup>
```
praneshg239, line 29: 👍

packages/ui/src/components/form-input/components/form-textarea.tsx
ankormoreankor: Fixed @praneshg239

packages/ui/src/components/form-input/components/form-textarea.tsx
ankormoreankor: Fixed

packages/ui/src/components/inputs/text-input.tsx
ankormoreankor: Fixed

packages/ui/src/views/repo/pull-request/components/pull-request-header.tsx
ankormoreankor: Linter ignores this file now, so those changes are needed🤷‍♂️

packages/ui/tailwind-utils-config/components/textarea.ts
ankormoreankor: Fixed

packages/ui/tailwind-utils-config/components/index.ts
abhinavrastogi-harness: agreed, that would be cleaner. we are not using the component directly anywhere anyay

ankormoreankor: It's not for this PR, right @praneshg239 ?

packages/ui/src/components/form-input/components/form-text-input.tsx
```diff
@@ -36,13 +22,10 @@ const FormTextInput = forwardRef<HTMLInputElement, FormTextInputPropsType>((prop
       name={props.name}
       control={formContext.control}
       render={({ field, fieldState }) => {
-        // Use proper React Hook Form reference handling
-        field.ref = setRefs
-
         // form error takes precedence over props.error
-        const errorMessage = fieldState.error?.message || props.error
+        const error = fieldState.error?.message ?? props.error
```
knagurski, line 26: Why change from `||` to `??`. The only real difference is to _allow_ the display of `0`, `false` and `''` to override an error deliberately passed in. Seems that we should only care about falsiness, not where it's _specifically_ `null` or `undefined`.

packages/ui/src/components/form-input/components/form-text-input.tsx
```diff
@@ -36,13 +22,10 @@ const FormTextInput = forwardRef<HTMLInputElement, FormTextInputPropsType>((prop
...
 
-        return <TextInput {...props} {...field} error={errorMessage} />
+        return <TextInput {...{ ...props, ...field, error }} ref={ref} theme={error ? 'danger' : 'default'} />
```
knagurski, line 28: Not a big fan of spreading into a spread. It takes a bit more thinking than just using props when possible.
```suggestion
        return <TextInput {...props} {...field} error={error} ref={ref} />
```

packages/ui/src/components/form-input/components/form-textarea.tsx
```diff
@@ -0,0 +1,33 @@
...
+    )
+  }
+
+  return (
+    <Controller
+      name={props.name}
+      control={formContext.control}
+      render={({ field, fieldState }) => {
+        const error = fieldState.error?.message ?? props.error
+        return <Textarea {...{ ...props, ...field, error }} ref={ref} theme={error ? 'danger' : 'default'} />
```
knagurski, line 25: Same as in the text input. Spreading into a spread isn't as obvious as just using available props and figuring out `theme` can be done in the component itself.
```suggestion
        return <Textarea {...props} {...field} error={error} ref={ref} />
```

packages/ui/src/components/form-primitives/form-caption.tsx
```diff
@@ -37,7 +46,10 @@ export const FormCaption = ({ theme = 'default', className, children }: PropsWit
   const effectiveIconName = theme === 'danger' ? 'cross-circle' : 'warning-triangle-outline'
 
   return (
-    <p className={cn(formCaptionVariants({ theme }), className)}>
+    <p
+      className={cn(formCaptionVariants({ theme }), { 'cn-caption-disabled': disabled }, className)}
+      aria-disabled={disabled}
```
knagurski, line 51: Not sure on the need to use `aria-disabled`. Screen readers won't see this tag as having anything to do with the input anyway, so it'll just be seen as a disabled paragraph of text, which isn't really a thing. Typically this is something that's used with interactive elements, not static text.

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,124 @@
...
+  defaultVariants: {
+    theme: 'default'
+  }
+})
+
+export interface TextareaProps extends TextareaHTMLAttributes<HTMLTextAreaElement> {
+  label?: string
+  theme?: VariantProps<typeof textareaVariants>['theme']
+  caption?: string
+  error?: string
```
knagurski, line 24: We need `warning` too
```suggestion
  error?: string
  warning?: string
```

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,124 @@
...
+        }
+      },
+      ref
+    ])
+
+    return (
+      <ControlGroup>
+        <div className="cn-textarea-label-wrapper">
+          {label && (
+            <Label disabled={disabled} optional={optional} htmlFor={id}>
```
knagurski, line 84: Since `id` is optional, can we detect that and generate an id to bind the label and input together? You can see how this can be done in the `TextArea` component. Could be worth extracting that functionality to a hook.

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,124 @@
...
+      node => {
+        if (!node) return
+
+        if (anyTypeOf(node.value, ['number', 'string'])) {
+          setCharactersCount(String(node.value))
+        }
+      },
+      ref
+    ])
+
```
knagurski, line 79: Can we add a line to determine the `theme` to use for the input to match the caption taking into account the `error` and `warning` props?

packages/ui/src/utils/mergeUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {} as T
+  }
+
+  return element
+}
+
+/**
+ * Helps to construct conditional arrays like this: [item1, condition && item2, item3].filter(item => !!item)
+ */
+export const wrapConditionalArrayElements = <T>(elements: T[], isPassing: boolean) => {
```
knagurski, line 17: I'm very much not a fan of the type of code this enables. The code it enables is very difficult to visually parse and understand and having the logic in a helper function doesn't save much in the way of typing, but does make it even less readable. I'm also not a fan of the naming since `wrapConditionalObjectElement` doesn't really match what it does... it won't return anything wrapped and it has nothing to do with `Element`s.

packages/ui/src/components/form-input/components/form-text-input.tsx
praneshg239: +1

I felt the current version is more readable than this. It looks bit complicated.

packages/ui/tailwind-utils-config/components/index.ts
praneshg239: Yes, it should be a separate PR. Not in scope of this.

packages/ui/src/utils/typeUtils.tsx
```diff
@@ -0,0 +1,15 @@
...
+ */
+type JavaScriptType = 'string' | 'number' | 'boolean' | 'undefined' | 'object' | 'function' | 'symbol' | 'bigint'
+
+/**
+ * Checks if a value is of one of the provided types
+ * @param value - The value to check
+ * @param types - Array of JavaScript type names to check against
+ * @returns True if the value matches any of the provided types, false otherwise
+ */
+export const anyTypeOf = (value: unknown, types: JavaScriptType[]): boolean => {
```
knagurski, line 12: I understand the idea of this utility function, but unfortunately TS won't follow the logic though and assert the correct type on the other side (in the version of TS we're on). It's probably safer to use `typeof myThing === 'string' || typeof myThing === 'number'` when the types matter. I wonder if it would be possible to write something generic like this that asserts the types using the `value is string | number` return syntax 🤔

packages/ui/src/components/form-input/components/form-text-input.tsx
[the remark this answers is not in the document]
ankormoreankor: Fixed

ankormoreankor: Fixed

packages/ui/src/components/form-input/components/form-textarea.tsx
```diff
@@ -0,0 +1,33 @@
...
+      'FormTextarea must be used within a FormProvider context through FormWrapper. Use the standalone Textarea component if form integration is not required.'
+    )
+  }
+
+  return (
+    <Controller
+      name={props.name}
+      control={formContext.control}
+      render={({ field, fieldState }) => {
+        const error = fieldState.error?.message ?? props.error
```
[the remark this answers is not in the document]
ankormoreankor: Fixed

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: Fixed

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: Fixed

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: Fixed

packages/ui/src/utils/mergeUtils.ts
ankormoreankor: Here's the difference in the screenshots below. For me, both variants look fine to read. But the variant without this helper annoys you with writing this `: {}` at the end of each ternary expression like this.

image

image

You also need to do type assertion, if a component has any required props
image

And the helper even helps with props suggestions
image

packages/ui/src/components/form-input/components/form-text-input.tsx
```diff
@@ -36,13 +22,10 @@ const FormTextInput = forwardRef<HTMLInputElement, FormTextInputPropsType>((prop
...
       render={({ field, fieldState }) => {
-        // Use proper React Hook Form reference handling
-        field.ref = setRefs
-
         // form error takes precedence over props.error
-        const errorMessage = fieldState.error?.message || props.error
+        const error = fieldState.error?.message || props.error
 
-        return <TextInput {...props} {...field} error={errorMessage} />
+        return <TextInput {...props} {...field} error={error} ref={ref} theme={error ? 'danger' : 'default'} />
```
knagurski, line 28: No need to pass in the `theme` since it's determined from in the `TextInput` component.
```suggestion
        return <TextInput {...props} {...field} error={fieldState.error?.message || props.error} ref={ref} />
```

packages/ui/src/components/form-input/components/form-textarea.tsx
```diff
@@ -0,0 +1,33 @@
...
+    )
+  }
+
+  return (
+    <Controller
+      name={props.name}
+      control={formContext.control}
+      render={({ field, fieldState }) => {
+        const error = fieldState.error?.message || props.error
+        return <Textarea {...props} {...field} error={error} ref={ref} theme={error ? 'danger' : 'default'} />
```
knagurski, line 25: We can worry about `theme` in the `Textarea` component
```suggestion
        return <Textarea {...props} {...field} error={fieldState.error?.message || props.error} ref={ref} />
```

packages/ui/src/utils/mergeUtils.ts
ankormoreankor: These helpers weren't added in this PR - they were moved. `wrapConditionalArrayElements` isn't used anywhere and can be easily dropped. `wrapConditionalObjectElement` has a few implementations.
Personally, I would use them (since I added them). But, you're the sheriff here 🙌

packages/ui/src/components/form-input/components/form-textarea.tsx
ankormoreankor: Sorry, updated

packages/ui/src/components/form-input/components/form-text-input.tsx
ankormoreankor: Sorry, updated

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,130 @@
...
+          {label && (
+            <Label disabled={disabled} optional={optional} htmlFor={id}>
+              {label}
+            </Label>
+          )}
+
+          {maxCharacters && (
+            <span
+              className={cn('cn-textarea-counter', {
+                'cn-textarea-counter-danger': theme === 'danger',
```
knagurski, line 98: We might not want to do this based on the `theme`. Some usages won't be in a scenario where there is validation triggering `error` messages. And there will be cases where the validation has nothing to do with the count. I'd change this to be based entirely on the count exceeding the max instead.
```suggestion
                'cn-textarea-counter-danger': counter > maxCharacters,
```

packages/ui/src/components/form-primitives/textarea.tsx
```diff
@@ -0,0 +1,130 @@
...
+          {...props}
+        />
+
+        {error ? (
+          <FormCaption theme="danger" disabled={disabled}>
+            {error}
+          </FormCaption>
+        ) : caption ? (
+          <FormCaption disabled={disabled}>{caption}</FormCaption>
+        ) : null}
```
knagurski, line 123: Can we add support for `warning` in here too?

packages/ui/src/views/repo/pull-request/compare/components/pull-request-compare-form.tsx
```diff
@@ -135,7 +131,7 @@ const PullRequestCompareForm = forwardRef<HTMLFormElement, PullRequestFormProps>
                   onDragLeave={handleDragLeave}
                   ref={dropZoneRef}
                 >
-                  <Textarea
+                  <FormInput.Textarea
                     id="description"
                     {...register('description')}
                     value={desc}
```
knagurski, line 137: Not one for this PR, but it seems weird that we're both using `register` to hook the state up to `useForm`, but also tracking and setting the values using standard state. Usually it's either/or, not both.

packages/ui/src/utils/typeUtils.tsx
ankormoreankor: I asked AI to do what you wondered for the TS version we use, and this is what he came up with. Looks nice

image

packages/ui/src/utils/typeUtils.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * @param types - Array of JavaScript type names to check against
+ * @returns True if the value matches any of the provided types, false otherwise
+ * @example
+ * const value: unknown = "hello";
+ * if (isAnyTypeOf(value, ['string', 'number'])) {
+ *   TypeScript now knows that value is string | number
+ *   console.log(value.toString());
+ * }
+ */
+export function isAnyTypeOf<T extends JavaScriptType[]>(value: unknown, types: [...T]): value is TypeFromArray<T> {
```
knagurski, line 39: Unfortunately this doesn't seem to work 😞  Screenshot 2025-05-19 at 14 34 10

packages/ui/src/views/repo/webhooks/webhook-create/components/create-webhooks-form-fields.tsx
```diff
@@ -26,54 +26,45 @@ export const WebhookToggleField: FC<WebhookFormFieldProps> = ({ register, watch,
...
-      autoComplete="new-password"
-      data-form-type="other"
-      id="payloadUrl"
-      {...register!('url')}
-      placeholder={t('views:repos.urlPlaceholder', 'https://example.com/harness')}
-      label={t('views:repos.urlLabel', 'Payload URL')}
-    />
-  </ControlGroup>
+  <FormInput.Text
+    autoComplete="new-password"
```
knagurski, line 50: Nothing to do with this PR, but this seems off here. Stating it should auto complete as a new password will trigger the browser to suggest a new password when this field is actually a URL field. Not sure why we'd want that. Same with other fields; the autoComplete values seem totally incorrect from my reading

packages/ui/src/views/repo/pull-request/compare/components/pull-request-compare-form.tsx
ankormoreankor: Agree

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: 😓

packages/ui/src/utils/typeUtils.tsx
ankormoreankor: Why are you trying to call `.toFixed` on a function?

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: Your comment made me think of what if we want a user to write a comment/description of no less than 40 characters - in this case, the counter should be in `"error"` state until it's `40/40`.

I understand your point that some errors cannot be related to the counter, but making the counter too smart or too complex in terms of validation can lead to a worse DX, imo.

knagurski: Oooohhhh, good point. So we might need a `minCharaters` prop that addresses that use-case. So with `maxCharacters` we can show `danger` with `counter > maxCharacters`, and with `minCharacters` we can show `danger` with `counter < minCharacters`. We might also want to think of how we can handle when it's not within the `minCharacters` and `maxCharacters`, but I think we'll need a different visual treatment for that.

knagurski: We can perhaps revisit as an enhancement and just keep it super-dumb without `cn-textarea-counter-danger` for now

packages/ui/src/utils/typeUtils.tsx
knagurski: Well spotted BTW!

knagurski: Just tried again and it works nicely! 👍🏻

packages/ui/src/utils/mergeUtils.ts
knagurski: It's not a blocker, so no need to change anything. We can worry about it later if it gets confusing.

apps/portal/src/components/docs-page/example.tsx
```diff
@@ -70,7 +72,7 @@ const Example: FC<ExampleProps> = ({ code, scope }) => {
   return (
     <div className="bg-cn-background-1 not-content my-12 overflow-hidden rounded-md border">
       <LiveProvider code={code} scope={scopeWithLayout} enableTypeScript>
-        <div className="grid place-items-center p-12">
+        <div className="w-full p-12 [&>div]:grid [&>div]:place-items-center">
```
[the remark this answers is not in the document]
knagurski: Ah yes, that's intentional. That div just makes sure we're centring without inadvertently dipping into the component that's being rendered. If we didn't have that div, the `grid` would apply to the example itself and could cause strange behaviour in some circumstances.

packages/ui/src/views/repo/webhooks/webhook-create/components/create-webhooks-form-fields.tsx
ankormoreankor: Should I address this issue in this PR? Tbh, I don't know the context of this part of the app🙇‍♂️

packages/ui/src/components/form-primitives/textarea.tsx
ankormoreankor: I would rather do the second small iteration when we decide on the behavior and API

ankormoreankor: Removed `cn-textarea-counter-danger`

knagurski (approved):
Great work!

PR #1615, opened 2025-05-12
add repo name to repo header

ankormoreankor:


packages/ui/src/views/repo/components/repo-header.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {isLoading && (
+        <>
+          <Skeleton className="h-8 w-28" />
+          <Skeleton className="h-6 w-14" />
+        </>
+      )}
+
+      {!isLoading && (
+        <>
+          <h2 className="font-heading-section text-cn-foreground-1">{name}</h2>
```
3em, line 27: Let's use `Text` — we'll be switching all tags to the `Text` component soon anyway

knagurski (approved):
I don't see any issues from a code POV

PR #1619, opened 2025-05-14
Multiselect component

sans-harness:
https://github.com/user-attachments/assets/44dedc15-7e13-43b7-adc5-9a0a1562323f

https://github.com/user-attachments/assets/e3c5781b-6913-48af-aee5-3384729e631f


abhinavrastogi-harness:
let's make sure to use the `Tag` component for all tags.. not buttons. 
today the sizes look different

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,437 @@
...
+  caption?: string
+  value?: MultiSelectOption[]
+  defaultValue?: MultiSelectOption[]
+  options?: MultiSelectOption[]
+  placeholder?: string
+  searchQuery?: string | null
+  setSearchQuery?: (query: string | null) => void
+  onChange?: (options: MultiSelectOption[]) => void
+  /** Limit the maximum number of selected options. */
+  maxSelected?: number
```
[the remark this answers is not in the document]
sans-harness: removing for now.

packages/ui/src/components/form-input/components/form-multi-select-v2.tsx
```diff
@@ -0,0 +1,63 @@
...
+    throw new Error(
+      'FormMultiSelectV2 must be used within a FormProvider context through FormWrapper. Use the standalone MultiSelectV2 component if form integration is not required.'
+    )
+  }
+  return (
+    <Controller
+      name={props.name}
+      control={formContext.control}
+      render={({ field }) => {
+        const setFieldRef = (element: MultiSelectRef | null) => {
```
praneshg239, line 40: would it be fine if we simply do

```
field.ref = setRefs;
```

packages/ui/src/components/form-input/components/form-multi-select-v2.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+interface FormMultiSelectV2PropsType
+  extends Omit<React.ComponentPropsWithoutRef<typeof MultiSelect>, 'value' | 'onChange'> {
+  name: string
+}
+
+const FormMultiSelectV2 = forwardRef<MultiSelectRef, FormMultiSelectV2PropsType>((props, ref) => {
+  const formContext = useFormContext()
+
+  const selectRef = useRef<MultiSelectRef | null>(null)
```
praneshg239, line 15: do we need this ref here? I see it is preset in `FormTextInput` as well. It can be removed, not needed here.

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+}
+
+export interface MultiSelectRef {
+  selectedValue: MultiSelectOption[]
+  input: HTMLInputElement
+  focus: () => void
+  reset: () => void
+}
+
+export const MultiSelect = forwardRef<MultiSelectRef, MultiSelectProps>(
```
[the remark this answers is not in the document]
sans-harness: Lets take this up as a followup

packages/ui/src/components/form-input/components/form-multi-select-v2.tsx
```diff
@@ -0,0 +1,55 @@
+import { forwardRef } from 'react'
+import { Controller, useFormContext } from 'react-hook-form'
+
+import { MultiSelect, type MultiSelectOption, type MultiSelectRef } from '@/components/multi-select-v2'
+
+interface FormMultiSelectV2PropsType
```
abhinavrastogi-harness, line 6: i think the form wrapper should also accept other form-related props like label, placeholder, error, warning, caption etc

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+            }
+          }
+          if (e.key === 'Enter' && input.value && !disallowCreation) {
+            if (
+              !options.some(option => option.key.toLowerCase() === input.value.toLowerCase()) &&
+              !(isControlled ? value : selected).some(s => s.key.toLowerCase() === input.value.toLowerCase())
+            ) {
+              let newOption: MultiSelectOption
+
+              if (input.value.includes(':')) {
```
abhinavrastogi-harness, line 132: we can pull this out into a util function which returns early as conditions match

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+                    id: input.value
+                  }
+                }
+              } else {
+                newOption = {
+                  key: input.value,
+                  id: input.value
+                }
+              }
+              const newOptions = [...(isControlled ? value : selected), newOption]
```
abhinavrastogi-harness, line 153: rename to something like `newValue`

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+              if (isControlled) {
+                onChange?.(newOptions)
```
abhinavrastogi-harness, line 155: `onChange` should be always called

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+            input.blur()
+          }
+        }
+      },
+      [handleUnselect, selected, disallowCreation, onChange, options, setInputValue, isControlled, value]
+    )
+
+    useEffect(() => {
+      if (open) {
+        document.addEventListener('mousedown', handleClickOutside)
```
abhinavrastogi-harness, line 174: are stopping propagation of mousedown from inside the dropdown somewhere?

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+  label?: string
+  caption?: string
+  value?: MultiSelectOption[]
+  defaultValue?: MultiSelectOption[]
+  options?: MultiSelectOption[]
+  placeholder?: string
+  searchQuery?: string | null
+  setSearchQuery?: (query: string) => void
+  onChange?: (options: MultiSelectOption[]) => void
+  disabled?: boolean
```
abhinavrastogi-harness, line 27: we should add a `loading` prop if we are not adding promise options support yet

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+        document.removeEventListener('mousedown', handleClickOutside)
+        document.removeEventListener('touchend', handleClickOutside)
+      }
+    }, [open])
+
+    useEffect(() => {
+      setOptions(arrayOptions ?? [])
+    }, [arrayOptions])
+
+    const [availableOptions, setAvailableOptions] = useState<MultiSelectOption[]>([])
```
abhinavrastogi-harness, line 191: maybe this can be deduped from `arrayOptions`

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+    useEffect(() => {
+      if (!options || options.length === 0) {
+        setAvailableOptions([])
+        return
+      }
+
+      const filteredOptions = options.filter(option =>
+        isControlled
+          ? !value?.some(selectedOption => selectedOption.id === option.id)
+          : !selected?.some(selectedOption => selectedOption.id === option.id)
```
abhinavrastogi-harness, line 202:
```suggestion
        !(isControlled ? value : selected)?.some(selectedOption => selectedOption.id === option.id)
```

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+      )
+
+      setAvailableOptions(filteredOptions)
+    }, [options, selected, isControlled, value, inputValue, searchQuery, open])
+    return (
+      <div className="cn-multi-select-outer-container">
+        <Label disabled={disabled}>{label}</Label>
```
abhinavrastogi-harness, line 209: maybe form elements like Label, Caption, error, warning etc can be moved to the FormWrapper

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+          {...commandProps}
+          onKeyDown={e => {
+            handleKeyDown(e)
+            commandProps?.onKeyDown?.(e)
+          }}
+          // filter={(_value, _search) => {
+          //   return 1
+          // }}
+          shouldFilter={false}
+          className={cn('h-auto overflow-visible bg-transparent', commandProps?.className)}
```
abhinavrastogi-harness, line 221: is the dropdown using a portal? it should.. and use `portalcontainer` for it

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+          >
+            <div className="cn-multi-select-tag-wrapper">
+              {(isControlled ? value : selected).map(option => {
+                return (
+                  <Tag
+                    id={String(option.id)}
+                    key={option.id}
+                    variant="secondary"
+                    size="sm"
+                    theme={option?.value ? 'purple' : undefined}
```
abhinavrastogi-harness, line 242: not sure if we should hardcode purple. maybe add `option.theme?` prop

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+            role="textbox"
+            tabIndex={disabled ? -1 : 0}
+            aria-label={placeholder}
+          >
+            <div className="cn-multi-select-tag-wrapper">
+              {(isControlled ? value : selected).map(option => {
+                return (
+                  <Tag
+                    id={String(option.id)}
+                    key={option.id}
```
abhinavrastogi-harness, line 239: not blocker: but let's check if we can avoid id. maybe use `key+value`? might need to discuss with BE/product for API confirmation

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+                        </Command.Item>
+                      )
+                    })}
+                  </Command.Group>
+                )}
+              </Command.List>
+            )}
+          </div>
+        </Command.Root>
+        <Caption className={disabled ? 'text-cn-foreground-disabled' : ''}>{caption}</Caption>
```
abhinavrastogi-harness, line 323: move out to formwrapper

packages/ui/tailwind-utils-config/components/multi-select-v2.ts
```diff
@@ -0,0 +1,50 @@
...
+        '&::placeholder': {
+          color: 'var(--cn-state-disabled-text)'
+        }
+      }
+    },
+
+    '&-dropdown': {
+      backgroundColor: 'var(--cn-bg-2)',
+      borderColor: 'var(--cn-border-2)',
+      '@apply border rounded-md shadow-md mt-1 overflow-hidden animate-in absolute top-1 z-10 w-full outline-none': '',
```
abhinavrastogi-harness, line 46: lets check if `z-10` works when inside drawers/dialogs

packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -0,0 +1,329 @@
...
+          const newOptions = value?.filter(s => s.id !== option.id) || []
+          onChange?.(newOptions)
+          option.onReset?.()
+        } else {
+          const newOptions = selected.filter(s => s.id !== option.id)
+          setSelected(newOptions)
+          option.onReset?.()
+        }
+      },
+      [onChange, selected, isControlled, value]
```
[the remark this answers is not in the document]
sans-harness: `Value` would likely be managed using state in the parent which should give us a new reference each time the state updates.

abhinavrastogi-harness (approved):

PR #1632, opened 2025-05-16
Card select component

vivek-harness:
This PR implements the `Card Select` component based on the [design specs](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=6163-5263&m=dev) and the [API specs](https://gist.github.com/knagurski/f77008dd0f49a175d6fd3846798c53c9).

https://github.com/user-attachments/assets/65e3109d-8f2a-491f-bbe5-3a924109c0b9


packages/ui/src/components/card-select.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+import { Icon } from './icon'
+import { Logo } from './logo'
+
+type CardSelectType = 'single' | 'multiple'
+
+interface CardSelectRootProps<T> {
+  type: CardSelectType
+  name?: string
+  value?: T extends 'single' ? unknown : unknown[]
```
abhinavrastogi-harness, line 14: can we use the same type inference pattern as Pagination to avoid using `unknown`?
something like

```ts
interface CardSingleSelectProps<T> {
  value: T,
  type: 'single'
}

interface CardMultiSelectProps<T> {
  value: Array<T>,
  type: 'multiple'
}

type CardSelectProps = CardSingleSelectProps | CardMultiSelectProps
```

It would vastly simplify the types for all the following interfaces and callbacks

packages/ui/src/components/card-select.tsx
```diff
@@ -0,0 +1,184 @@
...
+    orientation: 'vertical',
+    gap: 'md'
+  }
+})
+
+function CardSelectRoot<T extends CardSelectType>({
+  type,
+  orientation = 'vertical',
+  gap = 'md',
+  name = `card-select-${Math.random().toString(36).slice(2)}`,
```
abhinavrastogi-harness, line 75: we added a uuid util for another component recently.. let's reuse that, along with a `useMemo` to avoid the name changing on rerender

packages/ui/src/components/card-select.tsx
```diff
@@ -0,0 +1,184 @@
...
+}
+
+const CardSelectItem = forwardRef<HTMLLabelElement, CardSelectItemProps>(
+  ({ className, value, icon, logo, disabled: itemDisabled = false, children, ...props }, ref) => {
+    const { type, name, currentValue, disabled: groupDisabled, onValueChange } = useCardSelect()
+    const isDisabled = itemDisabled || groupDisabled
+    const checked = isChecked(value, currentValue)
+
+    return (
+      <label
```
abhinavrastogi-harness, line 131: can/should we use the existing `Label` component instead of html label?

packages/ui/tailwind-utils-config/components/card-select.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  },
+
+  '.cn-card-select-item': {
+    '@apply relative cursor-pointer transition-all flex items-center': '',
+    borderWidth: 'var(--cn-card-border)',
+    borderStyle: 'solid',
+    borderColor: 'var(--cn-border-2)',
+    borderRadius: 'var(--cn-card-default-radius)',
+    backgroundColor: 'var(--cn-bg-2)',
```
abhinavrastogi-harness, line 26: how come border and background are not using card-specific tokens?

abhinavrastogi-harness (approved):

PR #1634, opened 2025-05-16
Editing tokens

cxramina:
fixed transparent issue


packages/core-design-system/design-tokens/core/colors_hex.json
```diff
@@ -17,7 +17,7 @@
           }
         }
       },
-      "$value": "#0f0f11"
+      "$value": "{pure.black}"
```
praneshg239, line 20: do we need to change hex too?

praneshg239 (approved):

PR #1637, opened 2025-05-16
Fix minor bugs

sans-harness:
- Fixes routing on commits pages
- Fixes branch-dialog resetting
- Adds `tags`, `branches` API calls back on files page.


apps/gitness/src/framework/routing/types.ts
```diff
@@ -10,7 +10,8 @@ export enum RouteConstants {
   toImportMultipleRepos = 'toImportMultipleRepos',
   toRepositories = 'toRepositories',
   toRepoSummary = 'toRepoSummary',
-  toRepoCommits = 'toRepoCommits',
+  toRepoCommitsBranches = 'toRepoCommitsBranches',
```
abhinavrastogi-harness, line 13: better name would be `toRepoBranchCommits`

apps/gitness/src/pages-v2/repo/repo-sidebar.tsx
```diff
@@ -109,6 +116,45 @@ export const RepoSidebar = () => {
     queryParams: { git_ref: normalizeGitRef(fullGitRef || selectedBranchTag?.name) }
   })
 
+  const { data: { body: branches } = {} } = useListBranchesQuery({
```
abhinavrastogi-harness, line 119: not a blocker, but i'm not convinved about this solution. let's sync with BE team to figure out the expected behaviour.

abhinavrastogi-harness (approved):

PR #1642, opened 2025-05-19
Minor fixes to `multi-select`

sans-harness:
https://github.com/user-attachments/assets/73ce8598-e4a0-4a8a-967b-908002e9e52b

Screenshot 2025-05-19 at 7 05 50 PM

1. Adds error/warning theme
2. Removes dropdown when no options are provided
3. Adds documentation
4. Adds internal search 
5. Fixes caption/error states

### DIFF
--- a/packages/ui/src/components/multi-select-v2.tsx
+++ b/packages/ui/src/components/multi-select-v2.tsx
@@ -171,7 +171,7 @@ export const MultiSelect = forwardRef<MultiSelectRef, MultiSelectProps>(
             // Perform case-insensitive comparison to prevent duplicate options with different casing
             // This ensures that 'React' and 'react' would be considered the same option
             if (
-              // !options?.some(option => option.key.toLowerCase() === input.value.toLowerCase())
+              !options?.some(option => option.key.toLowerCase() === input.value.toLowerCase()) &&
               (!availableOptions || availableOptions.length === 0) &&
               !getSelectedOptions().some(s => s.key.toLowerCase() === input.value.toLowerCase())
             ) {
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/multi-select-v2.tsx
```diff
@@ -309,13 +313,19 @@ export const MultiSelect = forwardRef<MultiSelectRef, MultiSelectProps>(
...
-                    No results found
-                  </Command.Item>
+                ) : availableOptions?.length === 0 ? (
+                  disallowCreation ? (
+                    <Command.Item value="-" disabled>
+                      No results found
+                    </Command.Item>
+                  ) : (
+                    <Command.Item value="-" disabled>
+                      Press enter to create value
```
abhinavrastogi-harness, line 323: we will need product/copy guidance on this. for now let's change it to just `Press Enter to create`

packages/ui/tailwind-utils-config/components/multi-select-v2.ts
```diff
@@ -1,3 +1,30 @@
...
+function createMultiSelectThemeStyles() {
+  const styles: CSSRuleObject = {}
+
+  themes.forEach(theme => {
+    styles[`&:where(.cn-multi-select-${theme}):not(:has(input[disabled]))`] = {
+      borderColor: `var(--cn-border-${theme})`,
+      boxShadow: `var(--cn-ring-${theme})`,
+
+      '&:where(:hover)': {
+        borderColor: `var(--cn-border-${theme})`,
```
abhinavrastogi-harness, line 14: if bordercolor doesn't change, do we need to provide it again on hover and on focus?

packages/ui/tailwind-utils-config/components/multi-select-v2.ts
```diff
@@ -1,3 +1,30 @@
+import { CSSRuleObject } from 'tailwindcss/types/config'
+
+const themes = ['danger'] as const
```
abhinavrastogi-harness, line 3: we should also have the warning variants

PR #1643, opened 2025-05-20
Add layout component

sans-harness:
- Introduces the `Layout` component.


packages/ui/src/components/layout-v2.tsx
```diff
@@ -0,0 +1,182 @@
...
+  gapY,
+  wrap,
+  as: Comp = 'div',
+  ...props
+}: FlexProps & HTMLAttributes<HTMLDivElement>) => {
+  return (
+    <Comp
+      className={cn(
+        flexVariants({ direction, align, justify, wrap }),
+        gap && `gap-${gap}`,
```
abhinavrastogi-harness, line 90: instead of allowing all possible values, I think we should limit `gap` to a few predefined values, controlled via design tokens like `sm`, `md`, `lg` etc.

packages/ui/src/components/layout-v2.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+const Vertical: FC<VerticalProps & HTMLAttributes<HTMLDivElement>> = ({ children, className, spacing, ...props }) => {
+  return (
+    <Flex direction="column" className={cn(spacingVariants({ spacing }), className)} {...props}>
+      {children}
+    </Flex>
+  )
+}
+
+export const LayoutV2 = {
```
abhinavrastogi-harness, line 194: can we avoid creating a new `v2` component, and instead add Grid, Flex etc on the existing Layout component?

packages/ui/src/components/layout-v2.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+type GapSize = 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl'
+
+interface LayoutProps {
+  children?: ReactNode
+  className?: string
+  gap?: GapSize
+  gapX?: GapSize
+  gapY?: GapSize
+  as?: 'div' | 'span'
```
abhinavrastogi-harness, line 106: why restrict this to just div and span? i could see semantic html tags being useful as well.. like `article`, `nav` etc. we can remove this type, and allow any html element.

packages/ui/src/components/layout-v2.tsx
```diff
@@ -0,0 +1,199 @@
...
+  gap,
+  gapX,
+  gapY,
+  as: Comp = 'div',
+  ...props
+}: GridProps & HTMLAttributes<HTMLDivElement>) => {
+  return (
+    <Comp
+      className={cn(gridVariants({ align, justify }), gapVariants({ gap, gapX, gapY }), className)}
+      style={{
```
abhinavrastogi-harness, line 157: can we avoid inline styles by adding dynamic tailwind classes like `grid-cols-[{columns}]`?
https://tailwindcss.com/docs/grid-template-columns#using-a-custom-value

packages/ui/src/components/layout-v2.tsx
sans-harness: The existing component is being used across the app, we should take up replacing the older component and adopting this as a separate PR IMHO.

abhinavrastogi-harness (approved):

packages/ui/src/components/layout-v2.tsx
sans-harness: Added comment explaining.

PR #1649, opened 2025-05-20
Translation provider

pixel-buba:
- Create TranslationContext, TranslationProvider and useTranslation hook
- update apps for TranslationContext
- repalaced useTranslationStore to useTranslation for all apps
- remove props drilling useTranslationStore for all apps
like `useTranslationStore={useTranslationStore}` and `t={t}`


packages/ui/src/context/translation-context.tsx
knagurski, line 1: This implementation tightly couples us to `i18next`. Instead, we need to ask the app that is using this to pass along _only_ a function to call to return a specific string. With this new context and hook, we should be able to remove `i18next` as a dependency. This is important since our main target environment does not use `i18next` and has its own custom internationalisation framework that needs to be compatible.

Something like this...

```typescript
import { createContext, PropsWithChildren, useMemo } from 'react'

interface TranslationPayload {
  t: (key: string, fallback: string) => string
}

type TranslationProps = PropsWithChildren<{ t: (key: string) => string }>

function defaultTranslator(_: string, fallback: string) {
  return fallback
}

const TranslationContext = createContext<UseTranslationPayload>({
  t: defaultTranslator
})

export function TranslationProvider({
  t: translator,
  children
}: TranslationProps): TranslationPayload {
  const payload = useMemo(
    () => ({
      t: (key: string, fallback: string) => translator(key) || fallback
    }),
    [translator]
  )
  
  return (
    <TranslationContext.Provider value={payload}>
      {children}
    </TranslationContext.Provider>
  )
}

export function useTranslation(): TranslationPayload {
  return useContext(TranslationContext)
}
```

packages/ui/src/context/translation-context.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) => {
+  const store = useTranslationStore()
+  return <TranslationContext.Provider value={store}>{children}</TranslationContext.Provider>
+}
+
+export const useTranslation = () => {
+  const context = useContext(TranslationContext)
+
+  if (!context) {
+    throw new Error('useTranslation must be used within TranslationProvider')
```
knagurski, line 28: If there is no context, we need to be fault-tolerant and instead return a function that just returns the fallback value for the string. See example in the commebnt above.

packages/ui/src/context/translation-context.tsx
knagurski: How about we pass those things into the view (in a simplified state, not connected to i18next) and keep the translation context as a simple function.

apps/portal/src/lib/viewUtils.ts
```diff
@@ -0,0 +1,26 @@
+export const noop = () => void 0;
+
+export const asyncNoop = async () => void 0;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const mockT = (key: string, options?: { [key: string]: any }) => {
```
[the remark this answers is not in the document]
3em: No, they weren’t moved — just copied into `Portal`. We use `mockT` in the `Design System` as well, since we also mock translations there. But once we implement the changes Kevin requested above, we can actually remove `mockT` entirely from both the `Portal` and the `Design System` applications.

PR #1652, opened 2025-05-21
Add FormInput.Number component

praneshg239:
https://github.com/user-attachments/assets/e5ffe86e-a970-4d6b-a9e5-37f5a39a3ebc

### DIFF
--- a/apps/portal/src/components/docs-page/component-example.tsx
+++ b/apps/portal/src/components/docs-page/component-example.tsx
@@ -1,9 +1,6 @@
 import { type FC, useMemo } from "react";
-import { z } from "zod";
-import { zodResolver } from "@hookform/resolvers/zod";
 import * as components from "@harnessio/ui/components";
 import Example, { type ExampleProps } from "./example";
-import { useForm } from "react-hook-form";
 
 export type ComponentExampleProps = Omit<ExampleProps, "scope"> & {
   scope?: ExampleProps["scope"];
@@ -11,7 +8,7 @@ export type ComponentExampleProps = Omit<ExampleProps, "scope"> & {
 
 const ComponentExample: FC<ComponentExampleProps> = ({ code, scope }) => {
   const combinedScope = useMemo<ExampleProps["scope"]>(
-    () => ({ ...components, ...scope, zodResolver, z, useForm }),
+    () => ({ ...components, ...scope }),
     [scope],
   );
 
--- a/apps/portal/src/components/docs-page/component-example.tsx
+++ b/apps/portal/src/components/docs-page/component-example.tsx
@@ -1,7 +1,9 @@
 import { type FC, useMemo } from "react";
-
+import { z } from "zod";
+import { zodResolver } from "@hookform/resolvers/zod";
 import * as components from "@harnessio/ui/components";
 import Example, { type ExampleProps } from "./example";
+import { useForm } from "react-hook-form";
 
 export type ComponentExampleProps = Omit<ExampleProps, "scope"> & {
   scope?: ExampleProps["scope"];
@@ -9,7 +11,7 @@ export type ComponentExampleProps = Omit<ExampleProps, "scope"> & {
 
 const ComponentExample: FC<ComponentExampleProps> = ({ code, scope }) => {
   const combinedScope = useMemo<ExampleProps["scope"]>(
-    () => ({ ...components, ...scope }),
+    () => ({ ...components, ...scope, zodResolver, z, useForm }),
     [scope],
   );
 
--- a/apps/portal/src/content/docs/components/form.mdx
+++ b/apps/portal/src/content/docs/components/form.mdx
@@ -17,7 +17,7 @@ const inputRef = React.useRef<HTMLInputElement | null>(null);
 const formSchema = z.object({
 name: z.string().trim().min(3, "Input must be at least 3 characters"),
 
-    // 🚨 z.coerce.number() - It is mandatory for number fields
+    // 🚨 z.coerce.number() - Must be used for number fields
     // https://github.com/shadcn-ui/ui/issues/421#issuecomment-1561080201
     age: z.coerce.number().min(1, "Number must be at least 1"),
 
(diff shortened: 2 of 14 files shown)


packages/ui/src/components/inputs/number-input.tsx
```diff
@@ -103,6 +115,7 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
             hideStepper ? null : (
               <div className="flex flex-col">
                 <Button
+                  type="button"
```
praneshg239, line 118: Any `button` placed inside a form without `type` attribute acts as a submit button. Hence it is needed.

apps/portal/src/content/docs/components/form.mdx
```diff
@@ -7,63 +7,86 @@ The `Form` component is a container for form elements, providing a structured la
...
+const inputRef = React.useRef<HTMLInputElement | null>(null);
 
-//...
-
-// Define the validation schema with zod
 const formSchema = z.object({
-  inputField: z.string().min(3, "Input must be at least 3 characters"),
+name: z.string().trim().min(3, "Input must be at least 3 characters"),
+
+    // 🚨 z.coerce.number() - It is mandatory for number fields
```
praneshg239, line 20: It is necessary for a `number` input component. We need to explicitly highlight in the docs.

apps/portal/src/components/docs-page/component-example.tsx
```diff
@@ -1,15 +1,17 @@
...
+import { useForm } from "react-hook-form";
 
 export type ComponentExampleProps = Omit<ExampleProps, "scope"> & {
   scope?: ExampleProps["scope"];
 };
 
 const ComponentExample: FC<ComponentExampleProps> = ({ code, scope }) => {
   const combinedScope = useMemo<ExampleProps["scope"]>(
-    () => ({ ...components, ...scope }),
+    () => ({ ...components, ...scope, zodResolver, z, useForm }),
```
praneshg239, line 14: Passing the packages via `ComponentExample`'s `scope` prop is not working. Hence passing here. It will be refactored.

apps/portal/src/content/docs/components/form.mdx
praneshg239: Changes from `number` component will be returned as string always, in order to use it as number, we need to `coerce` via zod. 

https://github.com/shadcn-ui/ui/issues/421#issuecomment-1561080201

abhinavrastogi-harness (approved):

apps/portal/src/components/docs-page/component-example.tsx
knagurski: Seems like a good place to do it. The complication with using the `scope` prop is that the things you're passing in need to be part of the same "island" to be available, otherwise your scoped stuff will be part of the server rendering, while the component itself will be part of the browser rendering.

An alternative to what you've done here (which is fine IMO) would be to create a `ComponentExampleWithZodAndForm` which then can use `ComponentExample`'s `scope` prop.

```typescript
import { type FC, useMemo } from "react";
import { z } from "zod";
import { zodResolver } from "@hookform/resolvers/zod";
import * as components from "@harnessio/ui/components";
import Example, { type ExampleProps } from "./example";
import { useForm } from "react-hook-form";

const ComponentExampleWithZodAndForm: FC<ComponentExampleProps> = ({ code, scope }) => {
  return <ComponentExample code={code} scope={{ zodResolver, z, useForm }} />;
};

export default ComponentExampleWithZodAndForm;
```

That way you can use it selectively by swapping out your usage of `<DocsPage.ComponentExample>` for <DocsPage.ComponentExampleWithZodAndForm>`. It's up to you, but it's fine where it is IMO.

abhinavrastogi-harness:
we should add some documentation in the same doc, around best practices on how to use the form component.

abhinavrastogi-harness (approved):

apps/portal/src/content/docs/components/form.mdx
```diff
@@ -7,63 +7,144 @@ The `Form` component is a container for form elements, providing a structured la
 
 import { DocsPage } from "@/components/docs-page";
 
+<DocsPage.ComponentExampleWithZodAndForm
+  client:only
+  code={`() => {
```
knagurski, line 12: The formatting of this code block is a bit all over the place. Unfortunately prettier only sees it as a big string and not code, so you'll have to do it manually; or copy it to a new doc, run formatting and then paste it back in.

apps/portal/src/content/docs/components/form.mdx
knagurski, line 1: Let's make sure we double back to fix this file. The name of the file and the opening paragraph don't reflect the actual components. We also need to adopt the typical format and flow from other name-spaced components.

apps/portal/src/content/docs/components/form.mdx
```diff
@@ -7,63 +7,144 @@ The `Form` component is a container for form elements, providing a structured la
...
+      </FormWrapper>
+    </div>
+
+);
+}`}
+/>
+
+### Form with Zod and React Hook Form
+
 ```tsx
```
knagurski, line 96: This works, but can we stick to `typescript jsx` for consistency with other docs?

apps/portal/src/content/docs/components/form.mdx
```diff
@@ -7,63 +7,144 @@ The `Form` component is a container for form elements, providing a structured la
...
-          </div>
-        )}
-      </FormWrapper>
-    </div>
+    <FormWrapper {...formMethods} onSubmit={handleSubmit(onSubmit)}>
+      <FormInput.Text {...register("name")} label="Name" />
+      <FormInput.Number {...register("age")} label="Age" />
+      <FormInput.Textarea {...register("comment")} label="Comment" />
+      <Button type="submit">Submit</Button>
+    </FormWrapper>
```
knagurski, line 147: It looks like we don't tell people where any of this stuff is imported from.

packages/ui/src/components/inputs/number-input.tsx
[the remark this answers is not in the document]
praneshg239: it makes sense. I'll update it.

packages/ui/src/components/inputs/number-input.tsx
```diff
@@ -69,13 +63,31 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
...
 
     const handleDecrement = () => {
-      inputRef.current?.stepDown()
-      focusAtEnd(inputRef.current)
+      const input = inputRef.current
+      if (input) {
+        input.stepDown()
+        input.focus()
+        /**
+         * input.stepDown() will change the value of the input field, but it does not automatically trigger the onchange event callback.
```
knagurski, line 85: Ooohhh, that's interesting. I just played about with it and the built-in stepper _does_ trigger the `input` event. You'd have thought it would just be calling the `stepDown` function as well. Most form inputs in browsers are built-in custom elements with shadow dom and just plain HTML, CSS and JS controlling them. Funny that the stepper itself is injected and not part of the HTML.
Screenshot 2025-05-22 at 10 46 15
TIL 🙂

praneshg239: I felt the same _aha moment_! 💡

apps/portal/src/content/docs/components/form.mdx
praneshg239: sure. I thought of naming it as `FormWrapper`, but it will be scoped only to `FormWrapper`. As we need to cover all form components from `FormInput` I added generic name as `Form`.

apps/portal/src/content/docs/components/form.mdx
praneshg239: By doing so, prettier pipeline is failing. reverting it.

apps/portal/src/content/docs/components/form.mdx
knagurski: The format we've followed so far is to have a component per page. So we could have one for `FormWrapper` and another for `FormInput`. I'd do that and focus those docs on the specific components and also create a new "Forms" page in the "Foundations" section that explores using them together to build real-world forms with validation, similar to how the "Icons and Logos" section explores those concepts and links off to the individual components for `Icon` and `Logo`

packages/ui/src/components/pagination/pagination.tsx
```diff
@@ -103,6 +103,7 @@ const PaginationItems: FC<PaginationItemsProps> = ({
 interface PaginationBaseProps {
   className?: string
   t: TFunction
+  truncateLimit?: number
```
knagurski, line 106: Looks like this snuck back in

packages/ui/src/components/button.tsx
```diff
@@ -136,6 +137,7 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         )}
         ref={ref}
         disabled={disabled || loading}
+        type={type}
```
knagurski, line 140: Remember to update the docs for this component to reflect the change

knagurski (approved):
We can follow up on the docs changes in the next PR

praneshg239:
> We can follow up on the docs changes in the next PR

regarding the docs changes, I thought you'd be making the changes by creating `Foundations` section for forms. Anyway, we can discuss.

PR #1673, opened 2025-05-26
Dropdown design migration

ankormoreankor:
##### Dropdown menu design migration

[Design](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=782-76620&t=LSqOAK4EmeP0UErx-0) * [API](https://gist.github.com/knagurski/d661c20709c8a81d5325bd376503519a) * [Discussion](https://pixelpointworkspace.slack.com/archives/C088CNHL5L3/p1747651000459939)

- redesigned dropdown menu components using new design tokens
- added new pre-built dropdown menu items with avatar, icon, logo, etc
- submenus are handled automatically now, no need for extra wrappers
- `Group` and `RadioGroup` can be labeled with the `label` prop
- updated the component implementation across the app
- updated the docs

image

image

image

image

image


packages/ui/src/components/app-sidebar/sidebar-user.tsx
```diff
@@ -48,61 +47,32 @@ export function User({ user, openThemeDialog, openLanguageDialog, handleLogOut }
             align="end"
             sideOffset={4}
           >
-            <DropdownMenu.Label className="p-0 font-normal">
-              <div className="flex items-center gap-2 px-1 py-1.5">
-                <UserAvatar user={user} />
-              </div>
-            </DropdownMenu.Label>
+            <DropdownMenu.Header className="flex items-center gap-2 p-0 px-1 py-1.5 font-normal">
```
abhinavrastogi-harness, line 50: why do we need custom css here (and many other places). Are the screen designs not following the standard component design? 
I believe that since the component was standardised after the initial screens, we should just use the component as-it-is, and avoid adding overrides.

packages/ui/src/components/app-sidebar/sidebar-item.tsx
```diff
@@ -90,16 +70,11 @@ export const SidebarItem = ({
       {!collapsed && (
         <DropdownMenu.Root>
           <DropdownMenu.Trigger asChild>
-            <Sidebar.MenuAction className="right-[3px] text-sidebar-icon-3 hover:text-sidebar-icon-1" showOnHover>
+            <Sidebar.MenuAction className="text-sidebar-icon-3 hover:text-sidebar-icon-1 right-[3px]" showOnHover>
```
praneshg239, line 73: `right-[3px]` class is something which is already present,  do we still need this?

praneshg239: Yes, that I get it. Do we need this `3px` right? If it is for centering, can we do it via layout properties?

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+  })
+
+  return style
+}
+
+export default {
+  '.cn-dropdown-menu': {
+    zIndex: '50',
+    minWidth: 'var(--cn-dropdown-min-width)',
+    maxHeight: 'calc(var(--radix-dropdown-menu-content-available-height) - 4px)',
```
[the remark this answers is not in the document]
praneshg239: we have `--cn-dropdown-max-height`, can we use that here?

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+        borderBottom: '1px solid var(--cn-border-3)'
+      },
+
+      '&-footer': {
+        borderTop: '1px solid var(--cn-border-3)'
+      }
+    },
+
+    '&-content': {
+      maxHeight: 'var(--cn-dropdown-max-height)'
```
praneshg239, line 57: Got it, it is being used here.

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+        width: 'var(--cn-badge-indicator-size-default)',
+        height: 'var(--cn-badge-indicator-size-default)',
+        borderRadius: 'var(--cn-rounded-full)',
+        flexShrink: '0',
+
+        ...createIndicatorVariantStyles()
+      }
+    },
+
+    '&-group-label': {
```
praneshg239, line 110: can we wrap these items with `:where`? So that, if needed the styles can be easily overridden without `!important`.

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+      paddingTop: 'var(--cn-dropdown-item-py)',
+      paddingLeft: 'var(--cn-dropdown-item-px)',
+      paddingRight: 'var(--cn-dropdown-item-px)',
+      paddingBottom: 'var(--cn-spacing-1)',
```
praneshg239, line 114: Do we need different bottom padding?

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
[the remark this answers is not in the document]
ankormoreankor: @praneshg239 `--cn-dropdown-max-height` describes the max height of the content area and doesn't include footer and header, which can be any height, and we can end up in a situation where we see no items at all.

Radix's dropdown provides the token `--radix-dropdown-menu-content-available-height` with calculated maximum height, which is the size between the menu content top position and the bottom of the window. If you don't use it, it's easy to face a situation when the bottom of the dropdown menu is way below/above the window.

It doesn't cost us any, so I don't see any reason not to use it🤷‍♂️

image

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+
+      '&-subtrigger': {
+        '&:hover': {
+          cursor: 'default'
+        }
+      },
+
+      '&:where([data-disabled])': {
+        opacity: '0.5'
```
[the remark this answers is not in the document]
cxramina: Screen Shot 2025-06-03 at 3 43 47 PM
@ankormoreankor it's strange that you see it like that in Figma, this value is actually tokenized

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+
+      '& .cn-radio-item, & .cn-checkbox-root': {
```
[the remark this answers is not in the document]
ankormoreankor: I'm overriding radio and checkbox styles, so `:where` doesn't fit here🫤 I need higher specificity

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+    backgroundColor: 'var(--cn-bg-3)',
+    boxShadow: 'var(--cn-shadow-4)',
+    '@apply data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2':
+      '',
+
+    '&-container': {
+      padding: 'var(--cn-dropdown-container)',
+
+      '&-header': {
+        borderBottom: '1px solid var(--cn-border-3)'
```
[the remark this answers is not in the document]
ankormoreankor: Fixed

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
```diff
@@ -0,0 +1,141 @@
...
+        height: 'var(--cn-badge-indicator-size-default)',
+        borderRadius: 'var(--cn-rounded-full)',
+        flexShrink: '0',
+
+        ...createIndicatorVariantStyles()
+      }
+    },
+
+    '&-group-label': {
+      paddingTop: 'var(--cn-dropdown-item-py)',
```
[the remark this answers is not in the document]
ankormoreankor: FIxed

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
ankormoreankor: Fixed

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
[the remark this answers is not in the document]
ankormoreankor: Fixed

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
ankormoreankor: Sure, thanks🫶

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+  title,
+  description,
+  label,
+  shortcut,
+  checkmark,
+  withSubIndicator
+}: DropdownBaseItemProps) => (
+  <div className={cn('cn-dropdown-menu-base-item', className)}>
+    {children}
+    <div className="grid w-full">
```
abhinavrastogi-harness, line 135: we can now use the `<Layout.Grid>` component directly instead of divs with classes

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+      {typeof title === 'string' ? (
+        <Text className="grid" color="foreground-1">
```
abhinavrastogi-harness, line 137: why do we need a `grid` classes on a string?
also, can we use a text variant intead of adding a color?

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+          {title}
+        </Text>
+      ) : (
+        title
+      )}
+      {typeof description === 'string' ? <Text color="foreground-2">{description}</Text> : description}
```
abhinavrastogi-harness, line 143: same here.. let's use a Text variant

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+      <DropdownBaseItem {...{ title, description, label, shortcut, checkmark, withSubIndicator: withChildren }}>
+        <div className="cn-checkbox-root" {...{ 'data-state': checkedDataState }}>
+          {checked && (
+            <div className="cn-checkbox-indicator" {...{ 'data-state': checkedDataState }}>
+              {checked === 'indeterminate' ? (
+                <Icon name="minus" className="cn-checkbox-icon" skipSize />
+              ) : (
+                <Icon name="check" className="cn-checkbox-icon" skipSize />
+              )}
+            </div>
```
abhinavrastogi-harness, line 239: can we use the `<Checkbox />` component instead of a div with icons?

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+DropdownMenuAvatarItem.displayName = displayNames.avatarItem
+
+interface DropdownMenuLogoItemProps extends Omit<DropdownMenuItemProps, 'prefix'> {
+  logo: LogoProps['name']
+  original?: LogoProps['original']
+}
+
+const DropdownMenuLogoItem = forwardRef<ElementRef<typeof DropdownMenuPrimitive.Item>, DropdownMenuLogoItemProps>(
+  ({ logo, original, ...props }, ref) => (
+    <DropdownMenuItem ref={ref} {...props} prefix={<Logo size={20} name={logo} original={original} />} />
```
abhinavrastogi-harness, line 351: should we allow user to pass custom props to override hardcoded values like the size here?

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+    | 'blue'
+    | 'purple'
+    | 'brown'
+    | 'cyan'
+    | 'indigo'
+    | 'lime'
+    | 'mint'
+    | 'orange'
+    | 'pink'
+    | 'violet'
```
abhinavrastogi-harness, line 380: we should use the existing `StatusBadge` colors here instead of hardcoded a new custom list

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+
+const DropdownMenuSpinner = ({ className, ...props }: HTMLAttributes<HTMLDivElement>) => (
+  <div className={cn('cn-dropdown-menu-spinner', className)} {...props}>
+    <Icon className="animate-spin" name="spinner" />
+  </div>
+)
+DropdownMenuSpinner.displayName = displayNames.spinner
+
+const DropdownMenuNoOptions = ({ className, children, ...props }: HTMLAttributes<HTMLDivElement>) => (
+  <div className={cn('cn-dropdown-menu-no-options', className)} {...props}>
```
abhinavrastogi-harness, line 428: do we need this div? can we not add this class on the `Text` directy?

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+    <Text className="text-cn-foreground-3">{children ?? 'No options available'}</Text>
```
abhinavrastogi-harness, line 429: can we use an existing variant here?

packages/ui/src/components/dropdown-menu.tsx
knagurski: The `XItem` components are opinionated specialisations of the `Item` component, so we want to keep the customisation options to a minimum. By letting devs customise the sizes etc, we'll start seeing inconsistencies creeping in. The `Item` component can still be used by the dev to BYO `XItem` for their needs, but these ones keep them on the rails.

packages/ui/src/components/dropdown-menu.tsx
ankormoreankor: Radix's checkboxes and radios are buttons with `aria-labels` under the hood. Having a clickable element inside a clickable menu item can lead to problems with focus when using keyboard navigation, which I encountered while working on the component.

image

packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -1,199 +1,438 @@
...
+            ref={ref}
+            className={cn('cn-dropdown-menu-item cn-dropdown-menu-item-subtrigger', className)}
+            onClick={e => {
+              e.stopPropagation()
+              e.preventDefault()
+
+              if (props.disabled) return
+
+              onClick?.(e)
+              onCheckedChange?.(checked === 'indeterminate' ? false : !checked)
```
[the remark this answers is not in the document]
ankormoreankor: This event handler is called when the checked state changes, so it should be inverted. And it only returns a boolean

packages/ui/src/components/dropdown-menu.tsx
ankormoreankor: Currently, StatusBadge uses only 6 colors, which are inherited from the Badge component. I can extend the Badge component with these colors and then update my component if its okay

image

image

cc @cxramina

packages/ui/src/components/dropdown-menu.tsx
ankormoreankor: Good catch👍
We can use `Text` directly

packages/ui/src/components/dropdown-menu.tsx
ankormoreankor: Yeah, `grid` is not needed here. That was an experiment

ankormoreankor: @abhinavrastogi-harness colors are not a part of variants. Text variants define font properties like size, height, weight, etc. Colors and variants can be combined then. That's why they are added separately.

Here, `Text` uses the default variant, which is `body-normal` with non-default color `cn-text-foreground-1`

packages/ui/src/components/dropdown-menu.tsx
ankormoreankor: Answered here - https://github.com/harness/canary/pull/1673#discussion_r2126448645

Also removed color prop, since `foreground-2` is a default value

packages/ui/src/components/dropdown-menu.tsx
ankormoreankor: Answered here - https://github.com/harness/canary/pull/1673#discussion_r2126448645
Changed to `color` property

packages/ui/tailwind-utils-config/components/dropdown-menu.ts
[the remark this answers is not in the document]
ankormoreankor: Finalized here - https://pixelpointworkspace.slack.com/archives/C084FDWFA21/p1749044850712009?thread_ts=1748460593.541219&cid=C084FDWFA21

packages/ui/src/components/dropdown-menu.tsx
cxramina: @ankormoreankor @praneshg239 @abhinavrastogi-harness 
No, it shouldn’t be a tag or a status badge – the status badge uses specific themes.
I shared with @ankormoreankor in Slack that I have a dedicated color-badge-indicator component in Figma:
https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System–3.0-?m=auto&node-id=8917-115591&t=LUyE8jMiEy5qLGeY-1

This component actually uses tokens that aren’t present in tag or status badge (partially in status badge) components:
--cn-set-{color}-solid-bg for each color variant

ankormoreankor: Seems to me I did this component right

abhinavrastogi-harness (approved):

PR #1674, opened 2025-05-26
Error states for Radio and Checkbox

vivek-harness:
The error states were added later to the Figma designs for Radio and Checkbox. This PR implements those error states.

https://github.com/user-attachments/assets/3c02353f-6038-45d0-9071-51c709e76692


apps/portal/src/content/docs/components/checkbox.mdx
```diff
@@ -78,6 +80,25 @@ import { DocsPage } from "@/components/docs-page";
             onCheckedChange={setCheckedWOLabel}
           />
         </div>
+
+        {/* Error state */}
+        <div className="space-y-2">
+          <h4 className="text-sm font-medium">Error state</h4>
+          <div className="flex items-center space-x-8">
+            <Checkbox
```
abhinavrastogi-harness, line 88: we should document error status with the Form Checkbox Group component, so that it shows the error message, and not just the red highlight on the checkbox

abhinavrastogi-harness (approved):

PR #1680, opened 2025-05-27
Form HOC

praneshg239:


packages/ui/src/components/form-input/index.ts
```diff
@@ -1,24 +1,23 @@
-import { FormCheckbox, type FormCheckboxPropsType } from './components/form-checkbox'
```
abhinavrastogi-harness, line 1: can we delete all the `form-x` components now? I wonder how much cleanup this will result in.
wish we had tests for this to verify easily if everything works as expected!

abhinavrastogi-harness (approved):

PR #1682, opened 2025-05-27
refactor: updated layout usage to layoutV2 and deprecated the existing layout

harish-viswa:
For gap properites, moved it to nearest available variant from layout
gap-1, gap1.5, gap2.5 -> xs
gap7 -> xl

_file-viewer-control-bar_: gap-2 -> xs
image

_Toast_ 
image

test-connection-dialog
image

pull-request-compare-page
gap-x-1.5 -> xs
image


packages/ui/src/components/file-toolbar-actions.tsx
```diff
@@ -17,7 +17,7 @@ export const FileToolbarActions: FC<FileToolbarActionsProps> = ({
   showEdit = false
 }) => {
   return (
-    <Layout.Horizontal gap="gap-0">
+    <Layout.Horizontal gap="none">
```
abhinavrastogi-harness, line 20: i believe that `gap` defaults to `none`? cc @sans-harness can you confirm pls?

packages/ui/src/components/toast/toaster.tsx
```diff
@@ -8,7 +8,7 @@ export function Toaster() {
     <>
       {toasts.map(({ id, title, description, action, variant, showIcon = true, ...props }) => (
         <Toast.Root key={id} variant={variant} {...props}>
-          <Layout.Horizontal gap="space-x-2.5" className="items-center">
+          <Layout.Horizontal gapX="xs" align="center">
```
abhinavrastogi-harness, line 11: i believe we can just use `gap`.. horizontal layout will not have vertical gap anyway.

packages/ui/src/views/connectors/components/connector-test-connection-dialog.tsx
```diff
@@ -79,18 +79,18 @@ export const ConnectorTestConnectionDialog = ({
         </Dialog.Header>
         <Dialog.Description className="gap-y-0.5">
           <div className="text-sm font-normal text-cn-foreground-4">
-            <Layout.Horizontal className="items-center justify-between space-x-2">
-              <div className="mb-2.5 flex flex-row items-start space-x-2">
+            <Layout.Horizontal gapX="xs" align="center" justify="between">
+              <Layout.Horizontal gapX="xs" align="start" className="mb-2.5">
```
abhinavrastogi-harness, line 83: nested layouts raise a yellow flag for me.. can we not manage with a single layout?

abhinavrastogi-harness (approved):

packages/ui/src/components/toast/toaster.tsx
harish-viswa: makes sense, will update

packages/ui/src/views/connectors/components/connector-test-connection-dialog.tsx
harish-viswa: That can be done, updated

PR #1696, opened 2025-05-28
add client side string validation and button disable to name field for create repo form

jake-bassett:
https://github.com/user-attachments/assets/12acb092-bccd-4b3c-9c69-0257d81b03bd


packages/ui/src/views/repo/repo-create/index.tsx
```diff
@@ -20,11 +20,15 @@ import {
...
+import { isEmpty } from 'lodash-es'
 import { z } from 'zod'
 
 // Define the form schema with optional fields for gitignore and license
 const formSchema = z.object({
-  name: z.string().min(1, { message: 'Please provide a name' }),
+  name: z
+    .string()
+    .min(1, { message: 'Please provide a name' })
+    .regex(/^[a-z0-9-_.]+$/i, { message: 'Name can only contain the following characters [a-zA-Z0-9-_.]' }),
```
abhinavrastogi-harness, line 31: not a blocker, but ideally we should not show a regex to the user.

abhinavrastogi-harness (approved):

PR #1705, opened 2025-05-30
fix: added new gap variants for layout

harish-viswa:


packages/ui/src/components/layout.tsx
```diff
@@ -70,32 +70,38 @@ const gapVariants = cva('', {
   variants: {
     gap: {
       none: 'gap-0',
-      xs: 'gap-[var(--cn-layout-xs,var(--cn-spacing-2))]',
-      sm: 'gap-[var(--cn-layout-sm,var(--cn-spacing-3))]',
-      md: 'gap-[var(--cn-layout-md,var(--cn-spacing-4))]',
-      lg: 'gap-[var(--cn-layout-lg,var(--cn-spacing-6))]',
-      xl: 'gap-[var(--cn-layout-xl,var(--cn-spacing-8))]'
+      '2xs': 'gap-[var(--cn-layout-2xs)]',
+      xxs: 'gap-[var(--cn-layout-xxs)]',
```
abhinavrastogi-harness, line 74: I think `2xs` and `xxs` mean the same thing.
did you mean to use `3xs`?

packages/ui/src/components/layout.tsx
```diff
@@ -70,32 +70,38 @@ const gapVariants = cva('', {
...
+      xl: 'gap-[var(--cn-layout-xl)]'
     },
     gapX: {
       none: 'gap-x-0',
-      xs: 'gap-x-[var(--cn-layout-xs,var(--cn-spacing-2))]',
-      sm: 'gap-x-[var(--cn-layout-sm,var(--cn-spacing-3))]',
-      md: 'gap-x-[var(--cn-layout-md,var(--cn-spacing-4))]',
-      lg: 'gap-x-[var(--cn-layout-lg,var(--cn-spacing-6))]',
-      xl: 'gap-x-[var(--cn-layout-xl,var(--cn-spacing-8))]'
+      '2xs': 'gap-x-[var(--cn-layout-2xs)]',
```
abhinavrastogi-harness, line 83: i am thinking, can we remove `gapX` and `gapY` entirely? any usecase where they would be different?

abhinavrastogi-harness (commented):
there was a conversation about removing `spacing` prop as well.. can you check if that needs to be done?

packages/ui/src/components/layout.tsx
harish-viswa: In grids, we might have different gap properties for rows and cols right?

PR #1707, opened 2025-05-30
use timeAgo wrapper for timestamps to allow hover details

jake-bassett:


apps/gitness/src/pages-v2/repo/transform-utils/repo-list-transform.ts
```diff
@@ -12,7 +10,7 @@ export const transformRepoList = (repoList: RepoRepositoryOutput[]): RepositoryT
     stars: 0,
     forks: repo.num_forks || 0,
     pulls: repo.num_pulls || 0,
-    timestamp: repo.updated ? timeAgoFromEpochTime(repo.updated) : '',
+    timestamp: repo.updated ? new Date(repo.updated).toISOString() : '',
```
abhinavrastogi-harness, line 13: how come we are doing `Date().toISOString()` in some places but not others?

PR #1711, opened 2025-06-02
feat: replace hardcoded values with design variants in `Spacer`

sans-harness:


packages/ui/src/components/spacer.tsx
```diff
@@ -3,36 +3,35 @@ import { ComponentProps } from 'react'
...
-      8: 'mt-8',
-      9: 'mt-9',
-      10: 'mt-10',
-      11: 'mt-11',
-      12: 'mt-12',
-      13: 'mt-[3.25rem]',
-      14: 'mt-14',
-      15: 'mt-[3.75rem]',
-      16: 'mt-16'
+      1: 'mt-[var(--cn-spacing-1)]',
```
abhinavrastogi-harness, line 9: let's make the API consistent by using `sm`, `md`, `lg` etc. we can reduce the number of variants to match what's available.

abhinavrastogi-harness: actually, nevermind. that would be a breaking changes requiring a large refactor.

abhinavrastogi-harness (approved):

packages/ui/src/components/spacer.tsx
abhinavrastogi-harness: it might be better to define tailwind tokens instead of doing inline styling like `mt-[x]`

sans-harness: As discussed merging for now. Will make updates based on discussion in this thread once changes are finalized: https://harness.slack.com/archives/C088CNHL5L3/p1748895938145099

PR #1714, opened 2025-06-03
Select design migration

ankormoreankor:
##### Select component design migration

[Design](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=8818-74789&t=aRc10KiKSGy8oaLM-0) * [API](https://gist.github.com/abhinavrastogi-harness/984c6511de5f989d651cd7a7ef2e0c59) * [Discussion](https://pixelpointworkspace.slack.com/archives/C088CNHL5L3/p1748603535296689)

Key changes:
- the composition-based API was replaced with the props-based one
- supports themes like `warning` and `danger`
- can show `loading` state and work with async options
- optionRenderer can be used for complex scenarios
More features can be found in the [Select docs](https://deploy-preview-1714--harness-design.netlify.app/components/select/)

image

image

image

image


packages/ui/src/components/form-primitives/select_v2.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+      <DropdownMenu.Root
+        open={isOpen}
+        onOpenChange={open => {
+          setIsOpen(open)
+          if (!open) setSearchQuery('')
+        }}
+      >
+        <DropdownMenu.Trigger id={id} className={cn(selectVariants({ theme }), className)} disabled={disabled}>
+          <span className={cn('cn-select-value', { 'cn-select-placeholder': !selectedOption })}>
```
3em, line 185: Let's use the `Text` component and pass the `color` and `truncate` props to it.

packages/ui/src/components/form-primitives/select_v2.tsx
```diff
@@ -0,0 +1,240 @@
...
+          {isItemsLoading && <DropdownMenu.Spinner />}
+
+          {isNoItems && (
+            <DropdownMenu.NoOptions>{searchQuery ? 'No results found' : 'No options available'}</DropdownMenu.NoOptions>
+          )}
+
+          {isWithItems &&
+            filteredOptions.map((option, index) => (
+              <DropdownMenu.Item
+                key={index}
```
3em, line 213: Let's not use the `index` — maybe we can stringify the `value` instead

packages/ui/src/components/form-primitives/select_v2.tsx
```diff
@@ -0,0 +1,240 @@
...
+                title={typeof option.label === 'string' ? option.label : ''}
+                disabled={option.disabled}
+                onSelect={() => handleSelect(option.value)}
+                checkmark={option.value === selectedValue}
+              >
+                {renderOption(option)}
+              </DropdownMenu.Item>
+            ))}
+
+          {isLoadingMoreItems && <DropdownMenu.Spinner />}
```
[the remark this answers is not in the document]
ankormoreankor: Fixed

apps/design-system/src/pages/view-preview/view-settings.tsx
```diff
@@ -55,30 +62,24 @@ const ViewSettings: FC<ViewSettingsProps> = ({ routes }) => {
...
-                  {Object.entries(group.items).map(([path, { label }]) => (
-                    <Select.Item key={path} value={path}>
-                      {label}
-                    </Select.Item>
-                  ))}
-                </Select.Group>
-              ))}
-            </Select.Content>
-          </Select.Root>
+          <Select<string>
```
3em, line 65: Why specify `string` if it’s already the default in `Select`?

packages/ui/src/components/form-primitives/select.tsx
```diff
@@ -0,0 +1,379 @@
...
+                value={searchQuery}
+                onChange={setSearchQuery}
+                debounce={false}
+                autoFocus
+              />
+            </DropdownMenu.Header>
+          )}
+
+          {isNoItems && (
+            <DropdownMenu.NoOptions>{searchQuery ? 'No results found' : 'No options available'}</DropdownMenu.NoOptions>
```
3em, line 348: add translation

apps/design-system/src/pages/view-preview/view-settings.tsx
ankormoreankor: Although I have the right data types in options and value, ts finds type `number` somewhere. The select works well, so to skip this I defined the type
image

ankormoreankor: Fixed

praneshg239:
https://github.com/user-attachments/assets/204b3b6d-227e-41ea-9b50-e614fe3a9bd5

@ankormoreankor can you please check this? Border in warning theme is malfunctioning.

packages/ui/src/components/form-input/components/form-hoc.tsx
```diff
@@ -1,41 +1,43 @@
...
 type ControllerProps = ComponentProps<typeof Controller>
 type ControllerRenderPropsParamsType = Parameters<NonNullable<ControllerProps['render']>>[0]
+export type WithFormProps = {
+  name: string
+  error?: string
+}
 
 export function withForm<T>(
-  FormPrimitiveComponent: React.ForwardRefExoticComponent<T>,
+  FormPrimitiveComponent: ComponentType<T> | ForwardRefExoticComponent<T>,
```
praneshg239, line 12: Can we still use `ForwardRefExoticComponent` and avoid `ComponentType<T>`?

packages/ui/src/components/form-input/index.ts
```diff
@@ -17,7 +17,17 @@ const FormInput = {
     checked: field.value,
     onCheckedChange: field.onChange
   })),
+  Select: withForm(Select, ({ field }) => ({
```
praneshg239, line 20: Do we need these type casting? It makes the file look too complicated. Can we have `onSelectedChange` as `onChange`? So that it can be simplified to 

```js
Select: withForm(Select)
```

If not, we can have it like

```ts
Select: withForm(Select, ({ field }) => ({
  onSelectedChange: field.onChange
})),
```

packages/ui/src/components/form-input/index.ts
```diff
@@ -17,7 +17,17 @@ const FormInput = {
...
+    onSelectedChange: field.onChange
+  })) as <T = string>(
+    props: FormSelectProps<T> & { ref?: React.ForwardedRef<HTMLButtonElement> }
```
praneshg239, line 23: can we remove this type casting and make relevant changes in `Select` component? `Select` should still return `ForwardRefExoticComponent`

packages/ui/src/components/form-input/components/form-hoc.tsx
ankormoreankor: Nope🫤
ForwardRefExoticComponent doesn't allow to use Select component as a generic, so I faced issues with unresolved `options` type

packages/ui/src/components/form-input/index.ts
ankormoreankor: According to the [API](https://gist.github.com/abhinavrastogi-harness/984c6511de5f989d651cd7a7ef2e0c59), `onChange` has a different props interface, this is why I had to add `onSelectedChange`

packages/ui/src/components/form-primitives/select.tsx
```diff
@@ -0,0 +1,427 @@
...
+          valueOptions.push(subOption)
+        }
+      })
+    }
+  })
+
+  return valueOptions
+}
+
+function SelectInner<T = string>(
```
praneshg239, line 111: can we update it to have `forwardRef`? So that all those type castings can we easily removed.

```tsx
const SelectInner = forwardRef<HTMLButtonElement, SelectProps>(<T>(props, ref) => {
// ...

})

Comment (praneshg239) [reply packages/ui/src/components/form-primitives/select.tsx:111]: The purpose of implementing `forwardRef` is to enable future enhancements where we can expose component features through the `ref` using `useImperativeHandle`. For example, in a `Select` component, we could add an `onSelectChange` method to the `ref`, allowing consumer components to trigger selection changes programmatically.

Comment (ankormoreankor) [reply packages/ui/src/components/form-primitives/select.tsx:111]: The generics won't work this way. You can't place the generic `<T>` after the function parameters like that - it's invalid TypeScript syntax. Additionally, `forwardRef` doesn't support generic components directly. The generic type parameter needs to be on the outer component definition, not inside the `forwardRef` callback function.
That's why we need to use the type assertion approach at the bottom of this file

File: packages/ui/src/components/form-input/index.ts (hunk 7 above)
Comment (ankormoreankor) [reply packages/ui/src/components/form-input/index.ts]: Originally, I've added it for the integration with `react-hook-form`, because it passes `onChange` callback with a different props interface to the component. I tested if I can use the default Select's onChange, and it works. Don't know how and why, but anyways - I removed `onSelectedChange`🤷‍♂️

File: packages/ui/src/components/form-primitives/select.tsx [hunk 10]
```diff
@@ -0,0 +1,425 @@
...
```
praneshg239, line 110: Please add a backlog ticket to add `forwardRef` for `Select` component.

ankormoreankor: forwardRef is added on line 414

```js
const Select = forwardRef(SelectInner) as <T = string>(
  props: SelectProps<T> & { ref?: ForwardedRef<HTMLButtonElement> }
) => ReturnType<typeof SelectInner>
```

packages/ui/src/components/form-input/components/form-hoc.tsx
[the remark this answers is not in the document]
ankormoreankor: Done

packages/ui/src/components/form-input/index.ts
```diff
@@ -17,7 +17,12 @@ const FormInput = {
     checked: field.value,
     onCheckedChange: field.onChange
   })),
+  Select: withForm(Select) as <T = string>(
+    props: FormSelectProps<T> & { ref?: React.ForwardedRef<HTMLButtonElement> }
```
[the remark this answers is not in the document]
ankormoreankor: Done

PR #1716, opened 2025-06-03
simplify contribute dropdown to only show 'Compare' with updated message for opening a pull request

jake-bassett:
Both `Compare` and `Open pull request` went to the same URL. That screen is the launch point for opening a pull request, so removed `Open pull request` and updated the messaging to let users know that option is available from the "Compare" screen.

Before:
Old Contribute

After:
New Contribute


packages/ui/src/views/repo/components/branch-info-bar.tsx
```diff
@@ -78,22 +78,25 @@ export const BranchInfoBar: FC<BranchInfoBarProps> = ({
...
                 </span>
                 <StatusBadge className="mt-1" variant="secondary" theme="muted" size="sm">
                   <Icon name="branch" size={14} />
                   <span>{defaultBranchName}</span>
                 </StatusBadge>
                 .
-                <p className="mt-2.5 text-2 leading-tight text-cn-foreground-2">
-                  Open a pull request to contribute your changes upstream.
+                <p className="text-2 text-cn-foreground-2 mt-2.5 leading-tight">
+                  Compare branches to see your changes.
```
abhinavrastogi-harness, line 96: we should externalise this text to support i18n

abhinavrastogi-harness (approved):

PR #1717, opened 2025-06-04
Table component basic

sans-harness:
- Adds basic `table` component with variants and examples in documentation. 
- Adds tokens based on figma.


jake-bassett (commented):
Pulled this branch and tested it in my local. Looks good from my end, but I'll let others give a final approval on the code details.

apps/portal/src/content/docs/components/table-v2.mdx
```diff
@@ -0,0 +1,666 @@
...
+                    <Icon 
+                      name="chevron-down" 
+                      className={expanded ? "" : "transform -rotate-180"} 
+                    />
+                  </Button>
+                </TableV2.Cell> 
+                <TableV2.Cell>Jane Smith</TableV2.Cell>
+                <TableV2.Cell>jane@example.com</TableV2.Cell>
+                <TableV2.Cell>Jane</TableV2.Cell>
+                <TableV2.Cell>Not expandable</TableV2.Cell>
```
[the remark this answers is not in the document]
sans-harness: Removed `expansion` for now

apps/portal/src/content/docs/components/table-v2.mdx
```diff
@@ -0,0 +1,666 @@
...
+        <TableV2.Row>
+          <TableV2.Cell>Bob Johnson</TableV2.Cell>
+          <TableV2.Cell>bob@example.com</TableV2.Cell>
+          <TableV2.Cell>Bob</TableV2.Cell>
+          <TableV2.Cell>2023-03-20</TableV2.Cell>
+        </TableV2.Row>
+      </TableV2.Body>
+    </TableV2.Root>
+  </div>
+<Pagination
```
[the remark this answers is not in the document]
sans-harness: This is just representational - to showcase how can pagination would work with a `table`.

packages/ui/src/components/table-v2.tsx
```diff
@@ -0,0 +1,95 @@
...
+})
+
+export interface TableRootV2Props extends HTMLAttributes<HTMLTableElement>, VariantProps<typeof tableVariants> {
+  disableXScroll?: boolean
+  tableClassName?: string
+}
+
+const TableRoot = forwardRef<HTMLTableElement, TableRootV2Props>(
+  ({ variant, disableXScroll, className, tableClassName, ...props }, ref) => (
+    <div className={cn('cn-table-v2-container', tableVariants({ variant }), className)}>
```
abhinavrastogi-harness, line 26: why do we need this div wrapper? can we not apply the stlying directly on the `table` element?

packages/ui/src/components/table-v2.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <table
+        ref={ref}
+        className={cn('cn-table-v2-element', { 'cn-table-v2-disable-x-scroll': disableXScroll }, tableClassName)}
```
abhinavrastogi-harness, line 29: how does `disableXScroll` work? does the table have a built-in max-width after which it scrolls?

apps/portal/src/content/docs/components/table-v2.mdx
```diff
@@ -0,0 +1,590 @@
...
+
+### `Row`
+
+A row within the table.
+
+```typescript jsx
+<Table.Row
+  className="custom-class" // [OPTIONAL] Custom class name
+  expanded={false}       // [OPTIONAL] Whether the row is expanded
+  expandedContent={<TableV2.Cell colSpan={4}>Expanded content</TableV2.Cell>} // [OPTIONAL] Content to display when expanded
```
abhinavrastogi-harness, line 506: row doesn't these props anymore I think

apps/portal/src/content/docs/components/table-v2.mdx
```diff
@@ -0,0 +1,590 @@
...
+    },
+  ]}
+/>
+
+### `Row`
+
+A row within the table.
+
+```typescript jsx
+<Table.Row
```
abhinavrastogi-harness, line 503: don't we have support for a `selected` prop on the row? how is row selection highlight working right now?

apps/portal/src/content/docs/components/table-v2.mdx
```diff
@@ -0,0 +1,590 @@
...
+</Aside>
+
+### `Root`
+
+The root component for the table.
+
+```typescript jsx
+<TableV2.Root
+  variant="default"                 // [OPTIONAL] Variant of the table
+  disableXScroll={false}            // [OPTIONAL] Disable horizontal scroll
```
abhinavrastogi-harness, line 391: let's remove if not supported

apps/portal/src/content/docs/components/table-v2.mdx
```diff
@@ -0,0 +1,590 @@
...
+  ]}
+/>
+
+### `Body`
+
+The body section of the table.
+
+```typescript jsx
+<Table.Body
+  className="custom-class" // [OPTIONAL] Custom class name
```
abhinavrastogi-harness, line 458: missing `hasHighlightOnHover` prop here?

abhinavrastogi-harness: should we actually move the `highlight` prop to the `<Table.Root />` instead?
it doesn't seem very intuitive on the body.
also, I think we should enable highlight by default, and give option to disable it via a prop.

abhinavrastogi-harness: i'll update the api spec accordingly

sans-harness: sure, lets call it `disableHighlightOnHover` in that case.

abhinavrastogi-harness (approved):

PR #1728, opened 2025-06-07
Advanced data-table

sans-harness:


packages/ui/src/components/data-table.tsx
```diff
@@ -0,0 +1,260 @@
...
+  renderSubComponent,
+  enableColumnResizing = false,
+  getRowId
+}: DataTableProps<TData>) {
+  // Start with the base columns
+  let tableColumns = [...columns]
+
+  // If row selection is enabled, add a checkbox column at the beginning
+  if (enableRowSelection) {
+    tableColumns = [
```
abhinavrastogi-harness, line 103: same, also memoize

packages/ui/src/components/data-table.tsx
```diff
@@ -0,0 +1,260 @@
...
+        },
+        size: 20
+      },
+      ...tableColumns
+    ]
+  }
+
+  // If expanding is enabled, add an expander column at the beginning
+  if (enableExpanding) {
+    tableColumns = [
```
abhinavrastogi-harness, line 138: same, also memoize

packages/ui/src/components/data-table.tsx
```diff
@@ -0,0 +1,260 @@
...
+            </Button>
+          ) : null
+        },
+        size: 20
+      },
+      ...tableColumns
+    ]
+  }
+
+  const tableOptions: TableOptions<TData> = {
```
abhinavrastogi-harness, line 165: also memoize

packages/ui/src/components/data-table.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  className={cn(enableColumnResizing ? 'relative' : undefined)}
+                  sortable={header.column.getCanSort()}
+                  sortDirection={header.column.getCanSort() ? header.column.getIsSorted() || false : undefined}
+                  onClick={header.column.getCanSort() ? header.column.getToggleSortingHandler() : undefined}
+                  style={{
+                    width: header.getSize()
+                  }}
+                >
+                  {header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
+                  {enableColumnResizing && header.column.getCanResize() && (
```
abhinavrastogi-harness, line 220: are we keeping rezizing right now? maybe let's remove this until we have a good solution?

packages/ui/src/components/table-v2.tsx
```diff
@@ -97,8 +99,60 @@ const TableRow = forwardRef<HTMLTableRowElement, TableRowProps>(({ className, se
...
+   * Props for the tooltip component
+   */
+  tooltipProps?: Omit<TooltipProps, 'children'>
+}
+
+const TableHead = forwardRef<HTMLTableCellElement, TableHeadProps>(
+  ({ className, sortDirection, sortable, children, tooltipProps, ...props }, ref) => {
+    const childrenWithTooltip = tooltipProps?.content ? (
+      <Tooltip {...tooltipProps}>
+        <span className="border-b border-dashed">{children}</span>
```
abhinavrastogi-harness, line 121: why use a border? why not underline?

packages/ui/src/components/table-v2.tsx
```diff
@@ -97,8 +99,60 @@ const TableRow = forwardRef<HTMLTableRowElement, TableRowProps>(({ className, se
...
+      </Tooltip>
+    ) : (
+      children
+    )
+
+    const contentElement = (
+      <div className="flex items-center gap-1">
```
abhinavrastogi-harness, line 128: can we use `<Layout.Horizontal>` instead?

packages/ui/src/components/data-table.tsx
sans-harness: I've marked the prop with the `@internal` TSDoc tag, so it's not accessible outside the package. Internally, we can continue iterating on it as needed. Do you think we should remove support for it entirely instead?

abhinavrastogi-harness (approved):

PR #1730, opened 2025-06-09
Toggle design migration

3em:
- migrated component to the [new design](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=8776-7418&t=WDTU8n0n81j04PoB-4)
- documented Toggle usage with Portal

image

### DIFF
--- a/packages/ui/src/components/toggle.tsx
+++ b/packages/ui/src/components/toggle.tsx
@@ -23,7 +23,7 @@ export interface ToggleProps extends Pick<ButtonProps, 'rounded' | 'iconOnly'> {
   tooltipProps?: ToggleTooltipProps
 }
 
-const Wrapper: FC<{ children: ReactNode; tooltipProps?: ToggleTooltipProps }> = ({ children, tooltipProps }) =>
+const TooltipWrapper: FC<{ children: ReactNode; tooltipProps?: ToggleTooltipProps }> = ({ children, tooltipProps }) =>
   tooltipProps ? <Tooltip {...tooltipProps}>{children}</Tooltip> : <>{children}</>
 
 const Toggle = forwardRef<
@@ -65,7 +65,7 @@ const Toggle = forwardRef<
     )
 
     return (
-      <Wrapper tooltipProps={tooltipProps}>
+      <TooltipWrapper tooltipProps={tooltipProps}>
         <TogglePrimitive.Root ref={ref} asChild pressed={pressed} onPressedChange={onChange} disabled={disabled}>
           <Button
             className={className}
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/toggle.tsx
```diff
@@ -1,49 +1,87 @@
...
+    const [pressed, setPressed] = useState(pressedProp)
+
+    const buttonVariant = useMemo(() => {
+      const [inactiveVariant, activeVariant] = variant.split('-') as [ButtonVariants, ButtonVariants]
+      return pressed ? activeVariant : inactiveVariant
+    }, [variant, pressed])
+
+    useEffect(() => {
+      setPressed(pressedProp)
+    }, [pressedProp])
```
abhinavrastogi-harness, line 57: Looks like we are mixing the controlled and uncontrolled behaviour of the component here?
I think we should have two separate props:
defaultPressed: used to just initialise the state, and used for an uncontrolled component.
pressed: used for a controlled component, and we don't need to maintain internal state at all.

packages/ui/src/components/toggle.tsx
```diff
@@ -1,49 +1,87 @@
...
+type ToggleVariant = 'outline-primary' | 'outline-secondary' | 'ghost-primary' | 'ghost-secondary'
+type ToggleTooltipProps = Pick<TooltipProps, 'title' | 'content' | 'side' | 'align'>
+
+export interface ToggleProps extends Pick<ButtonProps, 'rounded' | 'iconOnly'> {
+  variant?: ToggleVariant
+  size?: ButtonSizes
+  tooltipProps?: ToggleTooltipProps
+}
+
+const Wrapper: FC<{ children: ReactNode; tooltipProps?: ToggleTooltipProps }> = ({ children, tooltipProps }) =>
```
abhinavrastogi-harness, line 26: `TooltipWrapper` might be a better name, as it'll help in debugging the component tree in dev tools.

packages/ui/src/components/toggle.tsx
3em: The reason why the `Toggle` component itself contains a `pressed` state and effectively always becomes a controlled component is because we need to pass different values to the `variant` prop of the `Button` component depending on the pressed state. 
The Radix primitive itself doesn't provide a way to control the state other than through a `data-attr`, so we had to manage it via state.

Because of this, there is no need for the `defaultPressed` prop, as it is fully replaced by the `pressed` prop — whether we want to pass a default value or control the state externally.

abhinavrastogi-harness: I still feel this can be simplified, but let's go ahead with the current implementation.

PR #1737, opened 2025-06-10
Download and migrate logos from Figma using script (LogoV2)

praneshg239:
- Deleted Logo.tsx :tada:, please use LogoV2 for logos 
  - Renaming changes regarding LogoV2 to Logo is in logov2-component-to-logo branch. Still needs some changes and testing. It can be merged later. Not a blocker.
- Added standard size for LogoV2 

image

image


packages/ui/scripts/logos.js
```diff
@@ -0,0 +1,501 @@
...
+              // Add wrapper group for perfect centering
+              // This uses a combination of techniques to ensure centering works across different SVG structures
+              const centeringGroup = {
+                type: 'element',
+                name: 'g',
+                attributes: {
+                  // viewport is 18x18, hence added translate(9, 9) to center the logo
+                  // scale(0.8) is used to scale the logo to 80% of its original size
+                  // translate(-9, -9) is used to center the logo after scaling
+                  transform: 'translate(9, 9) scale(0.8) translate(-9, -9)'
```
praneshg239, line 217: This is to center the icon based on the viewport.

apps/portal/src/content/docs/components/logo.mdx
```diff
@@ -2,72 +2,30 @@
 title: Logo
 description: Logo component
 beta: true
-draft: true
 ---
 
 The `Logo` component renders brand icons from the `simple-icons` package.
```
abhinavrastogi-harness, line 7: I don't think we are directly using `simple-icons` anymore, so this line is incorrect. let's remove it.

packages/ui/src/components/logo-v2/logo-v2.tsx
abhinavrastogi-harness, line 1: it looks like we're hardcoding the background color in the svg itself.
can we consider to do that in css as a background instead?

abhinavrastogi-harness (approved):

apps/portal/src/content/docs/components/logo.mdx
praneshg239: Sure. Portal will be updated entirely 👍

packages/ui/src/components/logo-v2/logo-v2.tsx
praneshg239: In this approach, we can encapsulate all the logo related styles inside SVG itself. If we have to update the background color based on the requirement, then we can do that.

In that case, we have to generate a css file through the same script, specific for Logo and have to add it the root styles and relevant class name needs to be added to the bg `rect` element.

abhinavrastogi-harness (approved):

PR #1747, opened 2025-06-11
feat: add checkbox support for entity-reference

sans-harness:


packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -91,21 +124,29 @@ export function EntityReference<T extends BaseEntityProps, S = string, F = strin
...
       <StackedList.Item
         onClick={() => onSelect?.(entity)}
-        className={cn({ 'bg-cn-background-hover': isSelected })}
-        thumbnail={<IconV2 name="page" className="text-cn-foreground-3" />}
+        className={cn('h-12 p-3', { 'bg-cn-background-hover': isSelected })}
+        thumbnail={
+          showCheckbox ? (
+            <Checkbox checked={isSelected} onCheckedChange={() => onSelect?.(entity)} />
+          ) : (
+            <IconV2 name={defaultRendererIcon ?? 'page'} className="text-cn-foreground-3" />
```
sans-harness, line 136: rename to icon

packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -14,23 +14,24 @@ import {
...
   // Callbacks
-  onSelectEntity: (entity: T) => void
   onScopeChange: (direction: DirectionEnum) => void
   onFilterChange?: (filter: string) => void
 
   // UI Configuration
   showFilter?: boolean
   showBreadcrumbEllipsis?: boolean
   filterTypes?: Record<string, string>
+  enableMultiSelect?: boolean
```
abhinavrastogi-harness, line 34: we can probably remove it from the common props?

packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -14,23 +14,24 @@ import {
   ParentFolderRendererProps
 } from './types'
 
-export interface EntityReferenceProps<T extends BaseEntityProps, S = string, F = string> {
+export interface CommonEntityReferenceProps<T extends BaseEntityProps, S = string, F = string> {
   // Data
   entities: T[]
-  selectedEntity: T | null
+  selectedEntity?: T | null
+  selectedEntities?: T[]
```
abhinavrastogi-harness, line 21: these can be moved into specific single/multi prop overrides below I think.

packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -79,9 +104,17 @@ export function EntityReference<T extends BaseEntityProps, S = string, F = strin
   })
   const handleSelectEntity = useCallback(
     (entity: T) => {
-      onSelectEntity?.(entity)
+      if (enableMultiSelect) {
+        const isEntitySelected = selectedEntities.some(item => item.id === entity.id)
```
abhinavrastogi-harness, line 108: entity ids may not be globally unique. they would only be accurate within their specific folder.
we would need to either add that condition (to match folder path), and maybe also allow the user to pass a custom predicate.

packages/ui/src/views/platform/entity-reference-component.tsx
sans-harness: If we do this, `enableMultiselect` stops being a valid `discriminator` in our union type. If we want this we will need to change the props and do a larger refactor across files. IMHO - its not worth it.

packages/ui/src/views/platform/entity-reference-component.tsx
sans-harness: I don't think there's a scalable way for us to match `folderpaths`. I have added support for users to pass in their custom comparator which they can use in case the default doesn't work, but the default for now should just compare ID's.

packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -42,12 +42,35 @@ export interface EntityReferenceProps<T extends BaseEntityProps, S = string, F =
   // Search
   searchValue?: string
   handleChangeSearchValue: (val: string) => void
+
+  // Icons for default entity renderer
+  icon?: IconPropsV2['name']
+
+  // Custom entity comparison
+  isEntityEqual?: (entity1: T, entity2: T) => boolean
```
abhinavrastogi-harness, line 50: let's call this `compareFn` to be consistent with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#comparefn

packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -61,27 +84,48 @@ export function EntityReference<T extends BaseEntityProps, S = string, F = strin
...
   const { search, handleSearchChange } = useDebounceSearch({
     handleChangeSearchValue,
     searchValue
   })
   const handleSelectEntity = useCallback(
     (entity: T) => {
-      onSelectEntity?.(entity)
+      if (enableMultiSelect) {
+        const defaultIsEqual = (item: T, entity: T) => {
+          return item.id === entity.id && (currentFolder === null || currentFolder === item.folderPath)
```
abhinavrastogi-harness, line 114: I think we need to compare `entity.folderPath`, not `currentFolder`

abhinavrastogi-harness: can be a pure function and moved out of the render path

packages/ui/src/views/platform/entity-reference-component.tsx
```diff
@@ -42,12 +42,35 @@ export interface EntityReferenceProps<T extends BaseEntityProps, S = string, F =
...
+  isEntityEqual?: (entity1: T, entity2: T) => boolean
+}
+
+export interface SingleSelectEntityReferenceProps<T extends BaseEntityProps, S = string, F = string>
+  extends CommonEntityReferenceProps<T, S, F> {
+  enableMultiSelect?: false
+  onSelectEntity: (entity: T) => void
+}
+
+export interface MultiSelectEntityReferenceProps<T extends BaseEntityProps, S = string, F = string>
```
abhinavrastogi-harness, line 59: can we try doing something like `MultiSelectBaseEntityProps` which makes `folderPath` required? because we don't need it for single select.
and this way, we can completely skip the custom comparator support.

sans-harness: As discussed, lets take this up as a follow up.

abhinavrastogi-harness (approved):

PR #1752, opened 2025-06-12
lineClamp prop to Text component and update default align

pixel-buba:
- Added lineClamp prop with possible values: '1' | '2' | '3' | '4' | '5' | '6' | 'default'
 - lineClamp now takes precedence over truncate
 - Removed the default value for align; it's now 'initial' instead of 'left'

### DIFF
--- a/packages/ui/src/components/text.tsx
+++ b/packages/ui/src/components/text.tsx
@@ -72,13 +72,13 @@ const textVariants = cva('', {
       danger: 'text-cn-foreground-danger'
     },
     lineClamp: {
+      default: '',
       1: 'line-clamp-1',
       2: 'line-clamp-2',
       3: 'line-clamp-3',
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/text.tsx
```diff
@@ -71,6 +71,15 @@ const textVariants = cva('', {
...
       danger: 'text-cn-foreground-danger'
     },
+    lineClamp: {
+      1: 'line-clamp-1',
+      2: 'line-clamp-2',
+      3: 'line-clamp-3',
+      4: 'line-clamp-4',
+      5: 'line-clamp-5',
+      6: 'line-clamp-6',
+      none: 'line-clamp-none'
```
3em, line 81: We can change to `default` and not use the `line-clamp-none` class — just an empty string instead.

PR #1754, opened 2025-06-12
Add suffix to Number Input

vivek-harness:
This PR adds suffix to the `NumberInput` component since we need it for multi-type inputs.

https://github.com/user-attachments/assets/34103bfa-e260-469b-a698-227df4173ae5

### DIFF
--- a/packages/ui/src/components/inputs/number-input.tsx
+++ b/packages/ui/src/components/inputs/number-input.tsx
@@ -4,7 +4,7 @@ import { Button, ControlGroup, FormCaption, IconV2, Label } from '@/components'
 
 import { BaseInput, InputProps } from './base-input'
 
-export interface NumberInputProps extends Omit<InputProps, 'type' | 'suffix'> {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/inputs/number-input.tsx
```diff
@@ -113,30 +114,33 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
...
+                    variant="ghost"
+                    iconOnly
+                    onClick={handleDecrement}
+                    disabled={disabled}
+                    size="sm"
+                  >
+                    <IconV2 name="nav-arrow-down" size={14} />
+                  </Button>
+                </div>
+                <div className="border-inherit">{suffix}</div>
```
praneshg239, line 143: Can we render suffix conditionally?

```suggestion
                {suffix && <div className="border-inherit">{suffix}</div>}
```

vivek-harness: Done

PR #1764, opened 2025-06-16
Icon size update

praneshg239:
image


jake-bassett (approved):

praneshg239:
> the large stroke looks too heavy.. i see a slack convo on it, but we will need to fix it, maybe in a follow up

👍

PR #1775, opened 2025-06-17
add popover design migration

ankormoreankor:
##### Popover component design migration

[Design](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=8818-24160&m=dev)

- migrated `Popover` component to the new design
- added new props to popover content: `title`, `description`, `hideArrow`, `linkProps`
- added docs for the component
- the popover component can be used via the Composition or Builder API pattern

image


jake-bassett:
Reviewing the demo page, in the `Popover with Custom Content` section, the content within the popover feels a bit off. I understand that content really isn't the point of the example, so not a huge priority, but if we have time we should clean that up a bit.

packages/ui/src/components/illustration/illustration.tsx
```diff
@@ -31,6 +31,11 @@ const Illustration: FC<IllustrationProps> = ({
       ? IllustrationsNameMap[`${name}-light` as keyof typeof IllustrationsNameMap]
       : IllustrationsNameMap[name]
 
+  if (!Component) {
+    console.warn(`Icon "${name}" not found in IllustrationsNameMap.`)
+    return null
+  }
+
```
vardanbansal-harness, line 38: Shouldn't this be caught by `typescript` itself? If there's no component mapped against a `name`, it should allow `Illustration` component to be rendered with that `name` itself.

packages/ui/src/views/unified-pipeline-studio/components/unified-pipeline-studio-footer.tsx
```diff
@@ -94,13 +94,13 @@ export const UnifiedPipelineStudioFooter: React.FC<PipelineStudioFooterProps> =
       {committedTimeAgo && authorName && (
         <Popover.Root>
           <Popover.Trigger>
-            <div className="flex text-[12px] text-cn-foreground-3">
+            <div className="text-cn-foreground-3 flex text-[12px]">
```
vardanbansal-harness, line 97: Same

packages/ui/src/components/illustration/illustration.tsx
vardanbansal-harness: Not a blocker though, something that you can discuss separately with @praneshg239 once he's back.

PR #1790, opened 2025-06-20
fix: migrate text tags to Text component

pixel-buba:
Migrate text tags to Text component:
 - heading tags (h1, h2, etc)
 - p
 - span 
 - label
 - div

### DIFF
--- a/packages/ui/src/components/form.tsx
+++ b/packages/ui/src/components/form.tsx
@@ -11,7 +11,6 @@ import {
   type UseFormReturn
 } from 'react-hook-form'
 
-import { Text } from '@components/text'
 import { Slot } from '@radix-ui/react-slot'
 import { cn } from '@utils/cn'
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/views/execution/execution-header.tsx
```diff
@@ -45,13 +45,13 @@ export const ExecutionHeader: FC<ExecutionHeaderProps> = ({
   return (
     <div className={cn('px-6 py-5', className)}>
       <div className="flex flex-col gap-[18px]">
-        <h1 className="max-w-[600px] text-4 font-medium leading-snug">
+        <Text as="h1" variant="heading-subsection" className="max-w-[600px]">
           {title.number && <span className="text-cn-foreground-2">#{title.number} </span>}
```
3em, line 49: let's change to Text

packages/ui/src/views/execution/execution-header.tsx
```diff
@@ -45,13 +45,13 @@ export const ExecutionHeader: FC<ExecutionHeaderProps> = ({
...
           <span className="text-cn-foreground-1">{title.title}</span>
```
3em, line 50: let's change to Text

PR #1799, opened 2025-06-24
Sidebar design migration

ankormoreankor:
In progress

[Design](https://www.figma.com/design/AIgjIyUzcuZzuVnuoOyhQE/Harness-Design-System--3.0-?node-id=9800-45551&m=dev) * [Discussion](https://pixelpointworkspace.slack.com/archives/C088CNHL5L3/p1750754483572939)

- migrated the Sidebar to the new design system
- refactored the implementations
- removed outdated tokens
- updated the component docs
- added new component Shortcut
- removed legacy sidebar search
- added missing portal wrapper to the Tooltip component

image


packages/ui/src/components/app-sidebar/sidebar-search/sidebar-search.tsx
```diff
@@ -22,36 +21,27 @@ export function SidebarSearch(props: SidebarSearchProps) {
...
-          type="button"
-          className="absolute right-1.5 top-1/2 z-[5px] h-5 -translate-y-1/2 select-none rounded-sm border p-0 px-1.5 opacity-100 transition-opacity group-data-[state=collapsed]:opacity-0"
+          placeholder={t('component:navbar.search', 'Search')}
+          className="pointer-events-none"
+          inputContainerClassName="[&>.cn-input-prefix]:w-[34px] max-w-full overflow-hidden"
+          suffix={
+            <Text
+              variant="caption-soft"
+              className="text-cn-foreground-2 border-cn-borders-2 mr-1.5 flex h-5 select-none items-center rounded-sm border px-1 opacity-100 transition-opacity group-data-[state=collapsed]:opacity-0"
+            >
```
3em, line 36: Let’s create a `ShortcutButton` component and use it here. We’ll also need to bring the token styles over from Figma.

image

packages/ui/src/components/app-sidebar/sidebar-search/sidebar-search.tsx
```diff
@@ -22,36 +22,20 @@ export function SidebarSearch(props: SidebarSearchProps) {
   const { setIsOpen } = searchContext
 
   return (
-    <div>
+    <Layout.Grid gapY="md">
       {props.logo}
```
praneshg239, line 26: Can we move logo out of search component to sidebar header?

packages/ui/src/components/sidebar/sidebar-context.tsx
```diff
@@ -0,0 +1,111 @@
...
+    (value: boolean | ((value: boolean) => boolean)) => {
+      const openState = typeof value === 'function' ? value(open) : value
+      if (setOpenProp) {
+        setOpenProp(openState)
+      } else {
+        _setOpen(openState)
+      }
+
+      // This sets the cookie to keep the sidebar state.
+      document.cookie = `${SIDEBAR_COOKIE_NAME}=${openState}; path=/; max-age=${SIDEBAR_COOKIE_MAX_AGE}`
```
praneshg239, line 57: What are the benefits of storing the status in cookie than localStorage? Where are we reading its value to persist the sidebar state?

packages/ui/src/components/sidebar/sidebar-context.tsx
```diff
@@ -0,0 +1,111 @@
...
+        toggleSidebar()
+      }
+    }
+
+    window.addEventListener('keydown', handleKeyDown)
+    return () => window.removeEventListener('keydown', handleKeyDown)
+  }, [toggleSidebar])
+
+  // We add a state so that we can do data-state="expanded" or "collapsed".
+  // This makes it easier to style the sidebar with Tailwind classes.
```
praneshg239, line 81: 🔥

packages/ui/src/components/sidebar/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  Header: SidebarHeader,
+  Inset: SidebarInset,
+  Item: SidebarItem,
+  MenuSkeleton: SidebarMenuSkeleton,
+  MenuSubItem: SidebarMenuSubItem,
+  Provider: SidebarProvider,
+  Rail: SidebarRail,
+  Separator: SidebarSeparator,
+  Trigger: SidebarTrigger,
+  ToggleMenuButton: SidebarToggleMenuButton
```
praneshg239, line 31: Just checking, can we add `SidebarSearch` also here?

packages/ui/tailwind-design-system.ts
```diff
@@ -293,39 +295,6 @@ export default {
             lime: 'var(--canary-label-background-lime-01)'
           }
         },
-        sidebar: {
```
praneshg239, line 296: 🔥

packages/ui/src/components/sidebar/sidebar-context.tsx
ankormoreankor: This thing has been there since the sidebar was set up for the first time. It probably was generated by the Shadcn. Currently, it's not used anywhere

image

packages/ui/src/components/sidebar/sidebar-context.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  // Helper to toggle the sidebar.
+  const toggleSidebar = useCallback(() => {
+    return isMobile ? setOpenMobile(open => !open) : setOpen(open => !open)
+  }, [isMobile, setOpen, setOpenMobile])
+
+  // Adds a keyboard shortcut to toggle the sidebar.
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === SIDEBAR_KEYBOARD_SHORTCUT && (event.metaKey || event.ctrlKey)) {
```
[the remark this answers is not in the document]
ankormoreankor: This thing has also been there since the sidebar was set up for the first time. I'm not sure if we even need it. I think, for the shortcuts, it makes sense to use some comprehensive and more reliable solution than just listeners here and there. From my perspective, it would make sense to even drop it for now. What do you think @praneshg239 ?

packages/ui/src/components/sidebar/sidebar-item.tsx
```diff
@@ -0,0 +1,402 @@
...
+type SidebarItemExtendedProps = SidebarItemIconProps | SidebarItemLogoProps | SidebarItemAvatarProps
+
+type SidebarItemButtonProps = SidebarItemExtendedProps &
+  Omit<ComponentPropsWithoutRef<'button'>, keyof SidebarItemExtendedProps>
+
+type SidebarItemLinkProps = SidebarItemExtendedProps & NavLinkProps
+
+export type SidebarItemProps = SidebarItemButtonProps | SidebarItemLinkProps
+
+export interface SidebarComponent {
```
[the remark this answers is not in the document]
ankormoreankor: I would argue here. Most of the types described in this file are **exclusively** used in the `SidebarItemTrigger` component and should not be exposed. The component itself is already split into a few smaller components for better readability. So basically, the file contains things that make sense when used together, so I believe they should stay together. The rest Sidebar components and utilities are already moved to separate files.

packages/ui/src/components/sidebar/sidebar-item.tsx
```diff
@@ -0,0 +1,402 @@
...
+  displayName?: string
+}
+
+const itemIsIcon = (item: SidebarItemProps): item is SidebarItemIconProps => !!(item as SidebarItemIconProps).icon
+const itemIsLogo = (item: SidebarItemProps): item is SidebarItemLogoProps => !!(item as SidebarItemLogoProps).logo
+const itemIsAvatar = (item: SidebarItemProps): item is SidebarItemAvatarProps =>
+  !!(item as SidebarItemAvatarProps).src || !!(item as SidebarItemAvatarProps).avatarFallback
+const isSidebarItemLink = (item: SidebarItemProps): item is SidebarItemLinkProps =>
+  'to' in item && (item as SidebarItemLinkProps).to !== undefined
+const isBadgeProps = (props?: string | SidebarBadgeProps): props is SidebarBadgeProps =>
```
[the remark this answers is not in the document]
ankormoreankor: Here is the same logic as in the [comment above](https://github.com/harness/canary/pull/1799#discussion_r2200092132): these are very component-specific utils that shouldn't be exposed. Splitting this file will most likely increase the cognitive load, and reduce it.

packages/ui/src/components/sidebar/index.ts
ankormoreankor: `SidebarSearch` is a part of the app-sidebar component group, which is built above the Sidebar component group. Sidebar component group is a group of primitives, and `SidebarSearch` is a much smarter component; it comes with its own context and does some side effects.

image

apps/gitness/src/components-v2/mfe/side-bar.tsx
```diff
@@ -84,73 +48,55 @@ const AppSidebar: FC<{ children: React.ReactNode }> = ({ children }) => {
...
-            <User
+
+          <Sidebar.Footer>
+            <Sidebar.ToggleMenuButton />
+            <Sidebar.Separator />
+            <AppSidebarUser
               user={currentUser}
-              openThemeDialog={() => setOpenThemeDialog(true)}
-              openLanguageDialog={() => setOpenLanguageDialog(true)}
+              openThemeDialog={noop}
```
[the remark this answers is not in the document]
ankormoreankor: Fixed

packages/ui/src/components/app-sidebar/sidebar-search/sidebar-search.tsx
ankormoreankor: Fixed

packages/ui/src/components/sidebar/sidebar-context.tsx
praneshg239: Got it 👍 ... can we add a logic to persist sidebar state of user using localStorage? You can take it as a separate PR.

packages/ui/src/components/sidebar/sidebar-context.tsx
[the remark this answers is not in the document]
praneshg239: Don't remove the logic for it, but comment it and add a TODO comment above. We can discuss with other folks and decide on this. Works?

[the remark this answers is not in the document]
ankormoreankor: Sure👍

packages/ui/src/components/app-sidebar/sidebar-search/sidebar-search.tsx
```diff
@@ -22,20 +16,17 @@ export function SidebarSearch(props: SidebarSearchProps) {
...
-          className="pointer-events-none"
-          inputContainerClassName="border-cn-borders-2 [&>.cn-input-prefix]:w-[34px] max-w-full overflow-hidden"
-          suffix={<Shortcut className="mr-1.5 transition-opacity group-data-[state=collapsed]:opacity-0">⌘K</Shortcut>}
-          readOnly
-          aria-hidden="true"
-          tabIndex={-1}
-        />
-      </button>
-    </Layout.Grid>
+    <button className="max-w-full overflow-hidden" onClick={() => setIsOpen(true)}>
```
praneshg239, line 19: do we still need this button here? Can we use `SearchInput` alone and add `onClick`. Thoughts?

packages/ui/src/components/sidebar/sidebar-item.tsx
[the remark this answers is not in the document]
ankormoreankor: If you need to debug a component, switching between multiple files can be exhausting, but in most cases, you don't even have to jump through types when you debug logic or styles.

If to move types to a separate file and export only `SidebarItemProps` and `SidebarComponent`, this won't be enough, because the type guards like `itemIsIcon` require some other types. So you either have to expose them too, or move type guards to the same `types.ts` file and expose them instead. In any case, we end up with having more specific things exposed than before. This is why I prefer to keep things together if they are only used together.

praneshg239 (approved):

PR #1800, opened 2025-06-24
Design system updates

vivek-harness:
This PR takes care of the following:
- Fix stroke width issue in icons
- Update avatar image in the docs

Icon stroke width issue

Avatar image update in docs


packages/ui/tailwind-utils-config/components/icon-and-logo.ts
```diff
@@ -16,10 +16,6 @@ function createIconandLogoSizeStyles(entity: 'icon' | 'logo') {
     style[`height`] = `var(--cn-icon-size-${size})`
     style[`min-height`] = `var(--cn-icon-size-${size})`
 
-    if (entity === 'icon') {
-      style['stroke-width'] = `var(--cn-icon-stroke-width-${size})`
```
vivek-harness, line 20: We don't need to apply different stroke-width for different sizes since we are scaling the same svg where the viewBox and path are the same, just the width and height is different, the stroke-width scales automatically.

PR #1810, opened 2025-06-25
Fixes to draggable card component

sans-harness:


packages/ui/src/components/draggable-card.tsx
```diff
@@ -63,17 +74,27 @@ export const DraggableCardList = ({
...
 
+  const sensors = useSensors(
+    useSensor(PointerSensor, {
+      // Add a small delay to allow clicks to be processed before starting drag
+      activationConstraint: {
+        delay: 100,
+        tolerance: 0
+      }
+    })
+  )
```
vardanbansal-harness, line 85: What happens if we don't add this?

PR #1823, opened 2025-06-28
Secrets fixes

sans-harness:


packages/ui/src/views/platform/components/entity-reference-filter.tsx
```diff
@@ -25,9 +25,11 @@ export const EntityReferenceFilter: React.FC<EntityReferenceFilterProps> = ({
 
   return (
     <DropdownMenu.Root>
-      <DropdownMenu.Trigger className="flex h-8 w-36 items-center justify-between rounded-md bg-cn-background-3 px-3 py-2 text-sm hover:bg-cn-background-hover">
-        <span className="truncate text-cn-foreground-1">{filterTypes[selectedType]}</span>
-        <IconV2 name="nav-arrow-down" size="2xs" className="chevron-down" />
+      <DropdownMenu.Trigger>
+        <Button variant="outline" size="sm" className="flex items-center justify-between h-8 w-36 !px-2">
```
vardanbansal-harness, line 29: Why is the padding override needed?

sans-harness: Button's internal padding causes text inside trigger to truncate. `px-2` stops that from happening.

abhinavrastogi-harness (approved):

packages/ui/src/views/platform/components/entity-reference-filter.tsx
abhinavrastogi-harness: can we use a Select instead of a custom dropdown?

PR #1838, opened 2025-07-01
fix: minor bug fixes

sans-harness:
https://github.com/user-attachments/assets/34fc5b2f-514d-449a-8823-0f940902d1f4


apps/gitness/src/components/FileExplorer.tsx
```diff
@@ -234,9 +238,13 @@ export default function Explorer({ selectedBranch, repoDetails }: ExplorerProps)
       value={openFolderPaths}
     >
       {isRootLoading ? (
-        <div>Loading...</div>
+        <div>
```
praneshg239, line 241: do we need this `div` here?

praneshg239 (approved):

PR #1849, opened 2025-07-03
fix: updated drawer to handle nested drawers without specifying manually

harish-viswa:


packages/ui/src/components/drawer/drawer-context.tsx
```diff
@@ -4,18 +4,15 @@ import { Drawer as DrawerPrimitive } from 'vaul'
...
 
 export const useDrawerContext = () => {
-  const ctx = useContext(DrawerContext)
-
-  if (!ctx) {
-    throw new Error('useDrawerContext must be used within <Drawer.Root>')
-  }
-
-  return ctx
+  return useContext(DrawerContext)
```
vardanbansal-harness, line 17: We can leave this around so that this hook is always called in a component wrapper in Drawer Provider.

vardanbansal-harness (approved):

PR #1854, opened 2025-07-07
Fix multiple minor issues

sans-harness:
1. Fix breadcrumbs on `profile-settings` page
2. Add truncation to title in `entity-ref`
3. Remove periods from cards


apps/gitness/src/routes.tsx
```diff
@@ -1083,11 +1083,11 @@ export const routes: CustomRouteObject[] = [
         element: <ProfileSettingsLayout />,
         handle: {
           breadcrumb: () => (
-            <>
+            <div className="flex items-center gap-1">
```
abhinavrastogi-harness, line 1086: can we use `<Layout.Horizontal />` instead?

PR #1868, opened 2025-07-10
Add Horizontal layout to Form Inputs

3em:
- Added `orientation` support to `TextInput`, `NumberInput`, `Textarea`, `Select`, `MultiSelect`, and `RadioGroup`.
- Refactored `MultiSelect`: added support for label, caption, error, and warning; removed this logic from `FormMultiSelect`.
- Refactored the `Radio` component.
- Refactored `ControlGroup` to support horizontal orientation.
- Passed the `orientation` prop from `FormWrapper` down to all `FormInput` components.
- Updated and fixed documentation for all affected components.

image

image

image

image


packages/ui/tailwind-utils-config/components/form-shared-styles.ts
```diff
@@ -6,6 +6,28 @@ export default {
...
     gap: 'var(--cn-input-wrapper-gap)',
     '@apply relative flex flex-col': '',
-    maxWidth: '100%'
+    maxWidth: '100%',
+
+    '&.cn-control-group-horizontal': {
+      '@apply flex-row': '',
+
+      '.cn-control-group-label': {
+        width: 'var(--cn-input-horizontal-label-max-width)',
```
praneshg239, line 15: If we use it inside a small container whose width is lesser than this, it will scroll. Is this expected?

3em: The label wrapper always has a fixed width, while the input wrapper takes up the remaining width of the container. If the overall container is less than 400px wide, this horizontal form layout should not be used at all.

packages/ui/src/components/form-input/index.ts
```diff
@@ -18,9 +16,9 @@ const FormInput = {
     onCheckedChange: field.onChange
   })),
   Select: withForm(Select) as <T = string>(
-    props: FormSelectProps<T> & { ref?: React.ForwardedRef<HTMLButtonElement> }
-  ) => React.ReactElement,
-  MultiSelect: FormMultiSelect
+    props: FormSelectProps<T> & { ref?: ForwardedRef<HTMLButtonElement> }
+  ) => ReactElement,
+  MultiSelect: withForm(MultiSelect)
```
praneshg239, line 21: 🔥

packages/ui/src/components/form-input/index.ts
```diff
@@ -18,9 +16,9 @@ const FormInput = {
     onCheckedChange: field.onChange
   })),
   Select: withForm(Select) as <T = string>(
-    props: FormSelectProps<T> & { ref?: React.ForwardedRef<HTMLButtonElement> }
-  ) => React.ReactElement,
-  MultiSelect: FormMultiSelect
+    props: FormSelectProps<T> & { ref?: ForwardedRef<HTMLButtonElement> }
```
praneshg239, line 19: This needs a clean up too. We can pick it later. Its only the typecasting which we need to remove.

praneshg239 (approved):

PR #1884, opened 2025-07-14
fix: default options from options/items/enum and connectorTypes from oneOf

shaurya-harness:
fix: default options from options/items/enum and connectorTypes from oneOf


packages/ui/src/views/run-pipeline/types.ts
```diff
@@ -51,6 +51,8 @@ export interface PipelineInput {
    * Label defines the input label.
    */
   label?: string
+
+  oneof?: any[]
```
jake-bassett, line 55: Can we use `unknown` instead of `any` here?

srdjan-harness (approved):

PR #1887, opened 2025-07-14
More PR comment box improvements

jake-bassett:
This adds the ability to select multiple lines and define a list. Also fixes a few selection related bugs.

### DIFF
--- a/packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-comment-box.tsx
+++ b/packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-comment-box.tsx
@@ -175,7 +175,6 @@ export const PullRequestCommentBox = ({
   useEffect(() => {
     if (textAreaRef.current) {
       textAreaRef.current.setSelectionRange(textSelection.start, textSelection.end)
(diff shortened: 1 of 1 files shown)


packages/ui/src/views/repo/pull-request/details/components/conversation/pull-request-comment-box.tsx
```diff
@@ -164,6 +175,7 @@ export const PullRequestCommentBox = ({
   useEffect(() => {
     if (textAreaRef.current) {
       textAreaRef.current.setSelectionRange(textSelection.start, textSelection.end)
+      textAreaRef.current.focus()
```
praneshg239, line 178: @jake-bassett This will bring back autofocus to the comment box when the page loads, the comment box will be focused by default, causing the page to scroll to that section.

Can we update the logic to focus only when it's not the initial mount of the component?

jake-bassett: Ah, good call out. Updated to only set focus after action click.

praneshg239 (approved):

PR #1893, opened 2025-07-15
Refactor UI Components to support Refs

pixel-buba:


packages/ui/src/components/progress.tsx
```diff
@@ -58,110 +58,118 @@ interface IndeterminateProgressProps extends CommonProgressProps {
...
+        <div className="cn-progress-container">{getProgress()}</div>
 
-      {(description || subtitle) && (
-        <div className="cn-progress-footer">
-          <div className="cn-progress-description-wrap">
-            {description && (
-              <Text className="cn-progress-description" variant="body-strong" color="foreground-3" truncate>
-                {description}
+        {(description || subtitle) && (
+          <div className="cn-progress-footer">
```
praneshg239, line 154: can we add design system text component?

packages/ui/src/components/logo-v2/logo-v2.tsx
```diff
@@ -37,24 +37,26 @@ interface LogoFallbackPropsV2 extends BaseLogoPropsV2 {
...
-    return null
-  }
+    if (!Component) {
+      console.warn(`Logo "${name}" not found in LogoNameMapV2.`)
+      return null
+    }
 
-  return <Component className={cn(sizeClasses, className)} />
-}
+    return <Component className={cn(sizeClasses, className)} ref={ref} {...props} />
```
praneshg239, line 57: same here. can we remove `{...props}`? Please ensure current functionality is not getting affected.

packages/ui/src/components/tooltip.tsx
```diff
@@ -15,36 +16,51 @@ export type TooltipProps = {
...
+      <TooltipPrimitive.Root delayDuration={delay} open={open}>
+        <TooltipPrimitive.Trigger asChild>{children}</TooltipPrimitive.Trigger>
+        <TooltipPrimitive.Portal container={portalContainer}>
+          <TooltipPrimitive.Content
+            ref={ref}
+            className={cn('cn-tooltip', className)}
+            side={side}
+            align={align}
+            sideOffset={4}
+            {...props}
```
praneshg239, line 48: do we need this in all places?

pixel-buba: Passed `...props` because in some cases this is required for `Tooltip` to work correctly. Without it, the component won’t function as expected.

packages/ui/src/components/progress.tsx
pixel-buba: Below, we use the Text component into `cn-progress-footer`

packages/ui/src/components/tag.tsx
```diff
@@ -56,126 +58,146 @@ type TagProps = Omit<React.HTMLAttributes<HTMLDivElement>, 'role' | 'tabIndex'>
...
-      {/* LEFT TAG - should never have a Reset Icon */}
-      <Tag {...sharedProps} icon={icon} value={label} className="cn-tag-split-left" />
+    return (
+      <div
+        className={cn('cn-tag-split flex w-fit items-center justify-center', {
+          'cursor-not-allowed': disabled,
+          'cn-tag-split-hoverable': !disabled && enableHover
+        })}
+        ref={ref}
+        {...props}
```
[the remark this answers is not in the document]
3em: @praneshg239 We need to keep the `props` spread here, because when wrapping the `Tag` component in a `Tooltip`, the `Tooltip.Trigger` passes some props required for the tooltip to open. Otherwise, it won’t work.

vardanbansal-harness:
Please rebase this PR. As part of another PR, with commit https://github.com/harness/canary/commit/327513c88b8e1dfc51f8eea784d045f76e1054f9, I already ended up adding a `ref` to `SplitButton` component. Feel free to override those changes with the change in this PR as both of them are doing the same thing.

vardanbansal-harness (approved):

PR #1920, opened 2025-07-18
feat: add simple sort component and enhance select with custom trigger renderer

harish-viswa:
image

### Breaking Change
Contract breaks in connector-list-page in platform mfe

### DIFF
--- a/apps/portal/src/content/docs/components/select.mdx
+++ b/apps/portal/src/content/docs/components/select.mdx
@@ -621,7 +621,7 @@ Available options are `horizontal` and `vertical` (default).
     const [value, setValue] = React.useState('banana');
 
     return (
-      <Layout.Vertical gap="lg" className="min-w-[400px]">
+      <Layout.Vertical gap="xl" className="min-w-[400px]">
         <Select
           label="Vertical"
             options={[
--- a/apps/gitness/src/pages-v2/repo/repo-list.tsx
+++ b/apps/gitness/src/pages-v2/repo/repo-list.tsx
@@ -8,7 +8,7 @@ import {
   useDeleteRepositoryMutation,
   useListReposQuery
 } from '@harnessio/code-service-client'
-import { SortValue, Toast, useToast } from '@harnessio/ui/components'
+import { Toast, useToast } from '@harnessio/ui/components'
 import { ExtendedScope, RepoListFilters, RepositoryType, SandboxRepoListPage } from '@harnessio/ui/views'
 
 import { useRoutes } from '../../framework/context/NavigationContext'
@@ -174,7 +174,7 @@ export default function ReposListPage() {
         }
       }}
       onSortChange={(sortValues: string) => {
-        const [type, direction] = sortValues?.split(',')
+        const [type, direction] = sortValues?.split(',') || []
         const sortKey = type as ListReposQueryQueryParams['sort'] | undefined
         const orderKey = direction as ListReposQueryQueryParams['order'] | undefined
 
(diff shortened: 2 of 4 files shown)


shaurya-harness (approved):

packages/ui/src/components/form-primitives/select.tsx
```diff
@@ -87,6 +87,8 @@ interface SelectProps<T = string>
   contentClassName?: string
   suffix?: ReactNode
   triggerClassName?: string
+  triggerRenderer?: (selectedLabel?: ReactNode) => ReactNode
```
abhinavrastogi-harness, line 90: Select component should not have a custom renderer for the trigger IMO. We can use the Dropdown for that.
Instead, maybe we should do a `variant` prop on Select and support a limited set of trigger designs powered via tokens or existing Buttons.
cc @praneshg239 @3em what do you think?

packages/ui/src/components/sorts/simple-sort.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const [sortSelections, updateSortSelections] = useState(defaultSort)
+
+  return (
+    <Select
+      value={sortSelections}
+      triggerRenderer={selectedLabel => (
+        <Button size="sm" variant="transparent">
+          <IconV2 name="arrows-updown" size="2xs" />
+          {selectedLabel || 'Sort'}
+          <IconV2 className="chevron-down text-icons-4" name="nav-solid-arrow-down" size="2xs" />
```
praneshg239, line 22: Do we need `chevron-down` here?

apps/portal/src/content/docs/components/select.mdx
```diff
@@ -675,7 +675,9 @@ The `Select` component is the main component for creating dropdown selections.
   isLoading                 // [OPTIONAL] loading state
   allowSearch               // [OPTIONAL] enable search
   onSearch={customSearch}   // [OPTIONAL] custom search function
-  optionRenderer={renderer} // [OPTIONAL] custom option renderer
+  optionRenderer={renderer} // [OPTIONAL] custom option
+  triggerRenderer={triggerRenderer} // [OPTIONAL] custom trigger renderer
+  dropdownAlign             // [OPTIONAL] dropdown alignment
```
praneshg239, line 680: Any reason we used `dropdownAlign` as boolean here?

praneshg239: `value: "'start' | 'end'",`

packages/ui/src/components/form-primitives/select.tsx
praneshg239: IMO, It is still achievable through `Dropdown` component, also `Dropdown` component natively comes with `checkmark` prop that adds check icon to the option. 

We don't want to alter `Select` component with trigger. It is like reinventing the wheel. 

@abhinavrastogi-harness thoughts?

packages/ui/src/components/sorts/index.ts
```diff
@@ -1,11 +1,13 @@
...
+import SimpleSort from './simple-sort'
 import { Sort as SortRoot } from './sort'
 import SortSelect from './sort-select'
 
 export const Sort = {
   Root: SortRoot,
   Select: SortSelect,
-  MultiSort: MultiSort
+  MultiSort: MultiSort,
+  SimpleSort: SimpleSort
```
vardanbansal-harness, line 10: nit: If the names already convey the type (`SortSelect`, `SimpleSort`, etc.), there's no need to repeat `Sort` in each of them.

```
export const Sort = {
  Root: SortRoot,
  Select: SortSelect,
  Multi: MultiSort,
  Simple: SimpleSort
}

```

PR #1936, opened 2025-07-21
feat: Add Project Pull Requests List page

vardanbansal-harness:
- Adds `Pull Request` menu option that lists pull requests across all the repositories in the current project.

https://github.com/user-attachments/assets/d1364034-0291-4f75-a899-13f21a0f50f2

image

- Hides `Create Pull Request` and `New Pull Request` button for Project level PRs listing page.


packages/ui/src/views/repo/pull-request/pull-request-list-page.tsx
```diff
@@ -171,13 +172,17 @@ const PullRequestListPage: FC<PullRequestPageProps> = ({
           imageName="no-data-folder"
           title="No pull requests yet"
           description={[
-            t('views:noData.noPullRequests', 'There are no pull requests in this project yet.'),
+            t('views:noData.noPullRequests', `There are no pull requests in this ${repoId ? 'repo' : 'project'} yet.`),
```
shaurya-harness, line 175: won't this default to the value in translation store and only if that is not available take the repId ? repo : project string here?

shaurya-harness (approved):

packages/ui/src/views/repo/pull-request/pull-request-list-page.tsx
vardanbansal-harness: Yes I realized that. I cleaned this up and added them as two separate strings instead.

PR #1957, opened 2025-07-22
Add border bottom as default for Tabs underlined variant

pixel-buba:
Screenshot 2025-07-22 at 16 24 57

### DIFF
--- a/packages/ui/src/components/repo-subheader.tsx
+++ b/packages/ui/src/components/repo-subheader.tsx
@@ -38,7 +38,7 @@ export const RepoSubheader = ({
   return (
     <SandboxLayout.SubHeader className={className}>
       <Tabs.NavRoot>
-        <Tabs.List className="border-b border-cn-borders-3 px-6">
+        <Tabs.List className="px-6">
           <Tabs.Trigger value={summaryPath || RepoTabsKeys.SUMMARY}>{t('views:repos.summary', 'Summary')}</Tabs.Trigger>
           <Tabs.Trigger value={filesPath || RepoTabsKeys.CODE}>{t('views:repos.files', 'Files')}</Tabs.Trigger>
           {showPipelinesTab && (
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/repo-subheader.tsx
```diff
@@ -38,7 +38,7 @@ export const RepoSubheader = ({
   return (
     <SandboxLayout.SubHeader className={className}>
       <Tabs.NavRoot>
-        <Tabs.List className="border-b border-cn-borders-3 px-6">
+        <Tabs.List className="px-6">
```
praneshg239, line 41: Do we need `px-6` in all instances? If so can we add it to design system styles (`tabs.ts`) by checking Figma mocks?

pixel-buba: At the moment, there’s only one place where `Tabs.List` has a different offset (ml-4) in the default variant

`packages/ui/src/views/unified-pipeline-studio/components/unified-pipeline-studio-panel.tsx`

```
<Tabs.List className="ml-4 border-none">
  <Tabs.Trigger value="problems">Problems</Tabs.Trigger>
  <Tabs.Trigger value="console">Console</Tabs.Trigger>
</Tabs.List>
```

Since there’s no mock for it, I think it makes sense to allow custom padding for now. What’s your take on that?

praneshg239: @pixel-buba We can go ahead with this now. But it should be solved via design system. 

@cxramina are we not having standard padding for tab list? If not, we have to apply padding every time we use tabs. Else tabs will be very close to its parent. Please check

cxramina: Screen Shot 2025-07-23 at 1 41 13 PM

we have specific token for this underlined tabs version

`tabs.underlined.container-gap`

PR #1962, opened 2025-07-22
first round of mock alignment and bug fixes for v2 repo list screen

jake-bassett:
Several style changes and bug fixes to align with reference designs.

Issues:
- Adjusted styling to align (close to) pixel perfect with mock
- In repository table if repository doesn’t have a description, the title isn’t vertically aligned
- Time text at row should be text-1 color
- The badge next to a repository name should be rounded
- The height of the search field is only 32px when it should be 36px

### DIFF
--- a/packages/ui/src/views/repo/repo-list/repo-list-page.tsx
+++ b/packages/ui/src/views/repo/repo-list/repo-list-page.tsx
@@ -1,6 +1,6 @@
 import { FC, useCallback, useMemo } from 'react'
 
-import { ListActions, NoData, Pagination, SearchInput, Spacer, SplitButton, Text } from '@/components'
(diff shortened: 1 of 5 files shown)


packages/ui/src/views/repo/repo-list/repo-list-page.tsx
```diff
@@ -76,11 +76,11 @@ const SandboxRepoListPage: FC<RepoListProps> = ({
...
   return (
     <SandboxLayout.Main>
-      <SandboxLayout.Content>
+      <SandboxLayout.Content className="pt-0">
         {showTopBar && (
           <>
-            <Spacer size={8} />
-            <div className="flex items-end">
+            <Spacer size={7} className="mt-[30px]" />
+            <div className="flex items-end tracking-[-0.8px]">
```
jake-bassett, line 83: This change to `letter-spacing` might make more sense as a component variant.

PR #1964, opened 2025-07-23
feat: labels page parity gaps 

sans-harness:
1. add checkbox on labels page to show from parent scope
2. Scope field in labels

### DIFF
--- a/apps/gitness/src/pages-v2/project/labels/hooks/use-fill-label-store-with-project-label-values-data.ts
+++ b/apps/gitness/src/pages-v2/project/labels/hooks/use-fill-label-store-with-project-label-values-data.ts
@@ -34,13 +34,14 @@ export const useFillLabelStoreWithProjectLabelValuesData = ({
     setValues,
     setRepoSpaceRef,
     resetLabelsAndValues,
-    setIsLoading
+    setIsLoading,
+    getParentScopeLabels
   } = useLabelsStore()
 
   const { data: { body: labels, headers } = {}, isLoading: isLoadingSpaceLabels } = useListSpaceLabelsQuery(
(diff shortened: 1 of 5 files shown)


apps/gitness/src/pages-v2/project/labels/hooks/use-fill-label-store-with-project-label-values-data.ts
```diff
@@ -34,13 +34,14 @@ export const useFillLabelStoreWithProjectLabelValuesData = ({
...
-    setIsLoading
+    setIsLoading,
+    getParentScopeLabels
   } = useLabelsStore()
 
   const { data: { body: labels, headers } = {}, isLoading: isLoadingSpaceLabels } = useListSpaceLabelsQuery(
     {
       space_ref: `${space_ref}/+`,
-      queryParams: { page: queryPage || 1, limit: 10, query: query ?? '' }
+      queryParams: { page: queryPage || 1, limit: 10, query: query ?? '', inherited: getParentScopeLabels }
```
praneshg239, line 44: @sans-harness are we sending function instance as query param? 🤔

praneshg239: okay, `getParentScopeLabels` name is misleading. Could be like `isParentScopeLabels` if it is a boolean.

sans-harness: Agreed, this was a name that already existed. Since its used in multiple places, changing should be taken up as a separate exercise.

PR #1974, opened 2025-07-23
Error display for maximum token limit and apply styles to link text.

radhakrishna68:
1. Generate Clone Credentials Dialog
    - Display Error validation where the max token limit is reached after clicking the Generate Clone Credentials button.
2. Link Text Styling
     - Updated the link texts styling for consistency (`here.` and `create an API token`).

https://github.com/user-attachments/assets/2f2e58ab-8046-4de0-8e38-b77fc399883d

### DIFF
--- a/packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
+++ b/packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
@@ -7,6 +7,7 @@ import {
   Fieldset,
   FormSeparator,
   Input,
+  Link,
   MarkdownViewer,
   NoData,
   Spacer,
@@ -97,7 +98,8 @@ git push -u origin main
               <span
                 role="button"
                 tabIndex={0}
-                className="hover:decoration-foreground-accent text-cn-foreground-accent underline decoration-transparent underline-offset-4 transition-colors duration-200"
+                className="cn-link cn-link-default cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-cn-foreground-accent"
+                data-disabled="false"
                 onClick={() => navigateToProfileKeys?.()}
                 onKeyDown={e => {
                   if (e.key === 'Enter' || e.key === ' ') {
@@ -106,7 +108,7 @@ git push -u origin main
                   }
                 }}
               >
-                here
+                here.
               </span>
             </Text>
           </ControlGroup>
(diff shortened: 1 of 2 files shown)


packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -84,12 +87,18 @@ git push -u origin main
...
+                <Alert.Description>{tokenGenerationError}</Alert.Description>
+              </Alert.Root>
+            )}
             <Text className="mt-2">
               You can also manage your git credential{' '}
               <span
                 role="button"
                 tabIndex={0}
-                className="hover:decoration-foreground-accent text-cn-foreground-accent underline decoration-transparent underline-offset-4 transition-colors duration-200"
+                className="cn-link cn-link-default cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-cn-foreground-accent"
```
sans-harness, line 100: Why do we need these custom styles?  Can we directly use the `Link` component with the `to` prop?

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -114,7 +123,8 @@ git push -u origin main
               <span
                 role="button"
                 tabIndex={0}
-                className="hover:decoration-foreground-accent text-cn-foreground-accent underline decoration-transparent underline-offset-4 transition-colors duration-200"
+                className="hover:underline cn-link cn-link-default cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-cn-foreground-accent"
```
sans-harness, line 126: Same here, lets avoid `onClick` navigation

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
vardanbansal-harness: @radhakrishna68 `to` prop on the `Link` component you'll add here should have the path returned by `navigateToProfileKeys?.()`.

radhakrishna68: @sans-harness @vardanbansal-harness 

`Link` component's `to` prop needs an path string. `navigateToProfileKeys?.()` is a navigation function. We need to hardcode path/path variable inside.

vardanbansal-harness: I would ideally not want to use `onClick` here but since it's already used, let it be, for now, we should come back and fix. Adding `Link` component for now looks better to achieve style consistency at least.

vardanbansal-harness (approved):

packages/ui/src/views/repo/repo-summary/repo-empty-view.tsx
```diff
@@ -124,7 +134,7 @@ git push -u origin main
                 }}
               >
                 create an API token
-              </span>
+              </span>{' '}
```
vardanbansal-harness, line 137: Check if these extra spaces are added by your `prettier` extension? Usually, I remove these from pull requests to avoid increasing lines of code changed.

radhakrishna68: @vardanbansal-harness  I have added this extra space intentionally to give space between `create an API token` and `In order to pull from or push into this repository.`. 

Screenshot 2025-07-23 at 8 15 28 PM

PR #1991, opened 2025-07-24
feat: add vulnerability scanning checkbox with api

sans-harness:
https://github.com/user-attachments/assets/732fc8f4-2748-469a-9010-95743c0753a7


packages/ui/src/views/repo/repo-settings/components/repo-settings-general-security.tsx
```diff
@@ -97,6 +111,26 @@ export const RepoSettingsSecurityForm: FC<RepoSettingsSecurityFormProps> = ({
             <Message theme={MessageTheme.ERROR}>{errors.secretScanning.message?.toString()}</Message>
           )}
 
+          {isMFE ? (
```
abhinavrastogi-harness, line 114: i'd prefer this to be a more specific condition.. eg the prop could be `showVulnerabilityScanning` instead.
not a blocker.

abhinavrastogi-harness (approved):

PR #1996, opened 2025-07-25
feat: show rules from parent scope checkbox, filters for rules

sans-harness:
https://github.com/user-attachments/assets/129662a7-4656-4227-b1cb-48a2dbf8ab96


apps/gitness/src/pages-v2/project/project-rules-list-container.tsx
```diff
@@ -46,7 +48,10 @@ export const ProjectRulesListContainer = () => {
     space_ref: `${space_ref}/+`,
     queryParams: {
       page: queryPage,
-      query: query ?? ''
+      query: query ?? '',
+      inherited: showParentRules,
+      // @ts-expect-error BE expects an array but the API only works with a string
+      type: ruleTypeFilter ? ruleTypeFilter : undefined
```
vardanbansal-harness, line 54: Is this not resolved with the latest version of code-service client as well? That's the most up-to-date version we have. We should raise this with backend team as a follow up task.

PR #2030, opened 2025-07-31
Update repository creation page

pixel-buba:


packages/ui/src/views/repo/repo-create/index.tsx
```diff
@@ -115,127 +127,141 @@ export function RepoCreatePage({
...
+                  error={errors.license?.message?.toString()}
+                  caption="A license tells others what they can and can't do with your code."
+                  contentWidth="triggerWidth"
+                />
+              </Layout.Vertical>
+
+              {/* ACCESS */}
+              <Fieldset>
+                <Layout.Vertical gap="xl">
+                  <Text variant="body-strong">Who has access</Text>
```
3em, line 204: `FormInput.Radio` has label, you should use it

PR #2050, opened 2025-08-06
fix multiple issues

ankormoreankor:
#### Fixed multiple issues

- made PR link ~~great~~ blue again
- disabled inset theme
- dropped outdated `-topbar-` tokens

image

### DIFF
--- a/packages/ui/src/styles/shared-style-variables.css
+++ b/packages/ui/src/styles/shared-style-variables.css
@@ -564,14 +564,6 @@
     --canary-toast-icon-danger-default: var(--canary-pure-white);
     --canary-toast-icon-danger-hover: var(--canary-pure-white) / 0.65;
 
-    /* Topbar */
-    --canary-topbar-background-01: var(--canary-pure-white);
-    --canary-topbar-foreground-01: var(--canary-base-chrome-1000);
-    --canary-topbar-foreground-02: 240 5% 15%;
-    --canary-topbar-foreground-03: 240 4% 40%;
-    --canary-topbar-foreground-04: 240 5% 25%;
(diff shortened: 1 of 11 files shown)


3em (approved):

packages/ui/tailwind-design-system.ts
```diff
@@ -276,20 +276,6 @@ export default {
             lime: 'var(--canary-label-background-lime-01)'
           }
         },
-        topbar: {
```
praneshg239, line 279: 🔥

praneshg239: thanks for doing this @ankormoreankor ! We need to remove all other older tailwind classes like this

praneshg239 (approved):

packages/ui/src/styles/shared-style-variables.css
```diff
@@ -564,14 +564,6 @@
     --canary-toast-icon-danger-default: var(--canary-pure-white);
     --canary-toast-icon-danger-hover: var(--canary-pure-white) / 0.65;
 
-    /* Topbar */
-    --canary-topbar-background-01: var(--canary-pure-white);
```
praneshg239, line 568: Just a heads up. Its good that we are removing this. But no need, this file will be removed completely soon. So we can avoid spending time in this file.

ankormoreankor: Sure, I understand that 🙌
I only think it would be easier to review it when it has less tokens. It took me only a few minutes to cleanup

3em (approved):

vardanbansal-harness (approved):

PR #2079, opened 2025-08-19
Fix input focus and keyboard navigation in the SearchFiles dropdown

3em:
image

### DIFF
--- a/packages/ui/src/components/search-files.tsx
+++ b/packages/ui/src/components/search-files.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useCallback, useEffect, useState } from 'react'
+import { KeyboardEvent, ReactNode, useCallback, useEffect, useRef, useState } from 'react'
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/search-files.tsx
```diff
@@ -94,14 +96,69 @@ export const SearchFiles = ({
...
+      const items = getItems()
+      items[isFirst ? 0 : items.length - 1]?.focus()
+    },
+    [getItems]
+  )
+
+  const handleInputKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
+      e.preventDefault()
+      queueMicrotask(() => focusItem(e.key === 'ArrowDown'))
```
harish-viswa, line 117: double raf would be better for focus management since the dom would have been settled

harish-viswa (approved):

rkapoor10 (approved):
LGTM

PR #2082, opened 2025-08-19
add props to dropdown subcontent

ankormoreankor:
#### Added props to dropdown subcontent


harish-viswa:
@3em @ankormoreankor The same has to be done for checkboxItem as well

PR #2089, opened 2025-08-20
Fix Dropdown keyboard navigation in ShadowDOM

3em:


packages/ui/src/components/dropdown-menu.tsx
```diff
@@ -68,21 +76,84 @@ interface DropdownMenuContentProps extends ComponentPropsWithoutRef<typeof Dropd
...
+      const rootEl = contentRef.current
+      if (!rootEl) return
+
+      const rootNode = rootEl.getRootNode()
+
+      if (!(rootNode instanceof ShadowRoot)) return
+
+      const activeEl = rootNode instanceof ShadowRoot ? rootNode.activeElement : document.activeElement
+      const items = Array.from(
+        rootEl.querySelectorAll<HTMLElement>('[data-radix-collection-item]:not([data-disabled])')
```
harish-viswa, line 123: In custom dropdowns like branch-selectors where we have tabs, it also has **data-radix-collection-item** which might interfere with this navigation
can we update the selector to have role as menuitem as well?

harish-viswa (approved):

abhinavrastogi-harness (approved):

PR #2098, opened 2025-08-25
Fix navigation in a Tabs within Shadow DOM

3em:
https://github.com/user-attachments/assets/92bdda2b-4b79-4e02-a914-f2005c881fa1

### DIFF
--- a/packages/ui/src/components/tabs.tsx
+++ b/packages/ui/src/components/tabs.tsx
@@ -175,19 +175,8 @@ const TabsList = forwardRef<ElementRef<typeof TabsPrimitive.List>, TabsListProps
       const { isShadowRoot, activeEl } = getShadowActiveElement(rootEl)
       if (!isShadowRoot) return
 
-      const orientation = rootEl.getAttribute('aria-orientation') || 'horizontal'
-      const cssDir = getComputedStyle(rootEl).direction
-
-      const isPrev =
-        (orientation === 'horizontal' &&
-          ((cssDir === 'ltr' && e.key === 'ArrowLeft') || (cssDir === 'rtl' && e.key === 'ArrowRight'))) ||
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/tabs.tsx
```diff
@@ -123,18 +127,103 @@ interface TabsListProps
...
+     */
+    const handleFocusCapture = (e: FocusEvent<HTMLDivElement>) => {
+      props?.onFocusCapture?.(e)
+      if (e.defaultPrevented) return
+
+      const rootEl = contentRef.current
+      if (!rootEl) return
+
+      const { isShadowRoot } = getShadowActiveElement(rootEl)
+      if (!isShadowRoot) return
```
praneshg239, line 155: @3em Was this change tested in standalone application? We need to make sure it is not breaking the standalone app.

packages/ui/src/utils/get-shadow-active-element.ts
```diff
@@ -0,0 +1,8 @@
+export const getShadowActiveElement = (rootEl: HTMLElement) => {
+  const rootNode = rootEl.getRootNode()
+
+  return {
+    isShadowRoot: rootNode instanceof ShadowRoot,
+    activeEl: rootNode instanceof ShadowRoot ? rootNode.activeElement : document.activeElement
```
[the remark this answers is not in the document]
3em: In this case, the error `Property 'activeElement' does not exist on type 'Node'` appears

packages/ui/src/components/tabs.tsx
3em: Yes, I tested it both with and without shadow DOM. This code is specifically meant to fix bugs related to shadow DOM. I already made a lot of similar fixes in other components last week, so everything is fine here

harish-viswa (approved):

PR #2108, opened 2025-08-29
DS migration StackedList

3em:
- Refactored the `StackedList` component
- Added the `StackedList` page to the `Portal`

image


praneshg239:
For header, can we `StackedList.Header` instead using `StackedList.Item` with `isHeader` prop? It is confusing.

image

praneshg239 (approved):

PR #2109, opened 2025-09-01
fix change branch dialog radio behavior

ankormoreankor:


packages/ui/src/views/repo/repo-create/default-branch-dialog.tsx
```diff
@@ -71,7 +71,7 @@ export const DefaultBranchDialog: FC<DefaultBranchDialogProps> = ({ formMethods
 
           <Dialog.Body>
             <FormInput.Radio id="default-branch-radio" {...register('customBranchRadio')}>
-              <Radio.Item id="default-branch-main" value="main" label="main" />
+              <Radio.Item id="default-branch-main" value="main" label="main" autoFocus />
```
3em, line 74: What will happen if the `defaultValue` from `customBranchRadio` is not `main`?

3em (approved):

PR #2118, opened 2025-09-03
fix favorite button focus

ankormoreankor:
#### Fix for Favorite button focus
Issue - https://docs.google.com/spreadsheets/d/1QNr-GLHeiAqdFVo5JiV2vkau-qYekvVDZnnpQ_uN1KM/edit?gid=0#gid=0&range=2:2

- only checkbox is accessible from a keyboard now

image

### DIFF
--- a/packages/ui/src/components/filters/filters-field.tsx
+++ b/packages/ui/src/components/filters/filters-field.tsx
@@ -1,6 +1,5 @@
 import { useMemo, useState } from 'react'
 
-import { Button } from '@components/button'
 import { Checkbox } from '@components/checkbox'
 import { Label } from '@components/form-primitives'
 import { cn } from '@utils/cn'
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/filters/filters-field.tsx
```diff
@@ -103,16 +102,19 @@ const FilterFieldInternal = <T extends string, V extends FilterValueTypes, Custo
       const checkboxId = `checkbox-${uniqId}`
       return (
         // TODO Need to remove button once we get the designs for checkbox filter
-        <Button variant="secondary" size="sm" className="gap-x-2.5 px-2.5 py-1.5">
```
praneshg239, line 106: can we add `tabIndex={-1}` to it?

PR #2142, opened 2025-09-11
Refactor dialog trigger on other repo page

pixel-buba:
Refactoring of the dialog trigger on other repo page:
- files
- pull request page
- tags
- repo settings
- general settings
- users list page

After modals closed, a focus returns to triggers, but not the page body.

### DIFF
--- a/packages/ui/src/components/no-data.tsx
+++ b/packages/ui/src/components/no-data.tsx
@@ -63,8 +63,8 @@ export const NoData: FC<NoDataProps> = ({
 }) => {
   const { NavLink } = useRouterContext()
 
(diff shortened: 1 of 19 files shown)


packages/ui/src/components/no-data.tsx
```diff
@@ -60,6 +63,9 @@ export const NoData: FC<NoDataProps> = ({
 }) => {
   const { NavLink } = useRouterContext()
 
+  const PrimaryWrapper = primaryButton?.isDialogTrigger ? Dialog.Trigger : Fragment
```
praneshg239, line 66: Can we name it more explicitly? Like `PrimaryDialogTrigger` and `SecondaryDialogTrigger`?

PR #2147, opened 2025-09-16
add secrets-related fixes

ankormoreankor:
#### Fixes for Create Secret flow

- `Drawer` gets focused on open now
- fixed `Accordion` Trigger/Content focus UI issues
- made a small change in `InputComponentRenderer.tsx` to fix an issue with `key` prop in mapped items
- fixed UI and keyboard navigation in EntityReference form

### DIFF
--- a/packages/ui/src/views/platform/entity-reference-list.tsx
+++ b/packages/ui/src/views/platform/entity-reference-list.tsx
@@ -1,7 +1,6 @@
 import { Fragment } from 'react'
 
-import { Breadcrumb, IconV2, ScrollArea, StackedList, Text } from '@/components'
-import { cn } from '@utils/cn'
+import { Breadcrumb, ScrollArea, StackedList } from '@/components'
 
 import {
   BaseEntityProps,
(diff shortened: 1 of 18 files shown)


packages/ui/src/views/platform/entity-reference-list.tsx
```diff
@@ -44,7 +42,6 @@ export function EntityReferenceList<T extends BaseEntityProps, S = string, F = s
   defaultEntityRenderer,
   parentFolderRenderer,
   childFolderRenderer,
-  apiError,
```
ankormoreankor, line 47: This prop was never actually passed, and was handled the wrong way.
API error for this component is handled properly in PlatformUI package.

packages/ui/tailwind-utils-config/components/accordion.ts
```diff
@@ -84,8 +100,22 @@ export default {
       '@apply font-body-normal': '',
 
       '&-container': {
+        marginTop: '-4px',
```
praneshg239, line 103: why is this needed?

packages/ui/tailwind-utils-config/components/multi-select.ts
```diff
@@ -56,6 +56,7 @@ export default {
     },
 
     '&-input': {
+      minWidth: '50px',
```
praneshg239, line 59: why we are adding `minWidth` here?

ankormoreankor: Without minWidth the input can be too small to type

image

ankormoreankor: With this change it goes to the next line when there's not enough space

packages/ui/tailwind-utils-config/components/accordion.ts
praneshg239: Here the focus state is still hidden at the corners. Could you please check it?

packages/ui/src/views/platform/entity-reference-list.tsx
```diff
@@ -139,8 +120,8 @@ export function EntityReferenceList<T extends BaseEntityProps, S = string, F = s
             })}
           </>
         ) : (
-          <StackedList.Item>
-            <StackedList.Field title={apiError || 'No entities found'} />
+          <StackedList.Item disableHover>
+            <StackedList.Field title="No entities found" titleColor="foreground-3" />
```
shaurya-harness, line 124: why is apiError section removed? afair @sans-harness added this to be able to support api failure messages

srdjan-harness (approved):

packages/ui/src/views/platform/types.ts
```diff
@@ -19,11 +19,11 @@ export interface EntityRendererProps<T extends BaseEntityProps> {
...
 
 export interface ParentFolderRendererProps<S = string> {
-  parentFolder: S
+  parentFolder?: S | null
   onSelect: (scope: S) => void
 }
 
 export interface ChildFolderRendererProps<F = string> {
-  folder: F
+  folder?: F | null
```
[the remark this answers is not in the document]
ankormoreankor: @sans-harness I basically incapsulated the render logic inside the render function, so it can be called unconditionally. This change doesn't affect anything else

image

PR #2149, opened 2025-09-17
fix label with no value

ankormoreankor:
#### Fix for label with no value

- additionally added more space to the search placeholder

image

image

### DIFF
--- a/packages/ui/src/views/labels/components/label-tag.tsx
+++ b/packages/ui/src/views/labels/components/label-tag.tsx
@@ -16,7 +16,7 @@ export const LabelTag: FC<LabelTagProps> = ({ scope, withIndicator = false, wrap
       size="sm"
       icon={scope !== undefined ? scopeTypeToIconMap[getScopeType(scope)] : undefined}
       variant="secondary"
-      className="grid grid-flow-col"
+      className="grid grid-flow-col content-center"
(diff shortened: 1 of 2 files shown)


3em (approved):

packages/ui/src/views/labels/components/label-tag.tsx
```diff
@@ -16,7 +16,7 @@ export const LabelTag: FC<LabelTagProps> = ({ scope, withIndicator = false, wrap
       size="sm"
       icon={scope !== undefined ? scopeTypeToIconMap[getScopeType(scope)] : undefined}
       variant="secondary"
-      className="grid grid-flow-col"
+      className="grid grid-flow-col content-center"
       labelClassName="grid grid-flow-col content-center"
       valueClassName="grid grid-flow-col content-center"
```
praneshg239, line 21: Do we really these classes? Without these classes we are achieving the layout already. Can you please tell the exact purpose of these classes?

Image

ankormoreankor: Yes, we absolutely need them. We discussed this when we had a call last week. After the call I tried the solution with max-width: 50%, but it's not working properly. We should keep it the way it presented.
cc @praneshg239

praneshg239 (approved):

PR #2150, opened 2025-09-17
Refactored the secrets form and removed empty divs in Forms

3em:
- In the `harness/forms` package, the `Row` component was used, which always rendered an empty `div` even if `InputComponentRenderer` was `null`
- Fixed the form paddings and removed the extra wrapper


packages/ui/src/views/secrets/secret-entity-form.tsx
```diff
@@ -127,7 +127,7 @@ export const SecretEntityForm = forwardRef<SecretEntityFormHandle, SecretEntityF
           return (
             <>
               <RenderForm
-                className="max-w-xl space-y-4"
+                className="max-w-xl space-y-6"
```
praneshg239, line 130: NIT: Can we update it to design system class?

praneshg239 (approved):

PR #2162, opened 2025-09-25
Fix Card-select prop type and Bump UI

3em:


packages/ui/src/components/card-select.tsx
```diff
@@ -1,27 +1,25 @@
 import { createContext, forwardRef, HTMLAttributes, ReactNode, useContext, useState } from 'react'
 
 import { cn } from '@utils/cn'
-import { cva } from 'class-variance-authority'
+import { cva, VariantProps } from 'class-variance-authority'
```
praneshg239, line 4: can you please import it as `type`?

praneshg239 (approved):

PR #2169, opened 2025-09-30
Input-reference-component refactoring

3em:
image

image

image


packages/ui/src/views/platform/input-reference-component.tsx
```diff
@@ -144,6 +126,10 @@ export const InputReference = <T,>({
   suffix,
   onOpen,
   wrapperClassName = '',
+  orientation = 'vertical',
+  labelSuffix,
+  tooltipProps,
+  tooltipContent,
```
sanskar-harness, line 132: Do we need any of these props to maintain functionality? If so we should also open a corresponding `pl-ui` PR.

PR #2184, opened 2025-10-08
Add setPageSize functionality across various stores and components

pixel-buba:
Screenshot 2025-10-08 at 16 16 59

Screen recording
branches - indeterminate variant

https://github.com/user-attachments/assets/75196abc-e5d6-4296-b85f-2b470df50c5e

labels - with currentPage variant 

https://github.com/user-attachments/assets/9f6a0baa-4f8f-4b70-9fcd-f7ae1854dff6

### DIFF
--- a/apps/gitness/src/pages-v2/pull-request/stores/pull-request-list-store.tsx
+++ b/apps/gitness/src/pages-v2/pull-request/stores/pull-request-list-store.tsx
@@ -79,11 +79,11 @@ export const usePullRequestListStore = create<PullRequestListStore>(set => ({
         })) || []
     }))
 
-    set({
+    set(state => ({
       pullRequests: transformedPullRequests,
       totalItems: parseInt(headers?.get(PageResponseHeader.xTotal) || '0'),
-      pageSize: parseInt(headers?.get(PageResponseHeader.xPerPage) || '10')
-    })
+      pageSize: parseInt(headers?.get(PageResponseHeader.xPerPage) || String(state.pageSize))
+    }))
   },
   setOpenClosePullRequests: (openPullReqs, closedPullReqs, mergedPullReqs) => {
     set({
--- a/apps/gitness/src/pages-v2/webhooks/stores/webhook-store.tsx
+++ b/apps/gitness/src/pages-v2/webhooks/stores/webhook-store.tsx
@@ -52,12 +52,12 @@ export const useWebhookStore = create<WebhookStore>(set => ({
     }))
   },
   setPaginationFromHeaders: (headers: Headers | undefined, isExecution: boolean = false) => {
-    set({
+    set(state => ({
       totalItems: parseInt(headers?.get(PageResponseHeader.xTotal) || '0'),
       [isExecution ? 'webhookExecutionPageSize' : 'pageSize']: parseInt(
-        headers?.get(PageResponseHeader.xPerPage) || '10'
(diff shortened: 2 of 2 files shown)


apps/gitness/src/pages-v2/pull-request/stores/pull-request-list-store.tsx
```diff
@@ -80,7 +82,7 @@ export const usePullRequestListStore = create<PullRequestListStore>(set => ({
     set({
       pullRequests: transformedPullRequests,
       totalItems: parseInt(headers?.get(PageResponseHeader.xTotal) || '0'),
-      pageSize: parseInt(headers?.get(PageResponseHeader.xPerPage) || '30')
+      pageSize: parseInt(headers?.get(PageResponseHeader.xPerPage) || '10')
```
rkapoor10: not sure but should we pass the `pageSize` variable as fallback instead of 10 here as well like we did 
`  pageSize: parseInt(headers?.get(PageResponseHeader.xPerPage) ?? String(pageSize))`

apps/gitness/src/pages-v2/repo/repo-commits.tsx
```diff
@@ -21,6 +21,8 @@ export default function RepoCommitsPage() {
   const { gitRefName, fullGitRef } = useCodePathDetails()
   const { toRepoCommits } = useRepoCommits()
 
+  const [pageSize, setPageSize] = useState(25)
```
rkapoor10: nit - why 25 ?

packages/ui/src/components/pagination/pagination.tsx
```diff
@@ -24,14 +25,16 @@ export function Pagination({
...
+
   const handleGoToPage = (selectedPage?: number) => (e: React.MouseEvent) => {
     e.preventDefault()
     if (!selectedPage) return
 
     goToPage?.(selectedPage)
   }
 
-  const totalPages = indeterminate || !totalItems || !pageSize ? undefined : Math.ceil(totalItems / pageSize)
+  const totalPages = indeterminate || !totalItems ? undefined : Math.ceil(totalItems / pageSize)
```
rkapoor10: question - so we usually get total pages for a paginated api along with total no of items...which we can use first before falling back to calculating the total based on items ?

Image

packages/ui/src/views/labels/labels-list-page.tsx
```diff
@@ -104,7 +105,13 @@ export const LabelsListPage: FC<LabelsListPageProps> = ({
               values={spaceValues}
               toRepoLabelDetails={toRepoLabelDetails}
             />
-            <Pagination totalItems={totalItems} pageSize={pageSize} currentPage={page} goToPage={setPage} />
+            <Pagination
+              totalItems={totalItems}
+              pageSize={pageSize}
+              onPageSizeChange={setPageSize}
+              currentPage={page}
+              goToPage={setPage}
```
rkapoor10: we can add a total pages prop as well ?

packages/ui/src/components/pagination/pagination.tsx
praneshg239: What @rkapoor10 is saying does makes sense to me. We can add a prop for `totalPages` and that can take precedence. 

@rkapoor10 can we do it as separate PR, right next to this PR? So that we can avoid big PRs.

cc: @3em

PR #2263, opened 2026-06-18
Update Dark Mode Colors 1000-700

Kgrantier:
Updated dark mode color ramp scale from grey.1000-700 both LCH and Hex values. This update is to match the Harness 3.0 vision that can be found here: https://clever-kringle-70f506.netlify.app/pipeline?flow=toggled&start=pipelines

Jira Ticket: https://harness.atlassian.net/browse/XD-184


Kgrantier:
Need to adjust the LCH values, they are not visually displaying the color that we want.

